PR #26, opened 2017-07-05
added operating system version check for build.

lianglia-apollo:


joaospinto:
Alternatively, we can force the build to occur inside the corresponding docker environment (as encouraged in the main README.md file).
Finding whether we are inside docker can be done with:
`if [ -f /.dockerenv ]; then ...`
We would also want to check if we are in the right environment, which can be done by setting some environment variable to mark this.

PR #27, opened 2017-07-05
Add dependency for ros in run env.

wanglei828:


wanglei828:
@startcode @ycool  Caros team removed some so files from ros output. I need to add these missing libraries to our run env image.

PR #100, opened 2017-07-09
Update install_docker.sh

gengqx:
It fix a bug when install docker use the install_docker.sh.The original line did not use --force-yes,it would get an error,it would not pass the security verification.May be 'sudo apt-get update' is run properly


lianglia-apollo:
I don't think blindly answer yes to all questions is good here. Could you post what error message you encountered?

PR #128, opened 2017-07-11
Fix env parameter in scripts, making it work in mac.

wanglei828:
The env parameters in scripts need to be the parameters in host machine not  container.


docker/scripts/dev_start.sh
```diff
@@ -96,7 +96,7 @@ function main(){
         -v /tmp/.X11-unix:/tmp/.X11-unix:rw \
         -v $LOCAL_DIR:/apollo \
         -v /media:/media \
-        -v $HOME:$HOME \
+        -v $HOME:/home/$USER \
```
startcode: what if $USER = root?

Issue #166, opened 2017-07-12
Error at apollo.sh build


Eulerianial:
I ran following commands
bash docker/scripts/install_docker.sh
docker ps
bash docker/scripts/dev_start.sh
bash docker/scripts/dev_into.sh
bash apollo.sh build

and at the build phase I am getting following errors:
WARNING: Running Bazel server needs to be killed, because the startup options are different.
INFO: Found 189 targets...
ERROR: /apollo/modules/hmi/ros_node/BUILD:3:1: Linking of rule '//modules/hmi/ros_node:ros_node_service' failed: gcc failed: error executing command /usr/bin/gcc -o bazel-out/local-dbg/bin/modules/hmi/ros_node/ros_node_service '-Wl,-rpath,$ORIGIN/../../../_solib_k8/_U@ros_S_S_Cros_Ucommon___Uexternal_Sros_Slib' ... (remaining 8 argument(s) skipped): com.google.devtools.build.lib.shell.BadExitStatusException: Process exited with status 1.
external/boringssl/boringssl-master-with-bazel/src/ssl/handshake_server.cc:778: error: undefined reference to 'tls13_server_handshake'
collect2: error: ld returned 1 exit status
INFO: Elapsed time: 1308.514s, Critical Path: 480.22s
============================
[ERROR] Build failed!
[INFO] Took 1722 seconds

siyangy:
This error has never been reported. It could just be a glitch of the bazel build system. Can you try only building the failed target by `bazel --batch --batch_cpu_scheduling build --define ARCH=$(uname -m) --define CAN_CARD=fake_can --cxxopt=-DUSE_ESD_CAN=false -c dbg //modules/hmi/ros_node:ros_node_service` to see if it happens again?

Also your build time seems very long, which indicates that your memory is not enough. Another guess is that g++ sometimes has weird behavior when it exhausts memory.

zwqaoy:
only building the failed target by bazel --batch --batch_cpu_scheduling build --define ARCH=$(uname -m) --define CAN_CARD=fake_can --cxxopt=-DUSE_ESD_CAN=false -c dbg //modules/hmi/ros_node:ros_node_service

the error happened again.  
my gcc version is gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4

quning78:
So you are not running  inside docker? If so, the ubuntu is supposed to be
14.04.

Best wishes,
-- 
Ning Qu

[...]

PR #185, opened 2017-07-13
use a working revision of boringssl

CecilHarvey:
The latest master revision does not work.
This fixed Issue #166


siyangy:
Thanks for fixing this problem!

PR #206, opened 2017-07-14
Limit aggressive steering based on vehicle speed

henryhu6:


henryhu6:
Line 235: minimum vehicle_state_.linear_velocity() value is 1.0

Issue #82, opened 2017-07-07
CI setup for apollo


ycool:
https://travis-ci.org/

Thanks.

xiaoxq:
Has been setup by wanglei828.

PR #208, opened 2017-07-15
Enable travis ci build.

wanglei828:


scripts/travis.sh
```diff
@@ -0,0 +1,207 @@
+#!/usr/bin/env bash
```
siyangy: why do we need travis.sh instead of just using apollo.sh? Looks like they are largely identical.

wanglei828: apollo.sh always source apollo_base.sh which needs user to input "y or Y" to accept the license.

siyangy: Possibly add an option to apollo.sh to skip sourcing it, otherwise keeping this and apollo.sh in sync might be a pain.

wanglei828: Another option is remove "accept license" function from apollo_base.sh. I don't think it is good to put it there.

wanglei828: @siyangy  I fixed your concern, and remove the travis.sh file. Now only the travis ci configure file is needed.

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+before_install:
+    - if [ ! -d "/root/.cache" ]; then sudo mkdir /root/.cache; fi
+    - sudo cp scripts/AGREEMENT.txt /root/.cache/.apollo_agreement.txt
+    - docker pull ${IMAG}
+    - docker run -d -it --name apollo_dev -v ${TRAVIS_BUILD_DIR}:/apollo -v /root/.cache:/root/.cache -w /apollo ${IMAG}
+    - docker ps 
+install: skip
+script:
+    - docker exec apollo_dev  bash -c "./apollo.sh clean"
+    - docker exec apollo_dev  bash -c "./apollo.sh build"
```
siyangy: Should we run test and lint here by ./apollo.sh check other than build?

wanglei828: I think this is the first step to enable this ci build feature. We can add more jobs later on by editing the configure file.

wanglei828: @siyangy Due to our repo does not have esd_can lib, many tests will fail (49 of 74 fails). I only add "build" and "lint" to the CI.

siyangy: We should definitely fix this. Unit test should be included in CI. It doesn't have to be fixed in this PR, though.

Btw, next time please add a new commit instead of amending the existing one so that the reviewer would know what change you made and the comments won't be lost. GitHub will squash all the commits into one upon merge.

wanglei828:
Resolve issue #82

PR #229, opened 2017-07-18
Civet webserver to handle websocket requests. Remove dependency on websocketpp

siyangy:
In addition, dreamview will no longer occupy port 8889.


siyangy:
@unacao

PR #321, opened 2017-09-19
planning: disable ReferenceLineInfo copy constructor

startcode:


modules/planning/common/reference_line_info.h
```diff
@@ -48,6 +48,9 @@ class ReferenceLineInfo {
       const common::TrajectoryPoint& init_adc_point,
       const ReferenceLineSmootherConfig& smoother_config);
 
+  ReferenceLineInfo(const ReferenceLineInfo&) = delete;
```
siyangy: Use DISALLOW_COPY_AND_ASSIGN macro. Also you need to do the same thing for PathDecision which owns an IndexedList.

startcode: I will refactor IndexedList. It should  be a general container, not a container only for unique_ptr.

PR #331, opened 2017-09-20
planning: refactor indexed_list

startcode:


modules/planning/common/frame.cc
```diff
@@ -84,7 +84,7 @@ void Frame::CreatePredictionObstacles(
   auto obstacles = Obstacle::CreateObstacles(prediction);
   for (auto &ptr : obstacles) {
     auto id(ptr->Id());
-    obstacles_.Add(id, std::move(ptr));
+    obstacles_.Add(id, *ptr);
```
siyangy: Is it intended to copy the obstacle here? Before this change this was a move without copy.

startcode: yes. Copy.

PR #339, opened 2017-09-20
commit mapauto

pany13579:
I had deleted BaiduMapAutoEngineAPIDoc.


startcode:
These are html documents. I think there is no point to checkin such documents in github. It is better to host them on the apollo.auto website.

PR #384, opened 2017-09-22
Planning: Fix init point bug

siyangy:


startcode:
I guess it will not affect integration test. There is no history trajectory in integration tests.

PR #429, opened 2017-09-27
HMI: use gevent as async handler

siyangy:


siyangy:
@xiaoxq Should docker image be updated as well? I had to do pip install gevent manually to get this to work.

siyangy:
@wanglei828 Please merge this after the image update is done.

xiaoxq:
Yes, if we set async_mode explicitly, it will crash if gevent is not installed.
So let's hold on for a moment before docker updated, then it works fine for either eventlet or gevent during migration.
After that we force to use gevent explicitly.

If you have to work on 16.04, you need to pip uninstall eventlet and install gevent.

siyangy:
I don't think it's necessary to pip uninstall eventlet, as you have explicitly set the mode as 'gevent'.

xiaoxq:
I mean, before merging this PL.
Eventlet has higher priority over gevent, so it will still fail on 16.04.

Anyway, our plan is just to update docker and set async_mode=gevent.

siyangy:
No without setting it to gevent it will always try to find eventlet. I've got HMI working in 16.04 without uninstalling eventlet.

xiaoxq:
That is because you have set async_mode=gevent, so you don't need to
uninstall eventlet.
The migration process is:

*Current:*
async_mode = None
in 14.04 docker it works fine with eventlet
in 16.04 you need to make sure only gevent is installed

*After docker updated:*
async_mode = None
in 14.04 docker it works fine with gevent (as we removed eventlet)
in 16.04 you need to make sure only gevent is installed

*Finally:*
async_mode = gevent
Good for 14.04/16.04, and don't care if eventlet is installed


My point is, during docker updating, the official 14.04 environment should
never break.

PR #549, opened 2017-10-04
changed delphi_esr/mobileye/l3_perception launch scripts

rongqiqiu:


scripts/delphi_esr_launch.sh
```diff
@@ -1 +1,28 @@
-./bazel-bin/modules/drivers/delphi_esr/delphi_esr --flagfile=modules/drivers/delphi_esr/conf/delphi_esr.conf
+#!/usr/bin/env bash
```
siyangy: It seems that the current launch scripts don't have the '_launch' suffix, so better to follow the convention.

PR #602, opened 2017-10-09
Tools: Add script to stat auto/manual mileage from bags.

xiaoxq:


startcode:
Seems this tool exclusively deal with rosbag, can we move this tool to modules/tools/rosbag?

xiaoxq:
Good point. Thanks!

PR #650, opened 2017-10-12
add create_map tool

rongqiqiu:
Add a tool to create map from localization and mobileye


rongqiqiu:
Temporarily closing as this tool depends on Mobileye. Will reopen after Mobileye is merged.

PR #705, opened 2017-10-17
Dreamview: Add reset backend data

siyangy:


modules/dreamview/frontend/src/store/websocket.js
```diff
@@ -30,6 +29,7 @@ class WebSocketEndpoint {
             switch (message.type) {
                 case "SimWorldUpdate":
                     this.checkMessage(message);
+                    console.log(message);
```
unacao: may need to suppress this log

siyangy: yeah just noticed and removed

PR #711, opened 2017-10-18
common: added thread_num suppport in apollo app

startcode:


modules/calibration/lidar_ex_checker/main.cc
```diff
@@ -20,4 +20,5 @@
 
 #include "modules/calibration/lidar_ex_checker/lidar_ex_checker.h"
 
-APOLLO_MAIN(apollo::calibration::LidarExChecker);
+int32_t thread_num = 1;
```
siyangy: default thread_num can be defined in apollo_app.h so that it doesn't need to be set in each module.

startcode: I think explicitly set thread_num is more clear. Easier for people to find the clue when they seeking mult-thread callback solutions.

startcode: I will change back to let each class set this number, to adapter for dynamic thread-num case.

PR #719, opened 2017-10-18
canbus: add socket_can_client_raw support

Capri2014:


scripts/canbus.sh
```diff
@@ -23,6 +23,11 @@ cd "${DIR}/.."
 
 source "$DIR/apollo_base.sh"
 
+MACHINE_ARCH=$(uname -m)
+if [ "$MACHINE_ARCH" == 'aarch64' ]; then
+  sudo ip link set can0 type can bitrate 500000
+  sudo ip link set can0 up
+fi
 # run function from apollo_base.sh
```
jinghaomiao: Put this in scripts/apollo_base.sh: setup_device function so that we can manage all devices at the same place.

scripts/canbus.sh
```diff
@@ -23,6 +23,11 @@ cd "${DIR}/.."
 
 source "$DIR/apollo_base.sh"
 
+MACHINE_ARCH=$(uname -m)
```
jinghaomiao: Put this in scripts/apollo_base.sh: setup_device function so that we can manage all devices at the same place.

PR #743, opened 2017-10-19
common: make gauss integration a template function

YajiaZhang:


startcode:
use template specialization, here is an example.
```
#include <iostream>
using namespace std;

template <int N>
void special_func() {
  std::cout << N << std::endl;
}

template <>
void special_func<1>() {
  std::cout << "one" << std::endl;
}

template <>
void special_func<2>() {
  std::cout << "two" << std::endl;
}

int main() {
  special_func<1>();
  special_func<2>();
  special_func<3>();
  return 0;
}
```

PR #749, opened 2017-10-20
 tools:make stat_mileage.py more accurate to count disengagements

vfengzhi:


modules/tools/rosbag/stat_mileage.py
```diff
@@ -55,7 +55,7 @@ def calculate(self, bag_file):
                 if topic == kChassisTopic:
                     # Mode changed
                     if cur_mode != msg.driving_mode:
-                        if msg.driving_mode == Chassis.EMERGENCY_MODE:
+                        if msg.driving_mode == Chassis.EMERGENCY_MODE and cur_mode == Chassis.COMPLETE_AUTO_DRIVE:
```
xiaoxq: Please limit line width to 80 chars.
You could also move it out from the upper "if" statement.

modules/tools/rosbag/stat_mileage.py
```diff
@@ -55,7 +55,8 @@ def calculate(self, bag_file):
                 if topic == kChassisTopic:
                     # Mode changed
                     if cur_mode != msg.driving_mode:
-                        if msg.driving_mode == Chassis.EMERGENCY_MODE:
+                        if (cur_mode == Chassis.COMPLETE_AUTO_DRIVE and
+                                msg.driving_mode == Chassis.EMERGENCY_MODE):
```
xiaoxq: How about wrap the line with brackets?

PR #762, opened 2017-10-20
Dreamview: Update HMI status by reading /apollo/hmi_status message.

xiaoxq:


modules/dreamview/backend/hmi/hmi_test.cc
```diff
@@ -0,0 +1,46 @@
...
+
+TEST(HMITest, UpdateHMIStatus) {
+  HMI hmi(nullptr);
+
+  EXPECT_GT(hmi.config_.modules_size(), 0);
+  EXPECT_EQ(hmi.config_.modules_size(), hmi.status_.modules_size());
+
+  // The initial status should be UNINITIALIZED.
+  const auto& module_name = hmi.config_.modules(0).name();
+  EXPECT_EQ(hmi.GetModuleStatus(module_name)->status(),
```
siyangy: NIT: EXPECT_EQ(expected, actual). same for the rest

modules/dreamview/backend/hmi/hmi.cc
```diff
@@ -0,0 +1,89 @@
...
+  AdapterManager::AddHMIStatusCallback(&HMI::OnHMIStatus, this);
+}
+
+void HMI::OnHMIStatus(const HMIStatus &hmi_status) {
+  // Note that we only merged status of modules and hardware, which matches the
+  // hmi_status_helper.
+  for (const auto& new_status : hmi_status.modules()) {
+    auto* old_status = GetModuleStatus(new_status.name());
+    if (old_status) {
+      old_status->MergeFrom(new_status);
```
siyangy: Can you just set the new status in the function instead of returning a pointer and call MergeFrom? not sure why there needs to be a MergeFrom rather than just setting the new status.

xiaoxq: MergeFrom is to support partial update.
Per offline discussion, we'll keep it for now and refactor later if necessary.

modules/dreamview/backend/hmi/hmi_test.cc
xiaoxq: done

PR #789, opened 2017-10-24
Prediction: move callback to public

siyangy:


modules/planning/planning.cc
```diff
@@ -163,10 +163,6 @@ Status Planning::Start() {
 
 void Planning::OnTimer(const ros::TimerEvent&) {
   RunOnce();
-  if (frame_) {
```
ycool: FrameHistory is used at em_planner.cc line 172.

startcode: I guess we should only depends on the previous trajectory, not previous frame. Actually many information in previous frame is invalid. I will remove this dependency tomorrow.

PR #798, opened 2017-10-24
Dreamview: Add handlers of RetrieveHMIConfig, ChangeMap and ChangeVehicle.

xiaoxq:
…icle.


modules/dreamview/proto/hmi_config.proto
```diff
@@ -17,8 +17,15 @@ message Component {
...
+message KV {
+  optional string name = 1;
+  optional string value = 2;
+}
+
 message HMIConfig {
   repeated Component modules = 1;
   repeated Component hardware = 2;
   repeated Command tools = 3;
+  repeated KV available_maps = 4;  // Dict of {map_name: map_dir}
```
siyangy: possible to use https://developers.google.com/protocol-buffers/docs/proto#maps?

PR #858, opened 2017-10-26
Monitor: added can checker factory

jinghaomiao:


modules/monitor/hwmonitor/hw_check/can_check.cc
```diff
@@ -57,9 +56,10 @@ int main(int argc, const char *argv[]) {
   apollo::monitor::hw::esdcan_print_summary(
       std::cout, *(const EsdCanDetails *)((can_rslt[0].details.get())));
 #else
-  PLATFORM_LOG(
-      apollo::monitor::hw::get_log_module(), apollo::monitor::log::LVL_DBG,
-      "Done checking ESD-CAN-%d, status: %d", can_id, can_rslt[0].status);
+  PLATFORM_LOG(apollo::monitor::hw::get_log_module(),
+               apollo::monitor::log::LVL_DBG,
+               "Done checking ESD-CAN-%d, status: %d", can_chk.get_id(),
```
Capri2014: We should remove "ESD" in line 61

jinghaomiao: Yes, we'll do that later on. This is the first checkin of the factory.

PR #864, opened 2017-10-26
Calibration: Retire unused data.

xiaoxq:


startcode:
I think we may keep these files as samples for developers, so that they can have a feel about what these files should look like.

PR #865, opened 2017-10-26
canbus: fix MonitorBuffer crash problem when AdapterManager is not in…

startcode:
…itialized

### DIFF
--- a/modules/common/adapters/adapter_manager.h
+++ b/modules/common/adapters/adapter_manager.h
@@ -115,16 +115,22 @@ namespace adapter {
                                                                                \
     observers_.push_back([this]() { name##_->Observe(); });                    \
   }                                                                            \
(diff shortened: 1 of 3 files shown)


modules/common/adapters/adapter_manager.h
```diff
@@ -115,16 +115,22 @@ namespace adapter {
...
+      } else {                                                                 \
+        AERROR << #name << " is not valid.";                                   \
+      }                                                                        \
     } else {                                                                   \
       /* For non-ROS mode, just triggers the callback. */                      \
-      name##_->OnReceive(data);                                                \
+      if (name##_) {                                                           \
+        name##_->OnReceive(data);                                              \
+      } else {                                                                 \
+        AERROR << #name << " is null.";                                        \
```
siyangy: This is unreachable and can be removed.

PR #908, opened 2017-10-27
Data: Add info collector.

xiaoxq:


xiaoxq:
Sort of human readable name, such as "MKZ8".
Could be removed if it's not that necessary.

PR #933, opened 2017-10-30
Prediction: added trajectory validation

jinghaomiao:


modules/prediction/prediction.cc
```diff
@@ -124,13 +122,29 @@ void Prediction::OnPerception(const PerceptionObstacles& perception_obstacles) {
...
+  }
 }
 
 Status Prediction::OnError(const std::string& error_msg) {
   return Status(ErrorCode::PREDICTION_ERROR, error_msg);
 }
 
+bool Prediction::IsValidTrajectoryPoint(
+    const TrajectoryPoint& trajectory_point) {
+  return (!std::isnan(trajectory_point.path_point().x())) &&
```
startcode: check trajectory_point.has_path_point()

jinghaomiao: This piece of code is to capture anything abnormal in real road test as I am not able to reproduce the nan issue reported in 101 test with the bags collected. Therefore if trajectory_point has no path_point, it will be captured and a core dump will be thrown. This piece of code will be removed with enough tests/verifications.

PR #964, opened 2017-10-31
Planning: use diff kernels at different lateral offset positions in q…

lianglia-apollo:
…p_spline_path.


modules/planning/tasks/qp_spline_path/qp_spline_path_generator.cc
```diff
@@ -403,23 +403,31 @@ void QpSplinePathGenerator::AddKernel() {
...
+                 qp_spline_path_config_.derivative_weight());
+    }
   }
 
   if (qp_spline_path_config_.second_derivative_weight() > 0.0) {
     spline_kernel->AddSecondOrderDerivativeMatrix(
         qp_spline_path_config_.second_derivative_weight());
-    spline_kernel->AddSecondOrderDerivativeMatrixForSplineK(
-        0, qp_spline_path_config_.second_derivative_weight());
+    if (init_frenet_point_.l() > qp_spline_path_config_.lane_change_mid_l()) {
```
ycool: need consider l negative case

PR #992, opened 2017-11-01
planning: further performance improvement for spiral reference line s…

YajiaZhang:
…moother


modules/planning/reference_line/spiral_reference_line_smoother.cc
```diff
@@ -99,6 +99,8 @@ bool SpiralReferenceLineSmoother::Smooth(
   ADEBUG << "Spiral reference line smoother time: "
          << (end_timestamp - start_timestamp) * 1000 << " ms.";
 
+  std::cout << "Spiral reference line smoother time:\t"
```
ycool: remove std::cout log?

YajiaZhang: Right, forgot to remove it.

PR #1001, opened 2017-11-02
Control: Add additional flags for test

Capri2014:


modules/control/controller/lat_controller.cc
```diff
@@ -543,37 +543,44 @@ void LatController::ComputeLateralErrors(
...
-  double cos_matched_theta = std::cos(matched_point.path_point().theta());
-  double sin_matched_theta = std::sin(matched_point.path_point().theta());
+  double cos_matched_theta = std::cos(target_point.path_point().theta());
+  double sin_matched_theta = std::sin(target_point.path_point().theta());
   // d_error = cos_matched_theta * dy - sin_matched_theta * dx;
   debug->set_lateral_error(cos_matched_theta * dy - sin_matched_theta * dx);
 
   double delta_theta =
-      common::math::NormalizeAngle(theta - matched_point.path_point().theta());
+      common::math::NormalizeAngle(theta - target_point.path_point().theta());
```
startcode: You may want to use the AngelDiff function here
```
 double AngleDiff(const double from, const double to) 
```

PR #1018, opened 2017-11-02
Dreamview: Listen to system_status and update clients.

xiaoxq:


modules/dreamview/backend/hmi/hmi.cc
```diff
@@ -27,6 +27,7 @@
 #include "modules/common/util/util.h"
 #include "modules/control/proto/pad_msg.pb.h"
 #include "modules/dreamview/backend/common/dreamview_gflags.h"
+#include "modules/monitor/proto/system_status.pb.h"
```
startcode: do we need to group *.pb.h includes together?

PR #1031, opened 2017-11-02
common: added a Dropbox utility, which is a key-value container.

startcode:


modules/common/util/dropbox.h
```diff
@@ -0,0 +1,80 @@
...
+namespace common {
+namespace util {
+
+using KeyType = std::string;
+
+/**
+ * @brief Dropbox class is a map based key-value storage container utility.
+ */
+template <class T>
+class Dropbox {
```
siyangy: how is this different with std::unordered_map?

PR #1043, opened 2017-11-03
dreamview: add initial velocity and acceleration for sim control

timboilermaker:


modules/dreamview/backend/common/dreamview_gflags.h
```diff
@@ -31,6 +31,10 @@ DECLARE_string(server_ports);
 
 DECLARE_bool(enable_sim_control);
 
+DECLARE_double(sim_control_start_velocity);
```
siyangy: How do you feed this into sim_control from replay_engine? Different scenario might have different start velocity.

timboilermaker: I add them in scenario proto so that we could obtain them from scenario conf.

siyangy: What about making them as arguments of Init function?

PR #1061, opened 2017-11-03
tools: updated mobileye viewer with auto drive status.

yifeijiang:


modules/tools/mobileye_viewer/plot_mobileye.py
```diff
@@ -25,19 +25,22 @@
...
+from chassis_data import ChassisData
 from view_subplot import ViewSubplot
 from subplot_s_speed import SubplotSSpeed
 from subplot_s_theta import SubplotSTheta
 from subplot_s_time import SubplotSTime
 from modules.localization.proto import localization_pb2
+from modules.canbus.proto import chassis_pb2
 
-PLANNING_TOPIC = '/apollo/planning'
+PLANNING_TOPIC = '/apollo/planning_lite'
```
Capri2014: Is it more easier to leave the topic as ''/apollo/planning'' as of now? We will not start the c++ one and this one the same time anyway.

PR #1095, opened 2017-11-06
planning: adjust dp path step size

startcode:


modules/planning/tasks/dp_poly_path/dp_poly_path_optimizer.cc
```diff
@@ -44,15 +44,15 @@ bool DpPolyPathOptimizer::Init(const PlanningConfig &config) {
 }
 
 Status DpPolyPathOptimizer::Process(const SpeedData &speed_data,
-                                    const ReferenceLine &reference_line,
+                                    const ReferenceLine &,
```
Capri2014: Line 47?

PR #1130, opened 2017-11-08
Dreamview: Implement changing vehicle with VehicleManager.

xiaoxq:


modules/dreamview/backend/hmi/vehicle_manager.cc
```diff
@@ -0,0 +1,144 @@
...
+
+  std::string result = src_path;
+  const auto pos = src_path.find(kRosPlaceholder);
+  if (pos != std::string::npos) {
+    result.replace(pos, kRosPlaceholder.length(), kRosRoot);
+  }
+  return result;
+}
+
+void CopyFileIfExists(const std::string &from, const std::string &to) {
```
startcode: can you move this function to common/util/file later? Thanks.

PR #1176, opened 2017-11-09
add usb camera driver

yibingl:
add usb camera dirver to catpure image for traffic lights detection


modules/drivers/usb_cam/AUTHORS.md
```diff
@@ -0,0 +1,9 @@
+Original Authors
+----------------
+
+ * [Benjamin Pitzer] (benjamin.pitzer@bosch.com)
```
xiaoxq: We don't include third party code into repo.
How about packing it to a library? Lei Wang could help on this.

PR #1189, opened 2017-11-10
Planning: use two attempts to solve qp_path -- solve the problem with…

lianglia-apollo:
…in the road first, if there is no solution, extend the road range and re-solve.


modules/planning/tasks/qp_spline_path/qp_spline_path_generator.cc
```diff
@@ -326,12 +333,12 @@ bool QpSplinePathGenerator::AddConstraint(
 
     if (evaluated_s_.at(i) - evaluated_s_.at(0) <
         qp_spline_path_config_.cross_lane_longitudinal_extension()) {
-      road_boundary.first =
-          std::fmin(road_boundary.first, init_frenet_point_.l() - lateral_buf);
-      road_boundary.second =
-          std::fmax(road_boundary.second, init_frenet_point_.l() + lateral_buf);
+      const double kRoadBoundaryBuff = 0.1;
```
Capri2014: Can you change the buffers into config in next iteration?

PR #1220, opened 2017-11-11
Add data-recorder to apollo/modules/data

yneofish:
Add data-recorder to apollo/modules/data

https://github.com/yneofish/apollo/tree/master/modules/data/data-recorder


modules/data/data-recorder/src/__main__.py
```diff
@@ -0,0 +1,23 @@
...
+################################################################################
+#
+# Copyright (c) 2016 Baidu.com, Inc. All Rights Reserved
+#
+################################################################################
+
+"""
+__main__.
+
+Authors: yuning(yuning@baidu.com)
```
Capri2014: Please correct the copywrite here.

modules/data/data-recorder/src/proto/recorder_info_pb2.py
```diff
@@ -0,0 +1,416 @@
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
Capri2014: Please do not checkin the generated file. Python proto would be generated via build_py_proto() function in ./scripts/apollo.sh, please refers to it for more details.

modules/data/data-recorder/conf/recorder.debug.yaml
```diff
@@ -0,0 +1,98 @@
+---
```
xiaoxq: Could you change config files to protobuf, as we do in other modules?
As a python example:

Proto: https://github.com/ApolloAuto/apollo/blob/master/modules/hmi/proto/config.proto
Config file: https://github.com/ApolloAuto/apollo/blob/master/modules/hmi/conf/config.pb.txt
To load: https://github.com/ApolloAuto/apollo/blob/master/modules/tools/common/proto_utils.py#L22

PR #1284, opened 2017-11-15
fix apollo.sh ros path error

fengqikai1414:


apollo.sh
```diff
@@ -510,8 +510,8 @@ function build_gnss() {
 
 function build_velodyne() {
   CURRENT_PATH=$(pwd)
-  if [ -d "${CURRENT_PATH}/bazel-apollo/external/ros" ]; then
-    ROS_PATH="${CURRENT_PATH}/bazel-apollo/external/ros"
+  if [ -d "/home/tmp/ros" ]; then
```
xiaoxq: How about use env variable? Current we have ${ROS_ROOT} which points to .../<ros>/share/ros.
So we can use ${ROS_ROOT}/../.. as ros path. Or add another ${ROS_HOME} to point to it directly.

The good thing is, it's transparent for dev, release and future changes.

fengqikai1414: fixed  with  ROS_ROOT

PR #1297, opened 2017-11-15
tools: added start script for navigation server, planning, and control.

yifeijiang:


scripts/navigation_control.sh
```diff
@@ -0,0 +1,26 @@
...
+###############################################################################
+
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+source "${DIR}/apollo_base.sh"
+
+# run function from apollo_base.sh
+# run command_name module_name
+run control "$@"
```
Capri2014: This might need to change, since control needs to load different configs here

scripts/navigation_planning.sh
```diff
@@ -0,0 +1,50 @@
...
+    echo "$0 start|stop|help"                                                             
+    ;;                                                                                              
+stop)                                                                                               
+    stop                                                                                       
+    ;;                                                                                              
+*)                                                                                                  
+    stop                                                                                           
+    start                                                                                           
+    ;;
+esac
```
Capri2014: This one looks similar to ./scripts/planning_lite.sh, so please consider deprecate one of them.

PR #1299, opened 2017-11-16
tools: added start script for navigation server and planning.

yifeijiang:


scripts/navigation_server.sh
```diff
@@ -0,0 +1,50 @@
...
+
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+source "${DIR}/apollo_base.sh"
+
+# run function from apollo_base.sh
+# run command_name module_name
+function start() {
+	echo "start"
```
xiaoxq: Plz fix indents?

PR #1311, opened 2017-11-16
navigation: changed server host.

yifeijiang:


modules/dreamview/frontend/dist/components/Navigation/navigation_viewer.html
```diff
@@ -84,7 +84,7 @@
             var json = '';
             $.ajax({
                 type: "POST",
-                url: "http://localhost:5001/",
+                url: "http://" + window.location.hostname + ":5001/",
```
vlin17: In case of using https, maybe replace "http:" with window.location.protocol, or detect both protocol and hostname together with window.location.origin?

yifeijiang: Currently navigation server and dreamview are not sharing same protocol. We might do that after migrating navigation server to dreamview.

PR #1361, opened 2017-11-20
Driver: remove excute permission from source and doc files

zhxt:
* also include LICENSE and .launch and .xml files


zhxt:
It is weird to make source and document-like files have excute permission.

And I'm not sure is it ok to do a update under the root of source directory with: 
find -type f -perm /111 -iregex ".*\.\(h\|cpp\|md\|txt\|rst\|launch\)\|.*LICENSE" -exec chmod a-x {} \;

command:
find -type f -perm /111 -iregex ".*\.\(h\|cpp\|md\|txt\|rst\|launch\)\|.*LICENSE" -exec ls -l {} \;
shows there are some files with the same issue.

PR #1399, opened 2017-11-22
planning: reduce default up speed limit to 35 mph.

ycool:


Capri2014:
why did we reduce the default speed back?

PR #1404, opened 2017-11-22
Fix matrix size in the notice

ixanezis:
Either I am completely mistaken and understand nothing, or the matrix always has the size (n+1)x(n+1) when representing the derivative of nth order polynom.


Capri2014:
@ixanezis lol, relax. You are right. That is a typo in the .md file and thanks for pointing this out.

PR #1409, opened 2017-11-22
navi: added routing smoother based on spline with constraints; also a…

yifeijiang:
…dded related debug tools.

### DIFF
--- a/modules/tools/navigation/planning/navigation_routing_debug.py
+++ b/modules/tools/navigation/planning/navigation_routing_debug.py
@@ -1,0 +1,162 @@
+#!/usr/bin/env python
+
+###############################################################################
+# Copyright 2017 The Apollo Authors. All Rights Reserved.
+#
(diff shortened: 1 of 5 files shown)


modules/tools/navigation/planning/navigation_routing_debug.py
```diff
@@ -0,0 +1,162 @@
...
+from provider_chassis import ChassisProvider
+from provider_localization import LocalizationProvider
+from provider_routing import RoutingProvider
+
+planning_pub = None
+routing_debug_pub = None
+
+PUB_NODE_NAME = "planning"
+PUB_TOPIC = "/apollo/" + PUB_NODE_NAME
+CRUISE_SPEED = 10  # m/s
```
Capri2014: line 43 ~ 45, three variables are defined but not used.

yifeijiang: first two variables are used, the cruise speed will be used soon.

PR #1441, opened 2017-11-27
Planning: fixed a safety bug in socket_can_client_raw.cc -- do not me…

lianglia-apollo:
…mcy when the returned can msg length is invalid.


modules/drivers/canbus/can_client/socket/socket_can_client_raw.cc
```diff
@@ -176,7 +176,13 @@ ErrorCode SocketCanClientRaw::Receive(std::vector<CanFrame> *const frames,
     }
     cf.id = recv_frames_[i].can_id;
     cf.len = recv_frames_[i].can_dlc;
-    std::memcpy(cf.data, recv_frames_[i].data, recv_frames_[i].can_dlc);
+    if (recv_frames_[i].can_dlc <= MAX_CAN_RECV_FRAME_LEN) {
```
Capri2014: Should compare it with "CAN_MAX_DLEN" in linux/can.h, which is defined as 8

Capri2014: FYI, https://www.kernel.org/doc/Documentation/networking/can.txt

PR #1484, opened 2017-11-29
navi: optimized planning algorithm.

yifeijiang:


modules/tools/navigation/planning/navigation_planning.py
```diff
@@ -40,7 +40,7 @@
 PUB_TOPIC = "/apollo/" + PUB_NODE_NAME
 CRUISE_SPEED = 10  # m/s
 ENABLE_FOLLOW = False
-ENABLE_ROUTING_AID = False
+ENABLE_ROUTING_AID = True
```
Capri2014: Do you want to enable it default true now ？

yifeijiang: Yes.

PR #1501, opened 2017-12-01
Monitor: Add camera-long and camera-short.

xiaoxq:


modules/common/adapters/adapter.h
```diff
@@ -90,6 +90,11 @@ class AdapterBase {
    */
   virtual double GetDelayInMs() const = 0;
 
+  /**
+   * @brief Gets message lag since last received message.
+   */
+  virtual double GetLagSinceLastMessage() const = 0;
```
siyangy: How’s this different with GetDelay?

xiaoxq: Actually the existing GetDelay() is a little weird to me.
It returns the interval between the last two messages. Which means, if a module worked fine but crashed suddenly, GetDelay() will always return a constant and look-good value.

siyangy: Sounds like a good idea, can you just replace the existing GetDelayInMs to fix the weird logic? I'd like to avoid confusingly having two methods getting delay/lag.

modules/common/adapters/adapter.h
```diff
@@ -86,9 +86,9 @@ class AdapterBase {
   virtual bool HasReceived() const = 0;
 
   /**
-   * @brief Gets message delay in milliseconds.
+   * @brief Gets message delay.
    */
-  virtual double GetDelayInMs() const = 0;
+  virtual double GetDelay() const = 0;
```
siyangy: can you include the time unit in function name to make it more prominent? As the downstream caller of the function might have different requirement of the time unit, like monitor vs dreamview.

PR #1535, opened 2017-12-03
Perception: Simple code cleaning for perception/common

Capri2014:


modules/perception/common/define.h
```diff
@@ -20,7 +20,6 @@
 namespace apollo {
 namespace perception {
 
-const double PI = 3.1415926535898;
 const double kRadianToDegree = 57.29577951308232;
```
startcode: do me one more favor:
change this one to:
constexpr double kRadianToDegree = 180.0 / M_PI;

Capri2014: @startcode  Thinking of retiring this define.h entirely since a lot of them are duplicates of modules/common

PR #1538, opened 2017-12-03
perception: fix compile warning and rename some long build target.

ycool:


modules/perception/obstacle/fusion/probabilistic_fusion/probabilistic_fusion_test.cc
```diff
@@ -28,9 +28,10 @@ namespace perception {
 TEST(ProbabilisticFusionTest, probabilistic_fusion_test) {
   FLAGS_work_root = "modules/perception";
   FLAGS_config_manager_path = "./conf/config_manager.config";
-  AINFO << "start probabilistic_fusion_test\n";
+  AERROR << "start probabilistic_fusion_test\n";
```
xiaoxq: Not error?

ycool: good point. Accidentally added the debug log. reverted in the new CL. thanks.

PR #1547, opened 2017-12-04
script: record more sensor data.

ycool:


scripts/record_bag_pnc.sh
```diff
@@ -0,0 +1,106 @@
...
+  LOG="/tmp/apollo_record.out"
+  NUM_PROCESSES="$(pgrep -c -f "rosbag record")"
+  if [ "${NUM_PROCESSES}" -eq 0 ]; then
+    nohup rosbag record -b 2048  \
+        /apollo/sensor/gnss/gnss_status \
+        /apollo/sensor/gnss/odometry \
+        /apollo/sensor/gnss/ins_stat \
+        /apollo/sensor/gnss/corrected_imu \
+        /apollo/sensor/mobileye \
+        /apollo/sensor/delphi_esr \
```
Capri2014: I refer to this line, do we still record delphi_esr by default?

ycool: yes.

PR #1677, opened 2017-12-08
perception: get radar extrinsics from file detectly

huiyujiang:
perception: get radar extrinsics from file detectly


xiaoxq:
My change has been merged, which will copy radar extrinsics to perception/data/params automatically.

huiyujiang:
perception: use new radar extrinstics link

Issue #1543, opened 2017-12-04
Could not open dreamview


gengqx:
I download the latest apollo,and i could not open dreamview,here is what I do
```
bash scripts/bootstrap.sh
tail /data/log/dreamview.out
```
and then I got
```
E1204 14:45:12.613981  8238 hdmap_util.cc:36] No existing file found in modules/map/data/sunnyvale_loop/base_map.bin|base_map.xml|base_map.txt. Fallback to first candidate as default result.
E1204 14:45:12.614317  8238 file.h:139] Failed to open file modules/map/data/sunnyvale_loop/base_map.bin in binary mode.
E1204 14:45:12.614334  8238 file.h:96] Failed to open file modules/map/data/sunnyvale_loop/base_map.bin in text mode.
E1204 14:45:12.614341  8238 hdmap_util.cc:62] Failed to load HDMap modules/map/data/sunnyvale_loop/base_map.bin
E1204 14:45:12.614367  8238 hdmap_util.cc:36] No existing file found in modules/map/data/sunnyvale_loop/sim_map.bin|sim_map.txt. Fallback to first candidate as default result.
E1204 14:45:12.614403  8238 file.h:139] Failed to open file modules/map/data/sunnyvale_loop/sim_map.bin in binary mode.
E1204 14:45:12.614419  8238 file.h:96] Failed to open file modules/map/data/sunnyvale_loop/sim_map.bin in text mode.
E1204 14:45:12.614425  8238 hdmap_util.cc:62] Failed to load HDMap modules/map/data/sunnyvale_loop/sim_map.bin
```

xiaoxq:
Did you run it within docker?

gengqx:
Yes @xiaoxq

gengqx:
I could open it use an older version,when the hmi.sh are still usefull,after it is update to bootstrap.sh,I could never open it.

xiaoxq:
Can you see the map data at "/apoll0/modules/map/data/sunnyvale_loop/"?

gengqx:
No,as you can see in the master of apollo
```
https://github.com/ApolloAuto/apollo/tree/master/modules/map/data
```
only a demo data with txt file.

xiaoxq:
The sunnyvale map is in docker image only, because it's too large to host in github repo.
So you can try starting a new container. Hopefully the data will appear.

gengqx:
The docker history is just like this:
```
dev-x86_64-20171103_2113
[...]

PR #1689, opened 2017-12-08
Update global_flagfile.txt

gengqx:
Because we could not find any data about sunnyvale_l
and this would cause that we could not open the dreamview,
So I think the default value should be demo in case that other people should debug with dreamview.

### DIFF
--- a/modules/common/data/global_flagfile.txt
+++ b/modules/common/data/global_flagfile.txt
@@ -7,4 +7,4 @@
 
 --vehicle_config_path=modules/common/data/mkz_config.pb.txt
 
---map_dir=modules/map/data/sunnyvale_loop
+--map_dir=modules/map/data/demo


xiaoxq:
sunnyvale_loop is so large that we have to host it in docker image. You should see it after starting the apollo_dev container.

ycool:
Can you check if your env has /home/tmp/modules_data/map/data/sunnyvale_loop/? More details can be found at https://github.com/ApolloAuto/apollo/issues/1543 ?

Thanks.

gengqx:
Yes,I could find the map,but it did not copy to the dev docker.

ycool:
the copy happens at 
https://github.com/ApolloAuto/apollo/blob/master/scripts/docker_adduser.sh#L53

Can you help to debug why the copy step doesn't happen in your env? 

Thanks.

ycool:
I am going to close this pull. please feel free to reopen or add more comments if more issues are found. Thanks.

PR #1720, opened 2017-12-11
planning: when rerouting, add unpassed waypoints.

startcode:


Capri2014:
@startcode The variable names are a bit confusing through.

PR #1747, opened 2017-12-13
perception: nonblocking fusion subnode event, use host velocity direc…

huiyujiang:
…tion when it is not static


modules/perception/obstacle/radar/modest/conti_radar_util.h
```diff
@@ -32,8 +32,8 @@ class ContiRadarUtil {
           const Eigen::Vector3f& velocity) {
     Eigen::Vector3f vector_temp1 = main_velocity;
     Eigen::Vector3f vector_temp2 = velocity;
-    if (vector_temp1.head(2).norm() > 1e-5
-        && vector_temp2.head(2).norm() > 1e-5) {
+    if (vector_temp1.head(2).norm() > 0.1
```
ycool: make 0.1 as a configure or gflag?

huburbub: Hi

PR #1766, opened 2017-12-14
Planning: restructured code so that we always calculate lane change f…

lianglia-apollo:
…irst. (We may still calculate others if prioritize_lane_change is false.


modules/planning/common/reference_line_info.h
```diff
@@ -96,10 +96,10 @@ class ReferenceLineInfo {
   std::string PathSpeedDebugString() const;
 
   /**
-   * Check if the current reference line is a change lane reference line, i.e.,
-   * ADC's current position is not on this reference line.
+   * Check if the current reference line is a change lane reference line.
    */
   bool IsChangeLanePath() const;
+  void SetIsChangeLanePath() { is_change_lane_path_ = true; }
```
startcode: recommend to use SetIsChangeLanePath(bool) to allow unset.

modules/planning/common/frame.cc
```diff
@@ -136,12 +136,25 @@ bool Frame::InitReferenceLineInfo() {
...
       AERROR << "Failed to init reference line";
       return false;
     }
   }
+
+  // set lane change path flag in reference line info
+  for_each(
+      reference_line_info_.begin(), reference_line_info_.end(),
+      [info_size](ReferenceLineInfo &info) {
+        if ((info_size == 2) && info.Lanes().NextAction() == routing::FORWARD) {
```
startcode: This may not true. when there are three parallel lanes, you are on the right most lane, and want to change to middle lane then change to left most lane, the middle lane's NextAction is LEFT

PR #1780, opened 2017-12-14
perception: change to use velocity of localization pose

huiyujiang:
perception: change to use velocity of localization pose

### DIFF
--- a/modules/perception/tool/export_sensor_data/export_sensor_data.cc
+++ b/modules/perception/tool/export_sensor_data/export_sensor_data.cc
@@ -269,7 +269,7 @@ bool ExportSensorData::GetCarLinearSpeed(double timestamp,
     distance = temp_distance;
   }
   const auto &velocity =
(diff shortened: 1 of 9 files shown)


modules/perception/tool/export_sensor_data/export_sensor_data.cc
```diff
@@ -238,34 +241,34 @@ bool ExportSensorData::GetCarLinearSpeed(double timestamp,
...
-    double temp_distance = fabs(timestamp - gps_buffer_[idx].first);
+    double temp_distance = fabs(timestamp - localization_buffer_[idx].first);
     if (temp_distance >= distance) {
       break;
     }
     distance = temp_distance;
   }
   const auto &velocity =
-      gps_buffer_[idx + 1].second.localization().linear_velocity();
+      localization_buffer_[idx + 1].second.pose().linear_velocity();
```
lianglia-apollo: It seems that [idx + 1] can cause a memory error ? (e.g., line 266 excuted during the first for loop)

PR #1845, opened 2017-12-18
tools: fix diagnostics error when default terminal width <= 80

zhxt:
Since I'm not familiar with Python language, I strongly recommend Python Master here to review this, thanks.


ycool:
Thanks a lot for your fix.

PR #1852, opened 2017-12-18
planning: fix a crash problem in case of sqrt negative number.

ycool:


modules/planning/tasks/path_decider/path_decider.cc
```diff
@@ -151,7 +151,7 @@ bool PathDecider::MakeStaticObstacleDecision(
                                           object_decision);
       }
     }
-  }
+}
```
lianglia-apollo: Is this correct?

ycool: good point. fixed

PR #1879, opened 2017-12-19
OTA: integrate security feature.

wanglei828:


modules/tools/ota/verify_client.py
```diff
@@ -0,0 +1,54 @@
+"""OTA verify client"""
```
Capri2014: !/usr/bin/env python

xiaoxq: Done.

PR #1883, opened 2017-12-19
Planning: speed up BuildStBoundary by using GetProject with some huer…

lianglia-apollo:
…istic range params, and using some tricks for early stop.

### DIFF
--- a/modules/planning/reference_line/reference_line.cc
+++ b/modules/planning/reference_line/reference_line.cc
@@ -409,18 +409,41 @@ bool ReferenceLine::IsOnRoad(const SLPoint& sl_point) const {
 // return a rough approximated SLBoundary using box length. It is guaranteed to
 // be larger than the accurate SL boundary.
 bool ReferenceLine::GetApproximateSLBoundary(
-    const common::math::Box2d& box, SLBoundary* const sl_boundary) const {
-  SLPoint center_sl_point;
-  if (!XYToSL(box.center(), &center_sl_point)) {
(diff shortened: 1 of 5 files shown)


modules/planning/reference_line/reference_line.cc
```diff
@@ -409,18 +409,21 @@ bool ReferenceLine::IsOnRoad(const SLPoint& sl_point) const {
...
     return false;
   }
 
-  sl_boundary->set_start_s(center_sl_point.s() - box.length() / 2.0);
-  sl_boundary->set_end_s(center_sl_point.s() + box.length() / 2.0);
-  sl_boundary->set_start_l(center_sl_point.l() - box.length() / 2.0);
-  sl_boundary->set_end_l(center_sl_point.l() + box.length() / 2.0);
+  sl_boundary->set_start_s(s - box.length() / 2.0);
+  sl_boundary->set_end_s(s + box.length() / 2.0);
+  sl_boundary->set_start_l(l - box.length() / 2.0);
```
startcode: box.width()?

startcode: or you can use box's heading to calculate the exact start_s, end_s, etc.

lianglia-apollo: It should be length because we need a box that can guarantee to cover thr actual box. 
We are unable to calculate the accurate start_s and end_s from he box’s heading, we also need to know the heading of the projected point on ref line. But it will double the calculation time.

PR #1910, opened 2017-12-20
Localization: Upgrade online visualization tool for localization.

yellowpoplar:
Localization: Upgrade online visualization tool for localization.


modules/localization/msf/local_tool/local_visualization/engine/visualization_engine.cc
```diff
@@ -176,10 +178,16 @@ void VisualizationEngine::Visualize(
...
+  if (!UpdateCarLocId(expected_car_loc_id_)) {
+    if (!UpdateCarLocId(car_loc_id_)) {
+      if (!UpdateCarLocId()) {
+        return;
+      }
+    } else {
+      if (expected_car_loc_id_ == loc_info_num_) {
+        expected_car_loc_id_ = car_loc_id_;
+      }
+    }
```
xiaoxq: Could you reduce these nested conditions? It's not easy to know which branch will be in effect.

PR #1940, opened 2017-12-21
Perception: add onboard sequence type fuser

xiaohuitu:


modules/perception/lib/base/mutex.h
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  *****************************************************************************/
 
-#ifndef apollo_PERCEPTION_LIB_BASE_MUTEX_H
-#define apollo_PERCEPTION_LIB_BASE_MUTEX_H
+#ifndef MODULES_PERCEPTION_LIB_BASE_MUTEX_H
```
lianglia-apollo: Please use google style (guard should be _H_)

lianglia-apollo: _H_

lianglia-apollo: I meant underscore H underscore (SOMETHING_H_). Some display issue.

PR #1953, opened 2017-12-21
Control : Change MPC hard constraint from vehicle configs

Capri2014:


Capri2014:
@lianglia-apollo Agree. Will definitely do.

modules/control/controller/mpc_controller.cc
```diff
@@ -229,8 +229,10 @@ Status MPCController::ComputeControlCommand(
     const canbus::Chassis *chassis,
     const planning::ADCTrajectory *planning_published_trajectory,
     ControlCommand *cmd) {
+  constexpr double min_speed_protection = 0.1;
```
lianglia-apollo: kAaaBbb

lianglia-apollo: https://google.github.io/styleguide/cppguide.html#Constant_Names

PR #1962, opened 2017-12-21
Control : Add feedforward term in MPC

Capri2014:


modules/control/controller/mpc_controller.cc
```diff
@@ -240,11 +240,15 @@ Status MPCController::ComputeControlCommand(
...
   debug->Clear();
 
+  ComputeLongitudinalErrors(&trajectory_analyzer_, debug);
+
   // Update state
   UpdateStateAnalyticalMatching(debug);
 
   UpdateMatrix();
 
+  FeedforwardUpdate(debug);
```
lianglia-apollo: It is not good to pass a debug as the function parameters in those major functions. You can refer planning's design (task/dp_poly_graph/dp_road_graph.h/cc) for details.

Capri2014: @lianglia-apollo Not serving the save purpose as planning do. But thanks for pointing out and will consider a redesign of the debug system later.

Capri2014: @lianglia-apollo Also to keep the code consistent with the reset of code as of now.

PR #1992, opened 2017-12-23
Planning: ignore far away obstacles when adc's decision is stop with …

lianglia-apollo:
…a closer distance.

### DIFF
--- a/modules/planning/common/path_decision.cc
+++ b/modules/planning/common/path_decision.cc
@@ -89,5 +89,44 @@ bool PathDecision::AddLongitudinalDecision(const std::string &tag,
   return true;
 }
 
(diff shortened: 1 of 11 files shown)


modules/planning/proto/decision.proto
```diff
@@ -111,6 +111,8 @@ message MainStop {
   // When stopped, the heading of the vehicle should be stop_heading.
   optional double stop_heading = 4;
   optional apollo.routing.ChangeLaneType change_lane_type = 5;
+  // the s of the stop point on reference line
+  optional double stop_reference_line_s = 6;
```
startcode: not recommend export relative info in planning's output.

PR #1995, opened 2017-12-23
tools: added a tool to calculate rosbag acc.

yifeijiang:


modules/tools/rosbag/stat_acc.py
```diff
@@ -0,0 +1,79 @@
...
+
+import sys
+import os
+import datetime
+
+from rosbag.bag import Bag
+
+
+kChassisTopic = '/apollo/canbus/chassis'
+kLocalizationTopic = '/apollo/localization/pose'
```
Capri2014: Looks like kLocalizationTopic and knano not needed?

modules/tools/rosbag/stat_acc.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def calculate(self, bag_file, log_file):
+        """Calculate and log acc."""
+        last_t = None
+        last_max_abs_acc = 0
+        last_speed = None
+        with Bag(bag_file, 'r') as bag:
+            for topic, msg, t in bag.read_messages(topics=[kChassisTopic]):
+                if last_t is not None:
+                    acc = (msg.speed_mps - last_speed) / (
```
Capri2014: Also not sure how accurate you want the acceleration to be, Lincoln can parse acceleration in chassis_details directly from chassis, you can refer to https://github.com/ApolloAuto/apollo/blob/master/modules/canbus/proto/chassis_detail.proto for more details.

modules/tools/rosbag/stat_acc.py
startcode: recommend use modules/tools/common/message_manager.py

modules/tools/rosbag/stat_acc.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def main():
+    """Main function."""
+    pgm_path = os.path.dirname(os.path.realpath(__file__))
+    now = datetime.datetime.now().strftime("%Y-%m-%d_%H.%M.%S")
+    log_file = open(pgm_path + "/" + now + ".txt", "w")
+
+    mc = AccCalculator()
+    for bag_file in sys.argv[1:]:
```
startcode: recommend to use argparse

PR #1997, opened 2017-12-23
planning: stop sign

jmtao:


modules/planning/common/planning_gflags.cc
```diff
@@ -239,8 +239,12 @@ DEFINE_bool(enable_rule_layer, true,
 /// common
 DEFINE_double(stop_max_distance_buffer, 4.0,
               "distance buffer of passing stop line");
-DEFINE_double(stop_min_speed, 0.1, "min speed for computing stop");
```
Capri2014: Is it possible to later change this flag to common so control can also share?

PR #2017, opened 2017-12-23
Planning: delayed change lane when a vehicle is nearby.

lianglia-apollo:


modules/planning/tasks/dp_poly_path/dp_road_graph.cc
```diff
@@ -305,6 +310,29 @@ bool DPRoadGraph::SamplePathWaypoints(
   return true;
 }
 
+bool DPRoadGraph::IsSafeForLaneChange() {
+  CHECK(reference_line_info_.IsChangeLanePath());
```
ycool: let's use if and log error instead of check crash.

PR #2031, opened 2017-12-24
scripts: add option to skip docker mirror prompt

zhxt:
* Also gives a chance to save 10 seconds for every automatically build.

`------------------------------`
This is a **Working In Progress** change, please do **NOT** merge.

Trying to add command line options to skip the interaction， 
this also can save 10 seconds for CI build. (read -t 10)

Also add an option: -image \<version\>  in order to compat with the original usage:
~~~
./docker/scripts/dev_start.sh dev-x86_64-20171221_1332-ci
~~~

And I'm not sure if it's okay to use the "-C" (China mirror) and "-nC" (not China mirror) option.
Any suggestions are welcome!


ycool:
Thanks a lot for the improvement. At present, it seems we only need "-C" (China mirror)  option. By default, we can use (not China mirror) option.

zhxt:
@ycool If we set default value for the variable INCHINA , then we have no chance/need to prompt. I will update for this.

zhxt:
Updated. And WIP sign was removed from issue title.

PR #2167, opened 2017-12-29
Planning: (1) used polynomial for stop profile; (2) a bug fix on traffic_decider/change_lane.cc

lianglia-apollo:


modules/planning/tasks/speed_optimizer.cc
```diff
@@ -101,6 +107,46 @@ SpeedData SpeedOptimizer::GenerateStopProfile(const double init_speed,
   return speed_data;
 }
 
+SpeedData SpeedOptimizer::GenerateStopProfileFromPolynomial(
+    const double init_speed, const double init_acc) const {
+  constexpr double kMaxT = 3.0;
+  for (double t = 2.0; t < kMaxT; t += 0.5) {
+    for (double s = 0.0; s < 50.0; s += 1.0) {
+      QuinticPolynomialCurve1d curve(0.0, init_speed, init_acc, s, 0.0, 0.0, t);
+      if (!IsValidProfile(curve)) {
```
YajiaZhang: It is possible that the velocity might be slightly below zero when the vehicle approaches to the destination. You might want to set that to zero or modify IsValidProfile function.

PR #2234, opened 2017-12-30
doc: fix typos

zhxt:


zhxt:
Another nitpick:  
the quality/resolution of the image which illustrates code snippets seems a little bit slow :P, not very clear.

zhxt:
As the document says:  "There are two important submodules **inside** perception. ", obstacle detection and traffic light detection.

Then, will it be more clearly to move the submodule rectangles of "obstacle detection" and "traffic light detection" **into** the "perception" rectangle in the architecture image?

Or maybe I was misunderstanding this.

PR #2263, opened 2017-12-31
Planning: used distance between reference lines instead of adc l to s…

lianglia-apollo:
…ample points. This way is more accurate and can avoid collision.


modules/planning/common/frame.cc
```diff
@@ -167,6 +167,25 @@ bool Frame::CreateReferenceLineInfo() {
     AERROR << "Failed to apply change lane decider";
     return false;
   }
+
+  if (reference_line_info_.size() == 2) {
```
Capri2014: @lianglia-apollo Do you really want to code as "== 2" ? Or should we re-write it so it can calculate distance between adjunct references lines when we have more than 2 reference lines available ?

lianglia-apollo: We will have at most two reference lines  ...

lianglia-apollo: It seems that multiple reference lines will be used in very advanced scenarios that far beyond our current capabilities. For example, when we need to decide if we need to change lane from left or right so that we can travel faster ... But it would require us to build a list of offsets that records every other reference lines offset. A little bit convoluted, with little gain ...

PR #2280, opened 2017-12-31
common::math: optimized HasOverlap calculations for box2d and polygon2d.

lianglia-apollo:


modules/common/math/box2d.h
```diff
@@ -236,15 +238,29 @@ class Box2d {
...
+  double min_y() const { return min_y_; }
+
  private:
   Vec2d center_;
   double length_ = 0.0;
   double width_ = 0.0;
   double half_length_ = 0.0;
   double half_width_ = 0.0;
-  double heading_ = 0.0;
+  double heading_ = 1.0;
```
startcode: why change default value to 1.0? 0.0 may help to indicate that it is not initialized.

PR #2301, opened 2018-01-02
Common & Canbus : Update drive state proto and publish from canbus

Capri2014:


modules/common/proto/drive_state.proto
```diff
@@ -9,8 +9,6 @@ message EngageAdvice {
         UNKNOWN = 0;
         DISALLOW_ENGAGE = 1;
         READY_TO_ENGAGE = 2;
-        KEEP_ENGAGED = 3;
```
ycool: let's keep these two types?

Capri2014: @ycool Done.

PR #2329, opened 2018-01-03
Dreamview: Fix frontend error unable to set visible of undefined

siyangy:


siyangy:
@unacao This fixed the speed display error as well, not sure why the stop reason was absent in the first place though.

PR #2358, opened 2018-01-04
Planning: added multi-thread for dp_st_graph.

lianglia-apollo:


modules/planning/common/planning_gflags.cc
```diff
@@ -331,6 +331,9 @@ DEFINE_bool(enable_sqp_solver, true, "True to enable SQP solver.");
 
 DEFINE_bool(
     enable_multi_thread_in_dp_poly_path, false,
-    "Enable multiple thread to calculation curve cost in dp_poly_graph.");
-DEFINE_int32(num_thread_dp_poly_path, 3,
-             "num of thread used in dp_poly_graph.");
+    "Enable multiple thread to calculation curve cost in dp_poly_path.");
```
Capri2014: threadpool for dp_poly_path defined but not used?

lianglia-apollo: it is used in dp_poly_path

PR #2363, opened 2018-01-04
Dreamview: updated offline websocket to reflect to the new sim_world format

vlin17:


modules/dreamview/frontend/src/store/websocket/websocket_offline.js
```diff
@@ -128,16 +128,17 @@ export default class OfflinePlaybackWebSocketEndpoint {
     }
 
     processSimWorld(message) {
-        if (STORE.playback.shouldProcessFrame(message.world)) {
+        const world = JSON.parse(message.world);
```
siyangy: we should probably extract out common code of websocket_offline and websocket_ros so that we don't have to change the same thing twice..

PR #2371, opened 2018-01-04
Tools: Stop scanning when finding vehicle name.

xiaoxq:


modules/tools/rosbag/stat_static_info.py
```diff
@@ -45,11 +46,9 @@ def process_file(self, bag_file):
...
         """
-        if not os.path.isfile(bag_file):
-            return False
         try:
             with Bag(bag_file, 'r') as bag:
-                for _, msg, _ in bag.read_messages(topics=[kStaticInfoTopic]):
+                for _, msg, _ in bag.read_messages(topics=kTopics):
                     if msg.vehicle.name:
                         self.vehicle_name = msg.vehicle.name.lower()
                     if msg.vehicle.license.vin:
```
Capri2014: vin already written in kv_db, no longer recorded into bag.

PR #2373, opened 2018-01-05
Prediction: add digital filter for cyclist and pedestrian heading

kechxu:


modules/prediction/container/obstacles/obstacle.cc
```diff
@@ -50,6 +50,10 @@ double Damp(const double x, const double sigma) {
 
 }  // namespace
 
+Obstacle::Obstacle()
```
Capri2014: Better add a >0 protection.

PR #2391, opened 2018-01-08
Localization: update doc and fix markdown syntax

zhxt:


zhxt:
Sounds still not that fluency :-(

Issue #2402, opened 2018-01-09
can_sender_ is redefined in class LincolnController (Lincoln_Controller.h)


jilinzhou:
The can_sender_ is first declared as a protected member variable in its base class VehicleController as the following:
```
protected:
  canbus::VehicleParameter params_;
  CanSender<::apollo::canbus::ChassisDetail> *can_sender_ = nullptr;
```
and then re-declared in the derived class LincolnController as a private variable as the following:
```
private:
  // control protocol
  Brake60 *brake_60_ = nullptr;
  Throttle62 *throttle_62_ = nullptr;
  Steering64 *steering_64_ = nullptr;
  Gear66 *gear_66_ = nullptr;
  Turnsignal68 *turnsignal_68_ = nullptr;

  CanSender<::apollo::canbus::ChassisDetail> *can_sender_ = nullptr;
```

Here is a simple test to show the potential problem.
```
#include <cstdlib>
#include <iostream>

class A {
public:

	A() = default;

	~A() = default;

	virtual void setA(int a) {this->a = a;}

	virtual int getA() {return a;}

protected:
	int a;
};

class B final : public A {
public:
	B() = default;
	~B() = default;

	void setA(int a) override {this->a = a + 1;}

//	int getA() override {return a;}

private:
	int a;
};

int main(int argc, char *argv[]) {
	std::cout << "Welcome to the Momentics IDE" << std::endl;

	B b = B();
	b.setA(3);

	std::cout << "b.a = " << b.getA() << std::endl;
	return EXIT_SUCCESS;
}
```
On my target, it returns b.a = 0 instead of 4 as expected. Simply removing the re-declaration of can_sender_  should fix the issue.

lianglia-apollo:
It seems can_sender_ is unnecessarily re-declared. Please take a look at the simple fix: https://github.com/ApolloAuto/apollo/pull/2404

Capri2014:
Thanks for pointing this out. The fix has been merged.

PR #2404, opened 2018-01-09
Canbus: removed the redundant definition of can_sender_ in LincolnCon…

lianglia-apollo:
…troller.


lianglia-apollo:
fix based on #2402

PR #2411, opened 2018-01-10
Planning: stop thread pool when planning stops. Control: fixed a timer error

lianglia-apollo:


modules/control/control.cc
```diff
@@ -89,7 +89,7 @@ Status Control::Start() {
   // need to sleep, because advertised channel is not ready immediately
   // simple test shows a short delay of 80 ms or so
   AINFO << "Control resetting vehicle state, sleeping for 1000 ms ...";
-  usleep(1000 * 1000);
+  std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
startcode: The log in 91 says sleep for 1000 ms

lianglia-apollo: 1 milli second == 1000 micro seconds

PR #2481, opened 2018-01-16
Update documentation for the Apollo 2.0

richtong:
This updates the documentation to make it clear you need to do a build and also explains how builds work without a GPU. Also ignores the demo bag in the docs directory as you should not check that in.

Also makes it clear you need a `download` verb to the rosbag downloader


richtong:
Thank u qi Luo!

Thanks,
Rich

Sent from my iPhone X
http://tongfamily.com | @richtong | https://fb.me/richtong1

PR #2550, opened 2018-01-22
Docs: rename doc file to be a valid markdown link

zhxt:


zhxt:
This will fix the broken link [here](https://github.com/zhxt/apollo/blob/doc-polish/docs/quickstart/README.md).

PR #2562, opened 2018-01-23
planning: refactor trajectory_cost

startcode:


modules/planning/tasks/dp_poly_path/trajectory_cost.cc
```diff
@@ -137,37 +136,37 @@ ComparableCost TrajectoryCost::CalculatePathCost(
...
     path_cost += l * l * config_.path_l_cost() * quasi_softmax(std::fabs(l));
 
     double left_width = 0.0;
     double right_width = 0.0;
-    reference_line_->GetLaneWidth(path_s, &left_width, &right_width);
-
-    const auto &vehicle_config =
-        common::VehicleConfigHelper::instance()->GetConfig();
-    const double width = vehicle_config.vehicle_param().width();
+    reference_line_->GetLaneWidth(curve_s, &left_width, &right_width);
```
lianglia-apollo: curve_s should be curve_s + start_s ?

lianglia-apollo: Maybe it is better to keep curve_s in this change and fix it later, so we know nothing affects the tests.

startcode: That's the plan

PR #2568, opened 2018-01-23
Planning: move plan precedure into planners themselves, because different planners have different logics

kechxu:


modules/planning/planner/em/em_planner.h
```diff
@@ -59,6 +59,16 @@ class EMPlanner : public Planner {
 
   common::Status Init(const PlanningConfig& config) override;
 
+  /**
+   * @brief Overrode function Plan in parent class Planner.
```
lianglia-apollo: override

modules/planning/planner/em/em_planner.cc
```diff
@@ -138,6 +138,38 @@ void EMPlanner::RecordDebugInfo(ReferenceLineInfo* reference_line_info,
   ptr_stats->set_time_ms(time_diff_ms);
 }
 
+Status EMPlanner::Plan(const TrajectoryPoint& planning_start_point,
```
lianglia-apollo: I suggest make some name changes on this function if we move them to the same class. Two "Plan" functions are doing different things. Keeping same name would make developers hard to understand.

PR #2587, opened 2018-01-24
fix problems found when compile for other platforms

startcode:


apollo.sh
```diff
@@ -99,6 +99,10 @@ function generate_build_targets() {
   if ! $USE_ESD_CAN; then
      BUILD_TARGETS=$(echo $BUILD_TARGETS |tr ' ' '\n' | grep -v "esd")
   fi
+  #skip msf for non x86_64 platforms
```
Capri2014: You probably want to update README to remind ppl to not change localization settings to RTK for non-x86 platforms.

PR #2650, opened 2018-01-30
Control : Update debug protos

Capri2014:


modules/control/proto/control_cmd.proto
```diff
@@ -127,13 +127,18 @@ message SimpleMPCDebug {
...
   optional double steer_angle_limited = 17;
   optional double station_reference = 18;
-  optional double station_error = 19;
-  optional double speed_reference = 20;
-  optional double speed_error = 21;
-  optional double acceleration_reference = 22;
-  optional bool is_full_stop = 23;
-  optional double station_feedback = 24;
-  optional double speed_feedback = 25;
+  optional double station_feedback = 19;
```
ycool: let's not change existing field proto number (except if it is fully inside one module)

Capri2014: @ycool The debug proto is only for diagnosis purpose as of now, do we also need to keep the proto?

PR #2658, opened 2018-01-30
dreamview: add voice command announcement.

ycool:


xiaoxq:
This uses speaker on each laptop/iPad, which might not be loud enough.
Would it be better to call /apollo/car_sound to use vehicle speaker?

Capri2014:
@ycool @xiaoxq Do we use car_sound or this one? If not intention to use car_sound, I may retire that tool then.

PR #2703, opened 2018-02-01
dreamview: added performance profiling mode

startcode:


modules/dreamview/backend/common/dreamview_gflags.cc
```diff
@@ -18,6 +18,12 @@
 
 DEFINE_string(dreamview_module_name, "dreamview", "dreamview module name");
 
+DEFINE_bool(dreamview_test_mode, false, "Run dreamview in test mode");
```
siyangy: better to call this profiling mode to distinguish from unit test? same for the rest.

PR #2726, opened 2018-02-02
planning: (1)add a test on keep clear zone; (2)rename keep_clear back…

jmtao:
… since it is a dmv term


modules/planning/common/planning_gflags.cc
```diff
@@ -250,10 +250,10 @@ DEFINE_double(max_valid_stop_distance, 3.0,
 DEFINE_double(creep_stop_distance, 0.5,
               "stop distance(m) to the stop line of next lane overlap "
               "while creeping ");
-/// clear_area
-DEFINE_bool(enable_clear_area, true, "enable clear_area");
-DEFINE_string(clear_area_virtual_object_id_prefix, "CA_",
-              "prefix for converting clear_area id to virtual object id");
+/// keep_clear
+DEFINE_bool(enable_keep_clear, true, "enable keep clear zone");
+DEFINE_string(keep_clear_virtual_object_id_prefix, "CA_",
```
ycool: s/CA_/KC_/ ?

PR #2733, opened 2018-02-03
allow loading of reflection background image

unacao:


modules/dreamview/frontend/src/renderer/ground.js
```diff
@@ -23,20 +23,71 @@ export default class Ground {
...
-        const { xres, yres, mpp } = PARAMETERS.ground;
+        if (!this.render(coordinates)) {
+            return false;
+        }
+
+        this.initialized = true;
+        return true;
+    }
+
+    update(world, coordinates, scene) {
```
siyangy: is this function still useful?

modules/dreamview/frontend/src/renderer/ground.js
```diff
@@ -23,20 +23,71 @@ export default class Ground {
...
+    }
+
+    updateImage(coordinates, mapName) {
+        if (!this.mesh) {
+            return;
+        }
+        const dir = this.titleCaseToSnakeCase(mapName);
+        /*import(`assets/map_data/${dir}/background.jpg`)
```
siyangy: is it okay to remove commented code?

modules/dreamview/frontend/src/renderer/ground.js
```diff
@@ -23,20 +23,71 @@ export default class Ground {
...
+        .then(ground => {
+            console.log("updating ground image with " + dir);
+            loadTexture(ground, texture => {
+                this.mesh.material.map = texture;
+                this.render(coordinates, dir);
+            });
+        })*/
+
+        const server = 'http://' + window.location.hostname + ':8888';
+        const imgUrl = server + '/assets/map_data/' + dir + '/background.jpg';
```
siyangy: I think the server part can be omitted?

modules/dreamview/frontend/src/renderer/ground.js
unacao: It is still called by render function of the main index

modules/dreamview/frontend/src/renderer/ground.js
unacao: Surely ok to remove. Was trying to leave a working code sample for dynamic import

modules/dreamview/frontend/src/renderer/ground.js
unacao: Server port needs to be fixed as the dist always serves from 8888 no matter dev or prod mode

PR #2735, opened 2018-02-03
GNSS: removed un-used gnss_status.proto (it is duplicated from common…

lianglia-apollo:
…/proto/gnss_status.proto).


ycool:
we need be careful to remove this proto. gnss driver is not built from bazel yet.

PR #2736, opened 2018-02-03
Drivers & Tools : Unify GNSS proto and provide temp tools to rewrite the topic

Capri2014:


Capri2014:
removed topic rewrite tools

PR #2752, opened 2018-02-05
planning: added reference line frame converter

YajiaZhang:


modules/planning/lattice/util/reference_line_frame_converter.cc
```diff
@@ -0,0 +1,44 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
ycool: s/2017/2018/

modules/planning/lattice/util/reference_line_frame_converter.h
```diff
@@ -0,0 +1,44 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
ycool: s/2017/2018/

modules/planning/lattice/util/reference_line_frame_converter.cc
YajiaZhang: Right. Done.

PR #2761, opened 2018-02-06
disable reflection map image on mobile devices

unacao:


modules/dreamview/frontend/src/renderer/ground.js
```diff
@@ -7,6 +7,7 @@ import PARAMETERS from "store/config/parameters.yml";
 export default class Ground {
     constructor() {
         this.type = "default";
+        this.gridOnly = false;
         this.loaded_map = null;
```
siyangy: Let's make the two members in camel case just for consistency

PR #2790, opened 2018-02-07
Localization: Try to load utm zone id from config file in local map folder

zhouyao4321:


modules/localization/msf/msf_localization.cc
```diff
@@ -168,8 +170,19 @@ void MSFLocalization::InitParams() {
...
+  // try load zone id from config file in local_map folder
+  if (FLAGS_if_utm_zone_id_from_file) {
+    bool success = LoadZoneIdFromFile(localizaiton_param_.map_path,
+                                      FLAGS_local_map_config_file,
+                                      &localizaiton_param_.utm_zone_id);
+    if (!success) {
+      AERROR << "Can't load utm zone id from config file, use default value.";
+    }
+  }
+  AERROR << "utm zone id: " << localizaiton_param_.utm_zone_id;
```
ycool: s/AERROR/AINFO/

modules/localization/msf/msf_localization.cc
```diff
@@ -16,6 +16,8 @@
 
 #include "modules/localization/msf/msf_localization.h"
 
+#include <boost/property_tree/ptree.hpp>
```
ycool: let's use protocol buffer text version of config. one example: https://github.com/ApolloAuto/apollo/blob/master/modules/localization/conf/localization_config.pb.txt

modules/localization/common/localization_gflags.cc
```diff
@@ -65,6 +65,9 @@ DEFINE_bool(enable_gps_timestamp, false,
 
 // msf parame
 DEFINE_string(local_map_name, "local_map", "The path of localization map.");
+DEFINE_string(local_map_config_file,
```
ycool: Let's use protobuf text instead of xml. We'd better keep config format consistent. Thanks.

PR #2796, opened 2018-02-07
Prediction: add downsample and merge h5

kechxu:


modules/tools/prediction/mlp_train/mlp_train.py
```diff
@@ -245,11 +278,13 @@ def save_model(model, param_norm, filename):
...
 
+    X_test = (X_test - param_norm[0]) / param_norm[1]
+
     model = setup_model()
 
     model.fit(X_train, Y_trainc,
         shuffle = True,
-        nb_epoch = 30,
+        nb_epoch = 20,
         batch_size = 32)
```
jinghaomiao: Use some macro or global variables at the head of the file so that it's easier for users to adjust these parameters.

PR #2804, opened 2018-02-08
 Accomodate clang build: remove uint abs, and add numeric header inclusion

yulongzhang:


modules/localization/msf/local_map/base_map/base_map_node.cc
```diff
@@ -111,12 +111,12 @@ bool BaseMapNode::Save() {
   if (!EnsureDirectory(path)) {
     return false;
   }
-  snprintf(buf, sizeof(buf), "/%08u", abs(index_.m_));
+  snprintf(buf, sizeof(buf), "/%08u", index_.m_);
```
startcode: can you try std::abs or abs defined in <cmath>?
I found a hint here: https://stackoverflow.com/questions/29136607/c-what-is-wrong-with-abs
Thanks.

yulongzhang: Clang complained about "ambiguous invocation". The root cause is that we are performing abs on UNSIGNED INT, where C++ only defines abs for SIGNED types. I verified that adding "std:" does not solve the problem.
Regardless of the compiler complaint, performing abs on non-negative numbers is meaningless. So probably it's better to remove it?

startcode: sounds good.

PR #2809, opened 2018-02-08
planning: move virtual obstacle creation to central place (frame)

jmtao:


modules/planning/common/frame.cc
```diff
@@ -195,29 +195,80 @@ bool Frame::CreateReferenceLineInfo() {
...
-  return AddStaticVirtualObstacle(object_id, stop_wall_box);
+  return CreateStaticVirtualObstacle(obstacle_id, stop_wall_box);
 }
 
-const Obstacle *Frame::AddStaticVirtualObstacle(const std::string &id,
-                                                const Box2d &box) {
+/**
+ * @brief: create static virtual object with lane width,
+ */
+const Obstacle *Frame::CreateVirtualStaticObstacle(
```
startcode: make function names consistent.   CreateStateVirtualObstacle at line 217, but another function called CreateVirutalStaticObstacle at line 223.

jmtao: Changed CreateStateVirtualObstacle() at line 217 to CreateStateObstacle(). This is a public method used to create a static obstacle. 

CreateVirutalStaticObstacle() at line 223 is a private method used to by all obstacle creation method (stop wall,  static obstacle and etc).  The name specify virtual + static.

PR #2833, opened 2018-02-10
display point cloud in dreamview

unacao:


modules/dreamview/backend/point_cloud/point_cloud_updater.cc
```diff
@@ -0,0 +1,90 @@
...
+
+void PointCloudUpdater::UpdatePointCloud(const PointCloud2 &point_cloud) {
+  last_receive_time_ = Clock::NowInSeconds();
+  // transform from ros to pcl
+  pcl::PointCloud<pcl::PointXYZ> pcl_data;
+  pcl::fromROSMsg(point_cloud, pcl_data);
+
+  point_cloud_.Clear();
+  int count = 0;
+  for (size_t idx = 0; idx < pcl_data.size(); idx += 4) {
```
siyangy: Let's make this a const member to be more prominent.

modules/dreamview/backend/point_cloud/point_cloud_updater.cc
```diff
@@ -0,0 +1,90 @@
...
+    pcl::PointXYZ& pt = pcl_data.points[idx];
+    if (!isnan(pt.x) && !isnan(pt.y) && !isnan(pt.z)) {
+      count++;
+      point_cloud_.add_num(pt.x);
+      point_cloud_.add_num(pt.y);
+      point_cloud_.add_num(pt.z + 1.91);
```
siyangy: Are we able to get the magic number from some config? Different car model would have different height.

modules/tools/supervisord/dev.conf
```diff
@@ -116,7 +116,7 @@ redirect_stderr=true
 stdout_logfile=/apollo/data/log/dreamview.out
 
 [program:sim_control]
-command=/apollo/bazel-bin/modules/dreamview/dreamview --flagfile=/apollo/modules/dreamview/conf/dreamview.conf --enable_sim_control --log_dir=/apollo/data/log
+command=/apollo/bazel-bin/modules/dreamview/dreamview --flagfile=/apollo/modules/dreamview/conf/dreamview.conf --log_dir=/apollo/data/log
```
siyangy: this section can be removed as a whole, as well as in release.conf.

modules/dreamview/backend/point_cloud/point_cloud_updater.cc
```diff
@@ -0,0 +1,90 @@
...
+        {
+          boost::shared_lock<boost::shared_mutex> reader_lock(mutex_);
+          to_send = point_cloud_str_;
+        }
+        websocket_->SendBinaryData(conn, to_send, true);;
+      });
+}
+
+void PointCloudUpdater::Start() {
+  AdapterManager::AddPointCloudCallback(&PointCloudUpdater::UpdatePointCloud,
```
siyangy: Do we need to check whether the PointCloudAdapter is present?

modules/dreamview/backend/point_cloud/point_cloud_updater.cc
Capri2014: @unacao @siyangy Is that case put it in https://github.com/ApolloAuto/apollo/blob/master/modules/common/configs/proto/vehicle_config.proto maybe? This is where we usually load vehicle related configs.

siyangy: looks like it's from velodyne64_height.yaml, not sure we have a programatic way to read it though.

PR #2855, opened 2018-02-13
planning: ignore out of time range obstacle.

ycool:


modules/planning/tasks/speed_decider/speed_decider.cc
```diff
@@ -132,7 +132,8 @@ Status SpeedDecider::MakeObjectDecision(
     const auto& boundary = path_obstacle->st_boundary();
 
     if (boundary.IsEmpty() || boundary.max_s() < 0.0 ||
-        boundary.max_t() < 0.0) {
+        boundary.max_t() < 0.0 ||
+        boundary.min_t() > dp_st_speed_config_.total_time()) {
```
lianglia-apollo: Actually, we can shrink the range to the last point's time in speed_profile?

PR #2894, opened 2018-02-15
Dreamview: integrating gmap navigation

vlin17:


modules/dreamview/frontend/gen_pbjs.sh
```diff
@@ -16,14 +16,16 @@
 # limitations under the License.
 ###############################################################################
 
+set -x
```
siyangy: remove set -x?

PR #2906, opened 2018-02-15
Control : Code cleaning and add more tests

Capri2014:


modules/control/integration_tests/mpc_controller_test.cc
```diff
@@ -0,0 +1,58 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
lianglia-apollo: 2018

Capri2014:
@lianglia-apollo  Done

PR #2929, opened 2018-02-21
Dreamview: handle fragmented message from websocket

vlin17:
1. handled fragmented message from websocket
2. updated destination of "TO Cananda West"
3. moved navigation's js files from utils to components


modules/dreamview/backend/handlers/websocket.cc
```diff
@@ -144,29 +144,50 @@ bool WebSocketHandler::SendData(Connection *conn, const std::string &data,
...
+        break;
+      case WEBSOCKET_OPCODE_BINARY:
+        result = handleBinaryData(conn, WebSocketHandler::data_.str());
+        break;
+      default:
+        AERROR << "Unknown WebSocket bits flag: " << bits;
+        break;
+    }
+
+    // rest opcode and data
```
siyangy: reset

modules/dreamview/backend/handlers/websocket.cc
```diff
@@ -144,29 +144,50 @@ bool WebSocketHandler::SendData(Connection *conn, const std::string &data,
...
+    WebSocketHandler::current_opcode_ = 0x00;
+    WebSocketHandler::data_.clear();
+    WebSocketHandler::data_.str(std::string());
```
siyangy: I guess just .clear() is enough? And data_ instead of WebSocketHandler::data_?

modules/dreamview/backend/handlers/websocket.cc
vlin17: will update. thanks for catching it.

modules/dreamview/backend/handlers/websocket.cc
vlin17: .clear() is used to clear the error state of the stream. Needs str(std::string()) to empty the stream.

PR #2953, opened 2018-02-22
Dreamview: downsample prediction points

siyangy:


modules/dreamview/backend/simulation_world/simulation_world_service.cc
```diff
@@ -813,10 +812,18 @@ void SimulationWorldService::CreatePredictionTrajectory(
   for (const auto &traj : obstacle.trajectory()) {
     Prediction *prediction = world_object->add_prediction();
```
lianglia-apollo: The function CreatePredictionTrajectory has a wrong order for input/output variables.

modules/dreamview/backend/simulation_world/simulation_world_service.cc
```diff
@@ -853,10 +860,8 @@ void SimulationWorldService::UpdateSimulationWorld(
...
     // Downsample the path points for frontend display.
-    // Angle threshold is about 5.72 degree.
-    constexpr double angle_threshold = 0.1;
-    std::vector<int> sampled_indices =
-        DownsampleByAngle(path.path_points(), angle_threshold);
+    auto sampled_indices =
+        DownsampleByAngle(path.path_points(), kAngleThreshold);
 
     route_paths_.emplace_back();
     RoutePath *route_path = &route_paths_.back();
```
lianglia-apollo: route_paths_ is a vector, the pointer address can change.

modules/dreamview/backend/simulation_world/simulation_world_service.cc
siyangy: done

modules/dreamview/backend/simulation_world/simulation_world_service.cc
siyangy: not sure I get this?

startcode: there will be a problem only when receiving two routing_responese simultaneously, which is unlikely for routing topic.

PR #2954, opened 2018-02-22
Drivers : unify the radar proto output

Capri2014:
proposed unified radar proto output for perception/third_party_perception


Capri2014:
@eugenegx  please also review

PR #2956, opened 2018-02-22
relative_map: added navigation lane config

startcode:


startcode:
I think `min_lane_marker_quality` is more precise than `QualityThreshold`.

PR #2969, opened 2018-02-23
common: missing error checking in method KVDB::Has() and KVDB::Get()

javeme:
Except `NotFound` error, there may be other errors when accessing `leveldb::DB::Get()`,
such as `IOError`.


modules/common/kv_db/kv_db.cc
```diff
@@ -86,7 +86,8 @@ bool KVDB::Has(const std::string &key) {
 
   std::string value;
   const auto status = GetDB()->Get(options, key, &value);
-  return !status.IsNotFound();
+  CHECK(status.ok() || status.IsNotFound()) << status.ToString();
```
xiaoxq: No need to CHECK. Even no need to show error.
Both Found and NotFound should be normal for a function named "Has()".

modules/common/kv_db/kv_db.cc
```diff
@@ -95,6 +96,7 @@ std::string KVDB::Get(const std::string &key,
 
   std::string value;
   const auto status = GetDB()->Get(options, key, &value);
+  CHECK(status.ok() || status.IsNotFound()) << status.ToString();
```
xiaoxq: I prefer "AERROR_IF(status.ok() || !default_value.empty())".
Only show error if the key doesn't exist and no default value is set. Even this case is sometimes the expected behavior.
We could add another API named "GetOrDie(key)" to do CHECK to guarantee existence.

xiaoxq:
Oops. I got it wrong.

PR #2999, opened 2018-02-24
Localizaiton: localizaiton result add vehicle orientation

zhouyao4321:


modules/localization/msf/msf_localization.cc
```diff
@@ -285,6 +316,22 @@ void MSFLocalization::OnRawImu(const drivers::gnss::Imu &imu_msg) {
...
+
+      Eigen::Quaternion<double> quat_vehicle_world =
+          quaternion * imu_vehicle_quat_;
+      apollo::common::Quaternion* orientation_vehicle_world =
+          posepb_loc->mutable_orientation_vehicle_world();
+      orientation_vehicle_world->set_qx(quat_vehicle_world.x());
+      orientation_vehicle_world->set_qy(quat_vehicle_world.y());
+      orientation_vehicle_world->set_qz(quat_vehicle_world.z());
+      orientation_vehicle_world->set_qw(quat_vehicle_world.w());
+
```
Capri2014: Should the optional apollo.common.Point3D euler_angles  be updated here as well? My understanding is that this one is still representing the IMU angles, without the IMU to vehicle corrections yet. Please correct me if I am wrong. Or the alternative question is: with this one, what is the usage example for euler_angle? Thanks.

PR #3000, opened 2018-02-25
Fix an estop bug in planning.cc

davidhopper2003:
Because the function "Control::ProduceControlCommand()" checks the "estop" signal with the following line (Line 170 in control.cc): estop_ = estop_ || trajectory_.estop().is_estop();
we should add more information to ensure the estop being triggered.


Capri2014:
@davidhopper2003 That is intentionally left un-implemented in my understanding, @startcode @lianglia-apollo correct me if I am wrong.

davidhopper2003:
@Capri2014, I don't think so. For the sake of   safety, it should be implemented.

PR #3067, opened 2018-02-28
add calibration manager

weidezhang:


modules/perception/lib/base/singleton.h
```diff
@@ -0,0 +1,67 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
jzhuucla: It should be "Copyright 2018 The Apollo Authors. All Rights Reserved."

PR #3083, opened 2018-02-28
Dreamview: implemented map element toggles in layer menu

vlin17:


modules/dreamview/frontend/src/renderer/map.js
```diff
@@ -472,10 +481,27 @@ export default class Map {
...
+        // Ex: mapping 'lane' to 'showMapLane' option
+        const optionName = `showMap${kind[0].toUpperCase()}${kind.slice(1)}`;
+
+        // NOTE: return true if the option is not found
+        return STORE.options[optionName] !== false;
+    }
+
     updateIndex(hash, elementIds, scene) {
-        if (hash !== this.hash) {
+        let newElementKindsDrawn = '';
```
siyangy: let's change this to a set

PR #3103, opened 2018-03-01
Navigation: fixed a bug for projection index, reordered the sequence …

lianglia-apollo:
…of navigation and lane marker, used combination of navigation and lane marker.

### DIFF
--- a/modules/map/relative_map/navigation_lane.cc
+++ b/modules/map/relative_map/navigation_lane.cc
@@ -52,8 +52,13 @@ bool NavigationLane::GeneratePath() {
   auto* path = navigation_path_.mutable_path();
   const auto& lane_marker = perception_obstacles_.lane_marker();
 
-  // priority: navigation line > perception lane marker
-  if (navigation_info_.navigation_path_size() > 0) {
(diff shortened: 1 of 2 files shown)


modules/map/relative_map/navigation_lane.cc
```diff
@@ -73,6 +78,78 @@ double NavigationLane::EvaluateCubicPolynomial(const double c0, const double c1,
...
+  int lane_marker_index = 0;
+  for (double s = 0.0; s < len; s += ds) {
+    auto p1 = GetPathPointByS(navigation_path, navigation_index, s,
+                              &navigation_index);
+    auto p2 = GetPathPointByS(lane_marker_path, lane_marker_index, s,
+                              &lane_marker_index);
+    auto* p = path->add_path_point();
+    p->set_x((p1.x() + p2.x()) / 2.0);
+    p->set_y((p1.y() + p2.y()) / 2.0);
+    p->set_theta((p1.theta() + p2.x()) / 2.0);
```
ycool: p2.x() => p2.theta() ?

PR #3121, opened 2018-03-02
planning: move traffic decider out of em planner

startcode:


modules/planning/common/frame.cc
```diff
@@ -191,8 +191,16 @@ bool Frame::CreateReferenceLineInfo() {
     reference_line_info_.back().SetOffsetToOtherReferenceLine(-offset);
   }
 
-  // delay the time-consumping reference_line_info init() step to planner.
-  return true;
+  bool has_valid_reference_line = false;
+  for (auto &ref_info : reference_line_info_) {
+    if (!ref_info.Init(obstacles())) {
+      AERROR << "Failed to init reference line";
```
zwccharlie: This log maybe confusing for people see the log and don't familiar with the logic. 
Maybe at last log the number of succeed and failed reference line.

PR #3165, opened 2018-03-03
Perception : Multiple commits to update common and feature_extractor

Capri2014:


Capri2014:
@jzhuucla comment was dismissed, please review again

PR #3197, opened 2018-03-06
Perception : remove namespace obstacle from feature_extractor

Capri2014:


Capri2014:
Dropped since previous commit has already handled this.

PR #3205, opened 2018-03-06
Driver : update ultrasonic radar conf

Capri2014:


Capri2014:
@lemonk3664 please also review. ZERO default chassis, ONE radar, TWO mobileye, THREE ultrasonic

PR #3215, opened 2018-03-06
fix coding style issues for lane_post_processing_subnode

jzhuucla:


modules/perception/obstacle/onboard/lane_post_processing_subnode.cc
```diff
@@ -18,13 +18,13 @@
 
 #include "modules/perception/obstacle/onboard/lane_post_processing_subnode.h"
 
-#include <cfloat>
```
lianglia-apollo: please include cfloat before those third party libraries. https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

PR #3219, opened 2018-03-06
navi: updated transition between navi line and lane markers.

yifeijiang:


modules/map/relative_map/navigation_lane.cc
```diff
@@ -136,7 +136,9 @@ void NavigationLane::ConvertNavigationLineToPath(common::Path *path) {
     return;
   }
   path->set_name("Path from navigation.");
-  UpdateProjectionIndex();
+  if (UpdateProjectionIndex()) {
```
lianglia-apollo: is this !UpdateProjectionIndex() ?

PR #3294, opened 2018-03-08
Dreamview: set map update frequency to 10Hz in navigation mode

vlin17:


modules/dreamview/frontend/src/store/websocket/websocket_realtime.js
```diff
@@ -7,17 +7,17 @@ export default class RosWebSocketEndpoint {
     constructor(serverAddr) {
         this.serverAddr = serverAddr;
         this.websocket = null;
-        this.counter = 0;
-        this.lastUpdateTimestamp = 0;
-        this.currMapRadius = null;
+        this.simWorldUpdatePeriod_ms = 100;
```
siyangy: NIT: camel case variable name

vlin17: done

PR #3340, opened 2018-03-11
planning: fix scenario 33 case.

ycool:


Capri2014:
Test needs to be updated

PR #3350, opened 2018-03-12
Dreamview: update sidebar and layer-menu icons

vlin17:


modules/dreamview/frontend/dist/assets/map_data
```diff
@@ -1 +0,0 @@
-/apollo/modules/map/data
```
siyangy: recover this

PR #3362, opened 2018-03-13
relative_map: adjust relative map length gflags and  monitor: add navigation planning into monitor.

ycool:


modules/map/relative_map/common/relative_map_gflags.cc
```diff
@@ -25,7 +25,7 @@ DEFINE_string(relative_map_config_filename,
 
 DEFINE_int32(relative_map_loop_rate, 10, "Loop rate for relative_map node");
 
-DEFINE_double(max_len_from_navigation_line, 100.0,
+DEFINE_double(max_len_from_navigation_line, 250.0,
               "max navigation path length from navigation line");
 
 DEFINE_double(min_len_for_navigation_lane, 150.0,
```
startcode: we did a little calculation tonight, the min length should be 225m for 65mph with -2 m/^2 deceleration.

PR #3368, opened 2018-03-13
Dreamview: removing hardcoded vehicle values and utm zone id

vlin17:
and fixing few typos


modules/dreamview/frontend/src/store/hmi.js
```diff
@@ -9,6 +9,16 @@ export default class HMI {
 
     vehicles = [];
     @observable currentVehicle = 'none';
+    vehicleParam = {
```
startcode: can we read in this value from "vehicle_config.proto"?

vlin17: Yes, the purpose of the change is to read values from vehicle_config.proto. The values defined here are default, which will be updated during initialization.

PR #3374, opened 2018-03-14
Drivers : Fix channel port

Capri2014:


Capri2014:
@lemonk3664 please also review

PR #3401, opened 2018-03-15
Prediction: use empty predictor for still obstacle in predictor_manager

kechxu:


modules/prediction/predictor/predictor_manager.cc
```diff
@@ -155,6 +155,9 @@ void PredictorManager::Run(const PerceptionObstacles& perception_obstacles) {
     prediction_obstacle.set_timestamp(perception_obstacle.timestamp());
     Obstacle* obstacle = obstacles_container->GetObstacle(id);
     if (obstacle != nullptr) {
+      if (obstacle->IsStill()) {
+        predictor = GetPredictor(ObstacleConf::EMPTY_PREDICTOR);
+      }
       switch (perception_obstacle.type()) {
```
jinghaomiao: need an "else", otherwise the predictor will be overwritten.

PR #3409, opened 2018-03-15
smoother_util: refactor reference line stitching method to further

startcode:
improve stitching smoothness result

### DIFF
--- a/modules/planning/reference_line/smoother_util.cc
+++ b/modules/planning/reference_line/smoother_util.cc
@@ -34,10 +34,7 @@
 
 DEFINE_string(input_file, "", "input file with format x,y per line");
 DEFINE_string(output_file, "", "output file with format x,y per line");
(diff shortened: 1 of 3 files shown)


modules/planning/reference_line/smoother_util.cc
```diff
@@ -171,10 +202,10 @@ class SmootherUtil {
  private:
   std::string filename_;
   std::vector<common::math::Vec2d> raw_points_;
-  std::vector<ReferenceLine> raw_reference_lines_;
-  std::vector<ReferencePoint> reference_points_;
+  std::vector<ReferencePoint> ref_points_;
+  ReferenceLine smoothed_ref_;
   ReferenceLineSmootherConfig config_;
-};
+};  // namespace planning
```
lianglia-apollo: this should be the end of class, not namespace.

PR #3423, opened 2018-03-15
Data: Add latest RoutingRequest to StaticInfo.

xiaoxq:


modules/data/util/info_collector.cc
```diff
@@ -121,6 +123,19 @@ const SoftwareInfo &InfoCollector::GetSoftwareInfo() {
 
   *software->mutable_configs() =
       LoadFiles(instance()->config_.software_configs());
+
+  // Store latest routing request.
+  auto* routing_request_adapter = AdapterManager::GetRoutingRequest();
+  if (routing_request_adapter) {
+    routing_request_adapter->Observe();
+    if (routing_request_adapter->Empty()) {
```
siyangy: should this be !routing_request_adapter->Empty() ?

xiaoxq: Done.

PR #3429, opened 2018-03-15
ci cuda compile in cmake

weidezhang:


perception_cuda_lib/cmake/Options.cmake
```diff
@@ -0,0 +1,40 @@
+# author: zhengwenchao@baidu.com
```
startcode: double check the License files

PR #3432, opened 2018-03-16
Planning: fix lattice feasible region

zwccharlie:


modules/planning/lattice/behavior/feasible_region.cc
```diff
@@ -45,15 +45,17 @@ void FeasibleRegion::Setup(const std::array<double, 3>& init_s,
...
     s_at_speed_limit_ = init_s_[0] + v * t_at_speed_limit_ +
-                        0.5 * FLAGS_longitudinal_acceleration_upper_bound *
+                        0.5 * FLAGS_longitudinal_acceleration_lower_bound *
                             t_at_speed_limit_ * t_at_speed_limit_;
   } else {
-    t_at_speed_limit_ = 0.0;
-    s_at_speed_limit_ = init_s_[0];
+    t_at_speed_limit_ = delta_v / FLAGS_longitudinal_acceleration_upper_bound;;
+    s_at_speed_limit_ = init_s_[0] + v * t_at_speed_limit_ +
+                        0.5 * FLAGS_longitudinal_acceleration_upper_bound *
```
kechxu: I think we should use FLAGS_longitudinal_acceleration_lower_bound here. and there are two ";" on the 55th line

PR #3443, opened 2018-03-16
added script to build perception cuda library

startcode:


scripts/docker_adduser.sh
```diff
@@ -59,8 +59,5 @@ if [ "$RELEASE_DOCKER" != "1" ];then
   # setup ros package
   # this is a temporary solution to avoid ros package downloading.
   ROS="/home/tmp/ros"
-  chmod a+w "${ROS}/share/velodyne/launch/start_velodyne.launch"
```
Capri2014: why delete these?

PR #3469, opened 2018-03-19
Dreamview: display grid ground in navi mode

siyangy:


modules/dreamview/frontend/src/renderer/ground.js
```diff
@@ -12,6 +13,8 @@ export default class Ground {
         this.mesh = null;
         this.geometry = null;
         this.initialized = false;
+        this.inNavigationMode = null;
+
```
vlin17: inNaviMode instead of inNavigationMode??

siyangy: done

PR #3504, opened 2018-03-20
Added camera calibration parameters

techoe:


modules/perception/data/params/front_camera_extrinsics.yaml
```diff
@@ -0,0 +1,21 @@
...
+        nsecs: 0
+    frame_id: white_mkz
+child_frame_id: onsemi_obstacle
+transform:
+    rotation:
+        y: -0.5079853786432498
+        x: 0.516129505674271
+        z: 0.4847175164537079
+        w: -0.4905202170628467
+    rotation_L3:
```
techoe: I will change "rotation_L3" to "rotation_euler" and provide correct value later.

PR #3527, opened 2018-03-22
Perception: Working 3D camera objects in top down view

eugenegx:


Capri2014:
@gengqx different vehicle specific configs like Car7_onsemi_objectDetectionLens_intrinsics6mm.yaml  can be later moved to internal and switch via HMI.

gengqx:
@Capri2014 I think @eugenegx would be a little confused that where gengqx comes from，LOL

PR #3568, opened 2018-03-23
Dreamview: allow flexible ws port for offline mode

vlin17:
this is needed since ws port is not fixed for offlineview


siyangy:
might want to update bundle?

PR #3572, opened 2018-03-25
code janitor trigered by latest commit (1b82f)

startcode:


Capri2014:
close this one since show motion is needed

PR #3590, opened 2018-03-26
Planning: added replan error msg for navigation mode for on vehicle d…

lianglia-apollo:
…ebug.


modules/planning/common/trajectory/trajectory_stitcher.cc
```diff
@@ -63,6 +64,7 @@ std::vector<TrajectoryPoint> TrajectoryStitcher::CalculateInitPoint(
     shift_direction.Normalize();
     ref_point +=
         shift_direction * (distance - FLAGS_replan_lateral_distance_threshold);
+    AERROR << "Replan is triggered. distance = " << distance;
```
ycool: AWARN

PR #3591, opened 2018-03-26
Perception: Change camera obstacle to use new topic. Modify related adapter manager. Change calibration exmaple for front camera

eugenegx:


Capri2014:
@eugenegx Does the topic changed on the driver side as well?

PR #3596, opened 2018-03-27
 [Perception] add PublishPerceptionPb() API to output lane marker proto messages

jzhuucla:


modules/perception/conf/dag_camera_obstacle_lane_vis.config
```diff
@@ -4,14 +4,14 @@ subnode_config {
     subnodes {
         id: 3
         name : "CameraProcessSubnode"
-        reserve : "device_id:camera;publish:1;"
```
eugenegx: "device_id:camera;publish:0;"

modules/perception/obstacle/onboard/lane_post_processing_subnode.cc
```diff
@@ -218,5 +181,69 @@ void LanePostProcessingSubnode::PublishDataAndEvent(
...
+  common::Header *header = obstacles.mutable_header();
+  header->set_lidar_timestamp(0);
+  header->set_camera_timestamp(timestamp_ns_);
+  header->set_radar_timestamp(0);
+
+  // generate lane marker protobuf messages
+  LaneMarkers* lane_markers = obstacles.mutable_lane_marker();
+  LaneObjectsToLaneMarkerProto(*lane_objects, lane_markers);
+
+  common::adapter::AdapterManager::PublishPerceptionObstacles(obstacles);
```
eugenegx: ... where do we want to publish sensor obstacle?  obstacles are published here with lane marking only, and the camera process subnode is turned off

Capri2014:
So we intend to publish lane marking here first to check if the whole pipline works only ? (The following modules would only need lane-makers to be active). @jzhuucla

PR #3602, opened 2018-03-27
HMI: Add velodyne16 to map-collection mode.

xiaoxq:


scripts/velodyne_16.sh
```diff
@@ -0,0 +1,57 @@
...
+source "${DIR}/apollo_base.sh"
+
+# TODO(all): This is just an initial commit. Velodyne 64 and 16 share lots of
+# things. We need to select their processes precisely in 'pgrep', 'pkill' and
+# the monitor module.
+function start() {
+    LOG="${APOLLO_ROOT_DIR}/data/log/velodyne16.out"
+    CMD="roslaunch velodyne start_velodyne_16.launch"
+    NUM_PROCESSES="$(pgrep -c -f "velodyne_nodelet_manager")"
+    if [ "${NUM_PROCESSES}" -eq 0 ]; then
```
Capri2014: Not sure if this works, did 16 and 64 use the same "velodyne_nodelet_manger" as process name?

xiaoxq: That's why I add a "TODO" comment there. Everything should work, except the "pgrep, pkill" and monitor module may get confused a bit.

PR #3623, opened 2018-03-29
planning: added kappa derivative in QuinticSpiralPath

YajiaZhang:


modules/planning/math/curve1d/quintic_spiral_path.cc
```diff
@@ -168,6 +168,29 @@ double QuinticSpiralPath::DeriveTheta(const std::size_t param_index,
...
+  double s4 = s2 * s2;
+
+  double derivative =
+      5.0 * coef_deriv_[5][param_index] * s4 +
+      4.0 * coef_deriv_[4][param_index] * s3 +
+      3.0 * coef_deriv_[3][param_index] * s2 +
+      2.0 * coef_deriv_[2][param_index] * s +
+      coef_deriv_[1][param_index];
+
+  if (param_index == DELTA_S) {
```
Capri2014: Would it be better to change the static const like THETA0 ~ DELTA_S to math common under planning?

PR #3624, opened 2018-03-29
Perception: publish 4 corners of obstacle to PNC

techoe:
It might be ugly but the addfourcorners function was added in object.cc

### DIFF
--- a/modules/perception/obstacle/base/object.cc
+++ b/modules/perception/obstacle/base/object.cc
@@ -85,6 +85,39 @@ std::string Object::ToString() const {
                        b_cipv, "]"));
 }
 
+// Add 4 corners in the polygon
+void Object::AddFourCorners(PerceptionObstacle* pb_obj) const {
+  double cos_theta = cos(theta);
+  double sin_theta = sin(theta);
+
+  double half_width = width / 2;
+  double minus_half_width = -half_width;
+  double half_length = length / 2;
+  double minus_half_length = -half_length;
+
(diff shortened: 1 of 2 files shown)


modules/perception/obstacle/base/object.cc
```diff
@@ -85,7 +85,39 @@ std::string Object::ToString() const {
...
+
+  double half_width = width / 2;
+  double minus_half_width = -half_width;
+  double half_length = length / 2;
+  double minus_half_length = -half_length;
+
+  Point* p1 = pb_obj->add_polygon_point();
+  p1->set_x(minus_half_width * cos_theta + half_length * sin_theta + center(0));
+  p1->set_y(half_length * cos_theta - minus_half_width * sin_theta + center(1));
+  p1->set_z(0.0);
```
Capri2014: setting point z to 0.0 might cause the dreamview visualization problem, is it possible to set a default height for different types of obstacles for now?

Capri2014: @techoe

techoe: Actually there are 8 corners since the object is cube. Among them only 4 corners touching a ground (in that case z=0) are passed to PNC. Do you want another 4 corners on the top where z will be the height of the cube.

Capri2014: Oh my bad, we are actually looking at the height. Thanks.

PR #3665, opened 2018-04-02
ci perf fix and fusion publisher id gflags

weidezhang:


modules/perception/obstacle/onboard/camera_process_subnode.cc
```diff
@@ -114,6 +114,8 @@ void CameraProcessSubnode::ImgCallback(const sensor_msgs::Image &message) {
         timestamp_ns_ = timestamp * 1e9;
   }
 
+  AINFO << "CameraProcessSubnode process frame";
```
startcode: remove this log?

PR #3676, opened 2018-04-02
two commits: added x86_64 dockerfile; force dynamically link qp-oases

startcode:


docker/scripts/image_create.sh
```diff
@@ -0,0 +1,60 @@
...
+    DOCKER_REPO=apolloauto/apollo
+fi
+
+
+APOLLO_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"
+ARCH=$(uname -m)
+TAG="${TYPE}-${ARCH}-${TIME}"
+
+# Build image from APOLLO_ROOT, while use the specified Dockerfile.
+docker build -t "${DOCKER_REPO}:tmp_tag" \
```
Capri2014: tmp_tag?

Capri2014: @startcode

startcode: It is a temporary tag.
Finally it will flat everything to one layer and delete the temporary tag (See line 51 to 60)

PR #3696, opened 2018-04-03
Add new generated protocol files and control func

lbucklandAS:
Prior to this commit we didn't have the correct protocol files due to a lack of differentiation between senders and receivers. This commit adds the correct protocol files and updates the controller to use the correct functions to do control and update chassis.

### DIFF
--- a/modules/canbus/tools/teleop.cc
+++ b/modules/canbus/tools/teleop.cc
@@ -139,7 +139,7 @@ class Teleop {
         perror("read():");
         exit(-1);
       }
-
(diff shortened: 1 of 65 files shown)


lianglia-apollo:
have you run ./apollo.sh check ? The last commit seems to have some lint errors.

lbucklandAS:
@lianglia-apollo I need to do some cleanup, thanks for the heads up.

Capri2014:
@lbucklandAS you can forget about the lint issue as of now since this is side branch as of now. A lot of cleaning up will need to be done after the whole protocol works.

PR #3700, opened 2018-04-04
Dreamview: set default navigation map to BaiduMap

vlin17:


modules/dreamview/frontend/src/store/config/parameters.yml
```diff
@@ -86,11 +86,11 @@ options:
 
 navigation:
   # possible options: BaiduMap or GoogleMap
-  map: "GoogleMap"
+  map: "BaiduMap"
 
   # Google Map API: "https://maps.google.com/maps/api/js"
   # Baidu Map API: "https://api.map.baidu.com/api?v=3.0&ak=0kKZnWWhXEPfzIkklmzAa3dZ&callback=initMap"
-  mapAPiUrl: "https://maps.google.com/maps/api/js"
```
siyangy: can we put this in the code instead of param?
const mapUrls = {
  "GoogleMap": "https://maps.google.com/maps/api/js",
   "BaiduMap": "https://api.map.baidu.com/api?v=3.0&ak=0kKZnWWhXEPfzIkklmzAa3dZ&callback=initMap"
}

ycool: param works better. ak is a key. need be changed by different users.

PR #3715, opened 2018-04-04
Add get objects gpu in yolo detector and comment some log 

owphoo:


xiaoxq:
Merge temporarily for road test.

PR #3730, opened 2018-04-05
[Canbus] Resolving GEM issues. 

Capri2014:
Prior to this commit there were some incorrect scaling values, as well as some assuptions about payload sizes that prevented the gem from fully working. This commit fixes those issues.

### DIFF
--- a/modules/canbus/vehicle/gem/gem_controller.cc
+++ b/modules/canbus/vehicle/gem/gem_controller.cc
@@ -230,6 +230,7 @@ Chassis GemController::chassis() {
 	if (chassis_detail.gem().global_rpt_6a().pacmod_status() ==
 		Global_rpt_6a::PACMOD_STATUS_CONTROL_ENABLED) {
 	  chassis_.set_driving_mode(Chassis::COMPLETE_AUTO_DRIVE);
+   	  global_cmd_69_->set_clear_override(Global_cmd_69::CLEAR_OVERRIDE_DON_T_CLEAR_ACTIVE_OVERRIDES);
 	}
 	else {
 	  chassis_.set_driving_mode(Chassis::COMPLETE_MANUAL);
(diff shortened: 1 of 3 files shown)


Capri2014:
@lbucklandAS This looks good to me, please double verify.

modules/canbus/vehicle/gem/gem_controller.cc
```diff
@@ -422,9 +421,9 @@ void GemController::Steer(double angle, double angle_spd) {
     AINFO << "The current driving mode does not need to set steer.";
     return;
   }
-  const double real_angle = params_.max_steer_angle() * angle / 100.0;
+  const double real_angle = params_.max_steer_angle() * angle / 100.0 * 3.1415/180.0;
 
-  const double real_angle_spd = (
+  const double real_angle_spd = 3.1415 / 180.0 * (
```
startcode: double check this statement.  Is it equivalent to M_PI / 180.0 * params_max_steer_angle_spd() * angle_spd / 100.0?

modules/canbus/vehicle/gem/gem_controller.cc
```diff
@@ -406,11 +404,12 @@ void GemController::Steer(double angle) {
     AINFO << "The current driving mode does not need to set steer.";
     return;
   }
-  const double real_angle = params_.max_steer_angle() * angle / 100.0;
+  // convert to radians
+  const double real_angle = params_.max_steer_angle() * angle / 100.0 * 3.1415 / 180.0;
```
startcode: use M_PI instead of 3.1415

modules/canbus/vehicle/gem/gem_controller.cc
Capri2014: @startcode Will just cherry-pick the commit as is and do the cleaning afterwards, before merging the branch back to master.

PR #3734, opened 2018-04-05
Canbus & others : Add initial support for GEM with further re-factor to be done. 

Capri2014:


modules/canbus/vehicle/gem/protocol/brake_cmd_6b.cc
```diff
@@ -0,0 +1,72 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
startcode: Change to 2018

modules/canbus/vehicle/gem/protocol/brake_motor_rpt_2_71.cc
```diff
@@ -0,0 +1,114 @@
...
+
+  Byte t1(bytes + 1);
+  int32_t t = t1.get_byte(0, 8);
+  x <<= 8;
+  x |= t;
+
+  x <<= 16;
+  x >>= 16;
+
+  int ret = x + -40.000000;
```
startcode: should this 40.0000 be an integer?

modules/canbus/vehicle/gem/protocol/brake_cmd_6b.cc
Capri2014: done

modules/canbus/vehicle/gem/protocol/brake_motor_rpt_2_71.cc
Capri2014: @startcode Done

PR #3743, opened 2018-04-06
[Perception] add CUDA implementation of connected component generator

jzhuucla:


modules/perception/obstacle/camera/lane_post_process/common/connected_component.cc
```diff
@@ -244,9 +244,9 @@ void ConnectedComponent::FindVertices() {
           vertices_->push_back(Vertex(p->x, p->y));
         }
       } else {
-        AERROR << "the point "
-               << "(" << p->x << ", " << p->y << ")"
-               << " is not on bounding box.";
+        std::cerr << "the point "
```
lianglia-apollo: why using cerr here?

modules/perception/cuda_util/block_uf.cu
```diff
@@ -0,0 +1,218 @@
+/******************************************************************************
+ * Copyright (c) 2014, Victor Matheus de Araujo Oliveira All rights reserved.
```
lianglia-apollo: Are we able to use this license?

jzhuucla: The block-wise union find algorithm in CUDA-based implementation of connected component generator is modified from an open-source code which use this BSD license. We consulted with laywer and confirmed that we could use the related code given that "Redistributions in source or binary form must contain the copyright notice for the licensed software and the conditions and disclaimer contained in the license".

PR #3765, opened 2018-04-09
Canbus & Common : start merge common configs

Capri2014:


modules/canbus/vehicle/vehicle_controller.h
```diff
@@ -152,6 +153,7 @@ class VehicleController {
 
  protected:
   canbus::VehicleParameter params_;
```
startcode: delete params_?

modules/canbus/vehicle/gem/gem_controller.cc
```diff
@@ -408,8 +409,7 @@ void GemController::Steer(double angle) {
     return;
   }
   // convert to radians
-  const double real_angle =
-      params_.max_steer_angle() * angle / 100.0 * M_PI / 180.0;
+  const double real_angle = vehicle_params_.max_steer_angle() * angle / 100.0;
```
startcode: the comment "conver to radians" does not match with the code.

modules/canbus/vehicle/vehicle_controller.h
Capri2014: @startcode It is still half-done,  so I left the legacy here for now

modules/canbus/vehicle/gem/gem_controller.cc
Capri2014: @startcode Will changle

Capri2014: @startcode Done

PR #3771, opened 2018-04-09
Planning: bugfix on speed limit decider.

lianglia-apollo:


modules/planning/tasks/st_graph/speed_limit_decider.cc
```diff
@@ -146,10 +146,14 @@ Status SpeedLimitDecider::GetSpeedLimits(
       const auto& nudge = const_path_obstacle->LateralDecision().nudge();
       bool is_close_on_left =
           (nudge.type() == ObjectNudge::LEFT_NUDGE) &&
-          (const_path_obstacle->PerceptionSLBoundary().end_l() > -kRange);
+          (const_path_obstacle->PerceptionSLBoundary().end_l() -
+               adc_sl_boundary_.start_l() >
+           -kRange);
```
startcode: < kRange?

modules/perception/obstacle/camera/lane_post_process/common/type.h
```diff
@@ -24,47 +24,25 @@
...
+#include "modules/perception/proto/perception_obstacle.pb.h"
 
 #ifndef MODULES_PERCEPTION_OBSTACLE_CAMERA_LANE_POST_PROCESS_COMMON_TYPE_H_
 #define MODULES_PERCEPTION_OBSTACLE_CAMERA_LANE_POST_PROCESS_COMMON_TYPE_H_
 
 namespace apollo {
 namespace perception {
 
-#ifndef MAX_LANE_HISTORY
-#define MAX_LANE_HISTORY 10
```
startcode: please convert to gflags or use proto conf. That will be more equivalent to the original code

lianglia-apollo: reversed changes on type.h. We might need to re-name some of the params in gflags as the current naming could be conflict. Might do the change in some future versions.

PR #3779, opened 2018-04-10
Dreamview : add default navi routing end point for 101 east

Capri2014:


Capri2014:
should be south btw, already updated in the file.

PR #3784, opened 2018-04-10
perception: use caffe::SyncedMemory

ghdawn:
use caffe::SyncedMemory directly instead of origin SyncedMemory, which is copied from caffe repo


eugenegx:
@ycool @xiaoxq @weidezhang Please take a look. We may need to update cuda library in released docker image due to updates on CUDA codes

PR #3839, opened 2018-04-12
Localization: Fix order of fstream operation in PoseInterpolation

zhouyao4321:


modules/localization/msf/local_tool/map_creation/poses_interpolation/poses_interpolation.cc
```diff
@@ -80,7 +80,6 @@ void PosesInterpolation::WritePCDPoses() {
   std::ofstream fout;
   fout.open(out_poses_path_.c_str(), std::ofstream::out);
```
lianglia-apollo: Need to check .fail() after openning file.

PR #3857, opened 2018-04-12
Removed incorrect lane filtering

techoe:
Also removed unnecessary AINFO printout.
Use C0 for lateral_distance


modules/perception/obstacle/camera/lane_post_process/cc_lane_post_processor/cc_lane_post_processor.cc
```diff
@@ -327,7 +327,7 @@ bool CCLanePostProcessor::AddInstanceIntoLaneObjectImage(
 
   auto graph = cur_frame_->graph(instance.graph_id);
 
-  ADEBUG << "show points for lane object: ";
+//  ADEBUG << "show points for lane object: ";
```
ycool: If these debug log is not needed any more, let's remove it instead of commenting out.

techoe: Agree! I removed unnecessary AINFO. However still keeping some ADEBUG just in case.

lianglia-apollo: It seems the test_lane_objects_image will be modified after running tests on our local machine.

PR #3865, opened 2018-04-13
planning: improve efficiency for TrajectoryEvaluator.

davidhopper2003:


startcode:
please run ./apollo.sh check and fix the failed lint tests.
Thanks.

modules/planning/lattice/trajectory_generation/trajectory_evaluator.cc
```diff
@@ -82,12 +82,13 @@ TrajectoryEvaluator::TrajectoryEvaluator(
...
         double cost = Evaluate(planning_target, lon_trajectory, lat_trajectory);
-        cost_queue_.push(PairCost({lon_trajectory, lat_trajectory}, cost));
+        // Use more efficient "emplace" member fucntion. 
+        cost_queue_.emplace(PairCost({lon_trajectory, lat_trajectory}, cost));
       } else {
         std::vector<double> cost_components;
         double cost = Evaluate(planning_target, lon_trajectory, lat_trajectory,
                                &cost_components);
-        cost_queue_with_components_.push(PairCostWithComponents(
+        cost_queue_with_components_.emplace(PairCostWithComponents(
```
lianglia-apollo: this is not necessary. push or emplace here have no difference.
It is generally preferred to keeping using push if there is no performance improvement.

modules/planning/lattice/trajectory_generation/trajectory_evaluator.cc
```diff
@@ -173,7 +174,7 @@ double TrajectoryEvaluator::Evaluate(
   std::vector<double> s_values;
   for (double s = 0.0; s < evaluation_horizon;
        s += FLAGS_trajectory_space_resolution) {
-    s_values.push_back(s);
+    s_values.emplace_back(s);
```
lianglia-apollo: this is not necessary. push or emplace here have no difference.

davidhopper2003: For native data types, the “emplace_back” or "emplace" function doesn't improve efficiency over the "push_back" or "push" fucntion; but for non-native data types, the formmer doesn't invoke an extra constructor and thus the efficiency is enhanced. Whether for native or non-native data types, the C++11 standard recommends using the "emplace_back" function. Therefore, I think my change is no problem.

davidhopper2003:
@startcode Yes, I ignored the code style checking test. I will finish this job at once.

modules/planning/lattice/trajectory_generation/trajectory_evaluator.cc
lianglia-apollo: Please use cost_queue_with_components_.emplace(lon_trajectory, lat_trajectory).

The way you wrote (emplace(PairCostWithComponents{xxx, xxx}) requires a construction of temporary instance and invokes the copy constructor, which is identical to what you will get with a push().

davidhopper2003: Great, thank you for pointing out my problems. I will correct them right away.

davidhopper2003: I have modified the code, please check again. You are awesome!

PR #3866, opened 2018-04-13
add velodyne16 & velodyne 64 for map collection script

msbeta:


scripts/record_map_data.sh
```diff
@@ -36,6 +36,8 @@ function start() {
         /apollo/sensor/gnss/imu \
         /apollo/sensor/gnss/odometry \
         /apollo/sensor/gnss/raw_data \
+        /apollo/sensor/velodyne64/VelodyneScanUnified
```
ycool: /apollo/sensor/velodyne64/VelodyneScanUnified  \

PR #3906, opened 2018-04-17
Planning: use linear interpolation as default method.

lianglia-apollo:


modules/planning/common/planning_util.cc
```diff
@@ -37,126 +37,6 @@ using common::SpeedPoint;
 using common::TrajectoryPoint;
 using common::adapter::AdapterManager;
 
-PathPoint Interpolate(const PathPoint &p0, const PathPoint &p1,
```
YajiaZhang: Keep this function.

modules/planning/common/planning_util.cc
```diff
@@ -37,126 +37,6 @@ using common::SpeedPoint;
...
-  p.set_theta(theta);
-  p.set_kappa(kappa);
-  p.set_dkappa(dkappa);
-  p.set_ddkappa(d2kappa);
-  p.set_s(s);
-  return p;
-}
-
-TrajectoryPoint Interpolate(const TrajectoryPoint &tp0,
-                            const TrajectoryPoint &tp1, const double t) {
```
YajiaZhang: This function as well.

modules/planning/common/planning_util.cc
lianglia-apollo: It is better to put them as some common functions, with specific names under common/math when some code really need them. Currently they are not used in any code yet.

modules/planning/common/planning_util.cc
lianglia-apollo: Please find the comments above.

modules/common/math/linear_interpolation.cc
```diff
@@ -91,22 +98,18 @@ TrajectoryPoint InterpolateUsingLinearApproximation(const TrajectoryPoint &tp0,
...
-  path_point->set_kappa(
-      common::math::lerp(pp0.kappa(), t0, pp1.kappa(), t1, t));
-  path_point->set_dkappa(
-      common::math::lerp(pp0.dkappa(), t0, pp1.dkappa(), t1, t));
-  path_point->set_ddkappa(
-      common::math::lerp(pp0.ddkappa(), t0, pp1.ddkappa(), t1, t));
-  path_point->set_s(common::math::lerp(pp0.s(), t0, pp1.s(), t1, t));
+  path_point->set_x(lerp(pp0.x(), t0, pp1.x(), t1, t));
+  path_point->set_y(lerp(pp0.y(), t0, pp1.y(), t1, t));
+  path_point->set_theta(lerp(pp0.theta(), t0, pp1.theta(), t1, t));
```
kechxu: Shall we use slerp to set theta? because theta is an angle.

modules/planning/common/planning_util.cc
lianglia-apollo: moved to common/math

modules/common/math/linear_interpolation.cc
lianglia-apollo: Yes, the original code seems to have a bug here.

PR #3923, opened 2018-04-17
refactor sim_control interface

startcode:


modules/dreamview/backend/sim_control/sim_control.cc
```diff
@@ -211,86 +205,79 @@ void SimControl::Freeze() {
   prev_point_ = next_point_;
 }
 
-double SimControl::AbsoluteTimeOfNextPoint() {
-  return current_trajectory_.header().timestamp_sec() +
-         current_trajectory_.trajectory_point(next_point_index_)
-             .relative_time();
+bool SimControl::IndexWithinRange(int index) {
```
siyangy: is this function still being called?

startcode: not used anymore. deleted.

PR #3952, opened 2018-04-18
planning: efficiency improvement for backup trajectory generator

YajiaZhang:


modules/planning/lattice/trajectory_generation/backup_trajectory_generator.cc
```diff
@@ -22,9 +22,10 @@ namespace apollo {
 namespace planning {
 
 using apollo::common::PathPoint;
+using State = std::array<double, 3>;
```
startcode: recommend using a less generic name

YajiaZhang: I think it is fine since it will be only used in this source file.

PR #3970, opened 2018-04-19
planning: move stop for static cross lane vehicle logic to traffic rule

startcode:


lianglia-apollo:
Why tests are changed? It looks like a code refactor and should generate same results.

startcode:
We enforced stop for particular obstacles, which was ignored before if path has not reach there.

PR #4025, opened 2018-04-25
Control : fix matrix bilinear transformation in MPC

Capri2014:


Capri2014:
@lianglia-apollo Totally agree, could you help add some?

Capri2014:
“

PR #4059, opened 2018-04-30
scripts: fix the usage of apollo_base.sh script

zhxt:


zhxt:
$0 would be 'bash' since this script is commonly used with `source` or `.` 
And also, it seems like it is not used as `apollo_base.sh COMMAND`

PR #4110, opened 2018-05-08
Relative_map: fixed a bug in setting "last_project_index_" for a cyclic/circular navigation line.

davidhopper2003:
Because the starting point and the end point cannot be completely consistent in a cyclic/circular navigaton line. The vehicle's end point is usually beyond the starting point a little when making a cyclic/circular navigation line. Therefore, the "last_project_index_" is reset to 0 if it is greater than 95% size of the navigaton line and the vehicle's current position is near the starting point of the navigatoin line.


davidhopper2003:
@lianglia-apollo Your consideration is reasonable. I have added "FLAGS_enable_cyclic_rerouting" to enable/disable auto rerouting in a in a cyclic/circular navigaton line. Is this okay?

PR #4174, opened 2018-05-15
docker: add a script to stop apollo containers

zhxt:


zhxt:
@Capri2014 @xiaoxq  Yeah, agree. 
One thing I concerned is that calling "dev_start.sh --stop" looks a little little ambiguous.

xiaoxq:
Haha, that also makes sense. Let's just take it as a temporary effort. I'm doing the migration to manage containers wisely: https://github.com/ApolloAuto/apollo/tree/master/docker/canary
I will move it to canary/stop_container.sh when things are ready.

docker/scripts/dev_start.sh
```diff
@@ -118,7 +137,11 @@ do
         shift
         source ${APOLLO_ROOT_DIR}/docker/scripts/restart_map_volume.sh \
             "${map_name}" "${VOLUME_VERSION}"
-    ;;
+        ;;
+    --stop)
```
xiaoxq: s/--stop/stop/ would be cleaner and aligned with our other scripts.

zhxt: Okay, fixed.

zhxt:
@xiaoxq Yeah, I have noticed that a new script/tool is under developing.

And I was wondering if it would be better to use a common name for the script instead of being with the specific operation("start/stop...").

If so, then we can take 'start/stop' as sub commands, and developers can do all the setups  with just  remembering only one script .

Then the commands may look like as follow:
(I will take "canary.py" as the script name here for demostration.)
```
canary.py start [options]   # prepare and start containers    = dev_start.sh
canary.py enter             # enter docker environment        = dev_into.sh
canary.py stop              # stop all containers             = dev_start.sh stop
canary.py clean    # clean the resources that no longer used  = cleanup_resources.sh

```
As a side effect, the logic of  the script may become a little more complicated.
Inspiration comes from [VBoxManage](https://www.virtualbox.org/manual/ch08.html#idm3728)
 
(hmm, seems no much improve :-( 
Just FYI.

PR #4180, opened 2018-05-15
planning: refactor dp_road_graph

startcode:
* reduce unnecessary code


startcode:
test changed because of removing resonate points

PR #4200, opened 2018-05-18
Improved the writing with the help of Grammarly.

mzweilin:
I strongly recommend the ApolloAuto team to make use of Grammarly in writing documents. BTW, I still don't understand the sentence 'Baidu is using HM solver from Google.'


Capri2014:
Thanks!

PR #4248, opened 2018-05-24
Canbus : re-factor engage advice and add safety check before allow en…

Capri2014:
…gage


modules/canbus/vehicle/lincoln/lincoln_controller.cc
```diff
@@ -342,16 +342,19 @@ Chassis LincolnController::chassis() {
...
-        apollo::common::EngageAdvice::READY_TO_ENGAGE);
-  } else {
+        apollo::common::EngageAdvice::DISALLOW_ENGAGE);
+    chassis_.mutable_engage_advice()->set_reason("Chassis error!");
+  } else if (chassis_.parking_brake() || !CheckSafetyError(chassis_detail)) {
     chassis_.mutable_engage_advice()->set_advice(
         apollo::common::EngageAdvice::DISALLOW_ENGAGE);
     chassis_.mutable_engage_advice()->set_reason(
-        "CANBUS not ready, firmware error or emergency button pressed!");
+        "Vehicle is not in a safe state to engage!");
```
startcode: the reason needs to be more specific. Otherwise, the driver or passenger would not know why not allowed to engage.

PR #4263, opened 2018-05-24
Common: refactored kalman filter 1d by inheritance from generic kalma…

jinghaomiao:
…n filter

### DIFF
--- a/modules/common/math/BUILD
+++ b/modules/common/math/BUILD
@@ -12,6 +12,7 @@ cc_library(
         ":geometry",
         ":integral",
         ":kalman_filter",
+        ":kalman_filter_1d",
         ":linear_interpolation",
         ":lqr",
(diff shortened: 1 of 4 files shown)


Capri2014:
@jinghaomiao Shall we make it more generic by : 1. customizable matrix size for A, B, C, Q, R, 2. Load these configurations from conf. 3.  Make it a template so it can accept different data type? Third one is probably an over-kill since most of the places I can imagine that we use Kalman Filter is taking double or float.

jinghaomiao:
@Capri2014 I agree that the internal parameters should be configurable. I'd like to keep the original interfaces intact at very beginning so that other code won't be affected. I will then look at how this kalman_filter_1d is used in various places within the code so that I can have a better understanding of how to refactor that in the next pull request. What do you think?

PR #4265, opened 2018-05-25
 gnss driver: refactor & add wheelvelocity command support

fengqikai1414:


Capri2014:
@fengqikai1414 Could you do a git rebase from upstream before push? Otherwise this would be less easier to be merged back to master.

PR #4300, opened 2018-05-30
Fix a typo of the location of install_docker.sh

c-xyli:
Replace "bash docker/scripts/install_docker.sh" with  "bash docker/setup_host/install_docker.sh"


ycool:
Can you help to click CLAassistant to sign CLA? thanks.

PR #4309, opened 2018-05-31
planning: (1)pull over more (2) tests

jmtao:


modules/planning/conf/traffic_rule_config.pb.txt
```diff
@@ -63,6 +63,16 @@ config: {
   }
 }
 
+config: {
+  rule_id: PULL_OVER
+  enabled: true
+  pull_over {
+    max_stop_deceleration: 2.0
+    stop_distance: 0.5
```
jilinzhou: Does it mean the ego vehicle stops at 0.5 meters before the destination ENU point?

modules/planning/tasks/traffic_decider/pull_over.cc
```diff
@@ -44,20 +44,20 @@ Status PullOver::ApplyRule(Frame* const frame,
...
 
   PointENU stop_point;
   double stop_heading = 0.0;
-  if (!GetPullOverStop(&stop_point, &stop_heading)) {
-    ADEBUG << "Could not find a safe pull over point";
-    return Status::OK();
-  }
-  if (!BuildPullOverStop(stop_point, stop_heading)) {
-    AWARN << "Not able to create a pull over stop";
+  if (GetPullOverStop(&stop_point, &stop_heading) < 0) {
```
jilinzhou: if (GetPullOverStop(&stop_point, &stop_heading) != 0)

modules/planning/tasks/traffic_decider/pull_over.cc
```diff
@@ -73,37 +73,61 @@ bool PullOver::IsValidStop(const PointENU& stop_point,
...
+
+  const auto& reference_line = reference_line_info_->reference_line();
+  double heading = reference_line.GetReferencePoint(stop_s).heading();
+  double lane_left_width = 0.0;
+  double lane_right_width = 0.0;
+  reference_line.GetLaneWidth(stop_s, &lane_left_width, &lane_right_width);
+  stop_point_sl.set_l(-lane_right_width);
+
+  common::math::Vec2d point;
+  reference_line.SLToXY(stop_point_sl, &point);
```
jilinzhou: What happen if stop_point_sl.s() > reference_line.length()?

modules/planning/tasks/traffic_decider/pull_over.cc
```diff
@@ -148,31 +172,31 @@ bool PullOver::CheckPullOver() {
...
 }
 
-bool PullOver::BuildPullOverStop(const PointENU& stop_point,
+int PullOver::BuildPullOverStop(const PointENU& stop_point,
                                  double stop_heading) {
   // check
   const auto& reference_line = reference_line_info_->reference_line();
   common::SLPoint sl;
   reference_line.XYToSL(stop_point, &sl);
   if (sl.s() < 0 || sl.s() > reference_line.Length()) {
```
jilinzhou: Is it necessary to double check sl.s() here?

modules/planning/conf/traffic_rule_config.pb.txt
jmtao: yes.  it's the distance adc keeps from virtual stop wall when adc stops.

modules/planning/tasks/traffic_decider/pull_over.cc
jmtao: stop_point is adc's position + plan_distance(config, currently it is set to 20m). It shall be within reference line.

modules/planning/tasks/traffic_decider/pull_over.cc
jmtao: it is always recommended to add checks. you can't assume the parameters passed in are all legit.

modules/planning/tasks/traffic_decider/pull_over.cc
jmtao: either way

PR #4365, opened 2018-06-04
Map: added the path priority parameter for a navigation path and fixed a bug in NavigationLane::MergeNavigationLineAndLaneMarker().

davidhopper2003:
screenshot from 2018-06-04 14-58-56

### DIFF
--- a/modules/map/relative_map/navigation_lane.cc
+++ b/modules/map/relative_map/navigation_lane.cc
@@ -61,6 +61,7 @@ bool NavigationLane::GeneratePath() {
     current_navi_path_ = std::make_shared<NavigationPath>();
     auto *path = current_navi_path_->mutable_path();
     ConvertLaneMarkerToPath(lane_marker, path);
+    current_navi_path_->set_path_priority(0);
   };
 
   // priority: merge > navigation line > perception lane marker
@@ -73,6 +74,8 @@ bool NavigationLane::GeneratePath() {
       auto current_navi_path = std::make_shared<NavigationPath>();
       auto *path = current_navi_path->mutable_path();
       if (ConvertNavigationLineToPath(path, i)) {
+        current_navi_path->set_path_priority(
+            navigation_info_.navigation_path(i).path_priority());
         navigation_path_list_.emplace_back(i, current_navi_path);
       }
     }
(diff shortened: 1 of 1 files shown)


davidhopper2003:
The method for generating the common boundary of left and right lanes needs to be improved.

modules/map/relative_map/navigation_lane.cc
```diff
@@ -162,20 +166,24 @@ void NavigationLane::MergeNavigationLineAndLaneMarker(common::Path *path,
...
   int lane_marker_index = 0;
-  for (double s = 0.0; s < len; s += ds) {
+  common::Path temp_path;
+  for (double s = start_s; s < len; s += ds) {
     auto p1 = GetPathPointByS(navigation_path, navigation_index, s,
                               &navigation_index);
     auto p2 = GetPathPointByS(lane_marker_path, lane_marker_index, s,
                               &lane_marker_index);
-    auto *p = path->add_path_point();
+    auto *p = temp_path.add_path_point();
```
lianglia-apollo: can we do the modification directly on path?

davidhopper2003: It may not be appropriate. Because directly modifying the points of the original navigation path will cause the starting and ending points of the original path may not be modified, and ultimately the resulting navigation path will not be smooth.  Also, the length of a navigation path does not exceed 250 meters, so using temporary variables does not reduce much efficiency.
Here is the implementation code based on your idea. The result is as follows. It can be seen that the lane boundary is not particularly smooth.

``` cpp
  const double ds = 1.0;
  int navigation_index = 0;
  int lane_marker_index = 0; 
  for (double s = start_s; s < len; s += ds) {
    auto p1 = GetPathPointByS(navigation_path, navigation_index, s,
                              &navigation_index);
    auto p2 = GetPathPointByS(lane_marker_path, lane_marker_index, s,
                              &lane_marker_index);
    // auto *p = temp_path.add_path_point();
    common::PathPoint *p = nullptr;
    if (path->path_point_size() > 0) {
      p = path->mutable_path_point(navigation_index);
    } else {
      p = path->add_path_point();
    }
    const double kWeight = 0.9;
    *p = common::util::GetWeightedAverageOfTwoPathPoints(p1, p2, kWeight,
                                                         1 - kWeight);
  }  
```
screenshot from 2018-06-04 15-58-58

Yuyn1234:
Why do you have four reference lines? I only have two.

PR #4386, opened 2018-06-05
Perception: add ultrasonic subnode framework

kechxu:


modules/perception/proto/perception_ultrasonic.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto2";
+
+package apollo.perception;
+
+import "modules/common/proto/header.proto";
+import "modules/common/proto/geometry.proto";
+
+message ImpendingCollisionEdge {
+    optional int32 id = 1;  // edge ID.
+    repeated int32 cone_id_list = 2;  // cone ID.
```
Capri2014: cone ID?

Capri2014:
@kechxu How would you plan to add sonars calibration parameters? load from configs in perception? or you want can bus to publish both range/calibration datas? @kechxu @techoe @gchen-apollo @gengqx @weidezhang

modules/perception/proto/perception_ultrasonic.proto
kechxu: Yes

techoe:
Awesome! Thank you so much!

PR #4415, opened 2018-06-06
Scripts&&Guardian : Fix mod and lock bug

Capri2014:


modules/guardian/guardian.cc
```diff
@@ -60,14 +60,22 @@ void Guardian::Stop() { timer_.stop(); }
...
-  if (system_status_.has_safety_mode_trigger_time() && FLAGS_guardian_enabled) {
+  bool safety_mode_triggered = false;
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (system_status_.has_safety_mode_trigger_time()) {
+      safety_mode_triggered = true;
+    }
+  }
+
+  if (safety_mode_triggered && FLAGS_guardian_enabled) {
```
startcode: move FLAGS_guardian_enabled to the front of the function to do quick rejection, or is this gflag really needed?

In addition, if all the ROS calls use the same mutex_, it is equivalent to using single thread receiving mode in ROS

PR #4447, opened 2018-06-07
Perception: Fixed perception crash on fusion_subnode of Apollo 2.0; 

techoe:
Also Removed unnecessary logging;

When motion service was not available, it returns false and crash happened. I fixed it in this PR.


modules/perception/obstacle/onboard/fusion_subnode.cc
```diff
@@ -91,18 +91,9 @@ bool FusionSubnode::InitInternal() {
     motion_service_ = dynamic_cast<MotionService*>(
         DAGStreaming::GetSubnodeByName("MotionService"));
     if (motion_service_ == nullptr) {
-      AERROR << "motion service not inited";
-      return false;
```
techoe: This "return false" caused crash. We don't have to return false even motion service is not on (especially for Apollo 2.0)

PR #4463, opened 2018-06-08
fix bugs for asynchronous track to track fusion

weidezhang:
1. tuned parameter for process noise
2. added better visualization for fusion result
3. added more debug info for visualization
4. fix bugs for imf process noise
5. added singularity check w/ correction for covariance matrix


modules/perception/obstacle/camera/filter/object_camera_filter.cc
```diff
@@ -57,6 +57,16 @@ void ObjectCameraFilter::Create(const int track_id, const double timestamp,
   tracked_filters_[track_id].x_.Init(obj_ptr->center.x());
   tracked_filters_[track_id].y_.Init(obj_ptr->center.y());
   tracked_filters_[track_id].theta_.Init(obj_ptr->theta);
+  auto x_state_cov = tracked_filters_[track_id].x_.GetCov();
+  auto y_state_cov = tracked_filters_[track_id].y_.GetCov();
+  obj_ptr->state_uncertainty.block(0, 0, 2, 2) << x_state_cov(0, 0), 0, 0,
```
Capri2014: @weidezhang Can you elaborate more how did you set it this way?

modules/common/math/kalman_filter_1d.cc
```diff
@@ -40,7 +40,7 @@ bool KalmanFilter1D::Init(const float& x) {
 
   Eigen::Matrix<float, 2, 2> q;
   q.setIdentity();
-  q *= 2.0f;
+  q *= 10.0f;
```
Capri2014: change it to load config instead of directly hard code the coefficient here.

weidezhang: yes, will fix later in a separate CR

PR #4469, opened 2018-06-11
Fix multi navigation issue

mfyanga:
1.fix issue:When the target lane is on the left of the car, the method called to find the nearest adjacent lane to the car in the target direction is wrong.

2.Optimize the code:
  1)format log output.
  2)when need change lane, only provide the navigation line where the vehicle is currently located and 
     the navigation line where the vehicle is about to change lane.
  3)when can not need change lane,only provide the navigation line where the vehicle is currently 
     located.
  4)delete something about vehicle_state_provider.

### DIFF
--- a/modules/planning/reference_line/reference_line_provider.cc
+++ b/modules/planning/reference_line/reference_line_provider.cc
@@ -30,7 +30,6 @@
 #include "modules/common/configs/vehicle_config_helper.h"
 #include "modules/common/time/time.h"
 #include "modules/common/util/file.h"
-#include "modules/common/vehicle_state/vehicle_state_provider.h"
 #include "modules/map/hdmap/hdmap_util.h"
 #include "modules/map/pnc_map/path.h"
 #include "modules/planning/common/planning_gflags.h"
@@ -297,6 +296,7 @@ bool ReferenceLineProvider::GetReferenceLinesFromRelativeMap(
     std::list<ReferenceLine> *reference_line,
     std::list<hdmap::RouteSegments> *segments) {
   if (relative_map.navigation_path_size() <= 0) {
+    AERROR << "relative_map.navigation_path_size() <= 0";
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


modules/planning/reference_line/reference_line_provider.cc
```diff
@@ -474,6 +477,14 @@ bool ReferenceLineProvider::GetReferenceLinesFromRelativeMap(
         segment.SetPreviousAction(lane_change_type);
       } else if (lane_id == adc_lane_id) {
         segment.SetNextAction(lane_change_type);
+      } else {
+        continue;
+      }
+    } else {
+      // can not need change lane,Only provide the navigation line where the
+      // vehicle is currently located
+      if (lane_id != adc_lane_id) {
```
startcode: This code has no effect. please remove.

modules/planning/reference_line/reference_line_provider.cc
```diff
@@ -317,8 +317,9 @@ bool ReferenceLineProvider::GetReferenceLinesFromRelativeMap(
     return false;
   }
   // get curent adc lane info by vehicle state
-  common::VehicleState vehicle_state =
-      common::VehicleStateProvider::instance()->vehicle_state();
+  // since the relative map is based on the coordinate system of the car itself,
+  // the vehicle is the origin of the coordinate system
+  common::VehicleState vehicle_state;
```
startcode: always get vehicle_state from vehicle state provider, to centralize the vehicle state generation logic.

mfyanga:
@lianglia-apollo  
1. It has been modified according to the review comments.
2.Optimize the code
   1)format log output.
   2)when need change lane, only provide the navigation line where the vehicle is currently located and
      the navigation line where the vehicle is about to change lane.
   3)it is better to use a while loop than a for loop to get the left and right neighbor lanes of the current 
      lane
  4)when changing lanes, use the first adjacent lane on the left or right of the current lane as the target 
     lane of this lane change

PR #4471, opened 2018-06-11
Driver: added options for sensor sync

lianglia-apollo:


modules/drivers/velodyne/velodyne_driver/src/driver/driver.h
```diff
@@ -40,6 +39,7 @@ struct Config {
   int firing_data_port = 0;
   int positioning_data_port = 0;
   int prefix_angle = 0;  // prefix angle to recv
+  bool use_sensor_sync = false;
```
startcode: how to set this variable?

PR #4477, opened 2018-06-11
ReadMe with improvements

natashadsouza:
Image 2.5 is updated with color and Blog is updated with the correct URL


ladixiu:
呼呼哈哈

PR #4481, opened 2018-06-11
Ultrasonic dev

kechxu:
Please merge via rebase


kechxu:
@Capri2014 pub_meta_events_ is in the base class Subnode. device_id_ will be initialized in the next CI.

Issue #4502, opened 2018-06-13
Pandar40P_internal: code consolidation


freeHackOfJeff:
1. isolate variable scope
2. use bit shift inside loop for performance benefits

mickeyouyou:
we have received you pull request. thank you so much.

Zhenni17:
@freeHackOfJeff can you please clarify your question about this issue?

natashadsouza:
Closing this issue as we have received a pull request for the same. Thanks!

PR #4503, opened 2018-06-13
Pandar40P_internal: code consolidation #4502 

freeHackOfJeff:
Isolate variable scope of return value and use bit shift inside loop for performance benefits.
Also, it's better to comment out code with if...endif statement.


ycool:
close for no file changed.

PR #4508, opened 2018-06-13
Planning: added framework code for the new NaviPlanner.

davidhopper2003:


modules/planning/proto/planning_config.proto
```diff
@@ -18,6 +18,12 @@ enum TaskType {
   POLY_ST_SPEED_OPTIMIZER = 6;
 };
 
+enum NaviTaskType {
```
ycool: move these task types into TaskType

davidhopper2003: Thank you. 
I thas been modified according to your suggestion.

PR #4529, opened 2018-06-14
emplace->emplace_back , std::move(unnecessary)

xxxx001:
re  pr  2018.06.14


xxxx001:
have a not diffence ,  Performance not .

xxxx001:
have a very difference here , The performance is completely different

PR #4533, opened 2018-06-14
planning:modify class LocalPath for NaviPlanner

lighton2012:


davidhopper2003:
@yifeijiang @lianglia-apollo 
Please review the code added by Luo Xuguang. 
Thank you.

modules/planning/navi/common/local_path.cc
```diff
@@ -0,0 +1,63 @@
...
+ */
+#include <vector>
+
+#include "modules/planning/navi/common/local_path.h"
+
+namespace apollo {
+namespace planning {
+
+bool LocalPath::GetInitY(double &y) {
+  if (points_.size() < 1) {
```
lianglia-apollo: points_.empty()?

modules/planning/navi/common/local_path.cc
```diff
@@ -0,0 +1,63 @@
...
+  }
+  y = points_[0].y();
+  return true;
+}
+
+std::vector<Vec2d> LocalPath::GetXYVec() { return points_; }
+
+int LocalPath::GetRange() { return static_cast<int>(path_points_.size() - 1); }
+
+void LocalPath::Shift(double dist) {
```
lianglia-apollo: please use const double dist (use const whenever possible)

modules/planning/navi/common/local_path.cc
```diff
@@ -0,0 +1,63 @@
...
+  for (int i = 0; i < path_points_.size(); ++i) {
+    path_points_[i].set_y(path_points_[i].y() + dist);
+  }
+}
+
+void LocalPath::Cut(double dist) { return; }
```
lianglia-apollo: function empty ?

modules/planning/navi/common/local_path.cc
```diff
@@ -0,0 +1,63 @@
...
+
+void LocalPath::Resample() { return; }
+
+void LocalPath::Merge(common::Path local_path, double weight) {
```
lianglia-apollo: use const reference for local_path, 
const double weight.

modules/planning/navi/common/local_path.cc
```diff
@@ -0,0 +1,63 @@
...
+  for (int i = 0; i < path_points_.size(); ++i) {
```
lianglia-apollo: size_t i = 0 (int will cause a compiling warning)

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,109 @@
...
+#include "modules/common/proto/pnc_point.pb.h"
+
+/**
+ * @namespace apollo::planning
+ * @brief apollo::planning
+ */
+namespace apollo {
+namespace planning {
+
+using apollo::common::math::Vec2d;
```
lianglia-apollo: only use "using ... " in .cc file to avoid contamination on other files.

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * @class LocalPath
+ * @brief LocalPath is used to privide some path point operations for decider
+ */
+class LocalPath {
+ public:
+  LocalPath(std::vector<common::PathPoint> path_points)
```
lianglia-apollo: please use reference for path_points.

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,109 @@
...
+      : path_points_(path_points) {
+    points_.clear();
+    for (int i = 0; i < path_points_.size(); ++i) {
```
lianglia-apollo: size_t i = 0;

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  virtual ~LocalPath() = default;
+
+  /**
+   * @brief get the first y of points vector.
+   * @param y input parm and the output.
+   * @return bool whether the y is valid or not.
+   */
+  bool GetInitY(double &y);
```
lianglia-apollo: use a pointer for output, unless there is a strong reason not to do so (such as swap()).

modules/planning/navi/common/local_path.cc
lighton2012: referring to local_path.py,delete this function?

lighton2012:
have modified class LocalPath for NaviPlanner according to the opinions

modules/planning/navi/common/local_path.cc
```diff
@@ -0,0 +1,73 @@
...
+  // TODO(all):not realized now and will be perfected in the future
+  return;
+}
+
+void LocalPath::Resample() {
+  // TODO(all):not realized now and will be perfected in the future
+  return;
+}
+
+void LocalPath::Merge(const common::Path local_path, const double weight) {
```
lianglia-apollo: we should use const common::Path& local_path to avoid duplicating the whole path.

modules/planning/navi/common/local_path.cc
```diff
@@ -0,0 +1,73 @@
...
+  for (size_t i = 0; i < path_points_.size(); ++i) {
+    auto y = path_points_[i].y();
+    if (i < local_path.path_point_size()) {
```
lianglia-apollo: static_cast<int>(i) < local_path.path_point_size().
In protobuf, the size is defined as an int type. We should avoid comparing between different types.

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,107 @@
...
+ */
+
+#ifndef MODULES_PLANNING_NAVI_LOCAL_PATH_H_
+#define MODULES_PLANNING_NAVI_LOCAL_PATH_H_
+
+#include <memory>
+#include <vector>
+
+#include "modules/common/math/vec2d.h"
+#include "modules/common/proto/pnc_point.pb.h"
```
lianglia-apollo: it is recommended to put system generated headers before user written headers, so a better way is:
```
#include <memory>
#include <vector>

#include "modules/common/proto/pnc_point.pb.h"

#include "modules/common/math/vec2d.h"
```

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,107 @@
...
+ */
+namespace apollo {
+namespace planning {
+/**
+ * @class LocalPath
+ * @brief LocalPath is used to privide some path point operations for decider
+ */
+class LocalPath {
+ public:
+  LocalPath(std::vector<common::PathPoint> &path_points)
```
lianglia-apollo: Please use const when you don't modify the params:
const std::vector<common::PathPoint> &path_points

In general, we never use non-const reference for function parameters, unless you have a strong reason to do it.

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,107 @@
...
+      : path_points_(path_points) {
+    points_.clear();
+    for (int i = 0; i < path_points_.size(); ++i) {
```
lianglia-apollo: size_t i = 0;
avoid comparing between different types.

modules/planning/navi/common/local_path.h
```diff
@@ -0,0 +1,107 @@
...
+   * @brief merge the points with local path.
+   * @param common::Path local_path the path which is used to merge with points.
+   * @param weight the weight coefficient.
+   * @return void.
+   */
+  void Merge(const common::Path local_path, const double weight);
+
+ private:
+  std::vector<common::PathPoint> path_points_;
+  std::vector<Vec2d> points_;
```
lianglia-apollo: Please double check other files for putting "using" in header files. If ```std::vector<Vec2d>``` is compilable, it indicates an contamination.

lighton2012:
modify class LocalPath for NaviPlanner and added BUILD file

PR #4549, opened 2018-06-15
 Usb Cam Sensor FPGA driver buffer count is set to four and add set-trigger function.

cicelprg:
Usb Cam Sensor FPGA driver buffer count is set to four and add set-trigger function.


modules/drivers/usb_cam/nodes/usb_cam_wrapper.cpp
```diff
@@ -168,6 +170,15 @@ UsbCamWrapper::UsbCamWrapper(ros::NodeHandle node, ros::NodeHandle private_nh) :
     if (focus_ >= 0)
       cam_.set_v4l_parameter("focus_absolute", focus_);
   }
+
+  //trigger enable
+  int trigger_ret = cam_.trigger_enable(trigger_fps_, trigger_internal_);
+  if (0 != trigger_ret)
```
startcode: no newline before { here.

PR #4550, opened 2018-06-15
Map: added a new item in MapMsg for relative map.

davidhopper2003:
The "vechicle_lane_id" is the lane id of the navigation path which the vehicle is currently on. It is used to indicate the current lane when changing lanes.

### DIFF
--- a/modules/map/relative_map/proto/navigation.proto
+++ b/modules/map/relative_map/proto/navigation.proto
@@ -10,7 +10,7 @@ import "modules/perception/proto/perception_obstacle.proto";
 
 message NavigationPath {
   optional apollo.common.Path path = 1;
- // highest = 0 which can directly reach destination; change lane indicator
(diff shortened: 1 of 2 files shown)


modules/map/relative_map/proto/navigation.proto
```diff
@@ -38,4 +38,7 @@ message MapMsg {
 
   // localization
   optional apollo.localization.LocalizationEstimate localization = 5;
+
+  // the lane id of the navigation path which the vehicle is currently on
+  optional string vechicle_lane_id = 6;
```
yifeijiang: typo: vechicle > vehicle 

it is better to decouple the vehicle state with map information.

davidhopper2003: Yes, your consideration makes sense, and I withdraw this submission.

PR #4590, opened 2018-06-19
Canbus : add test for gem protocol

Capri2014:


Capri2014:
@JasonXu12 please also review

PR #4592, opened 2018-06-19
perception: added EKF based object camera filter

YajiaZhang:


modules/perception/obstacle/camera/filter/object_camera_extended_kalman_filter.cc
```diff
@@ -0,0 +1,198 @@
...
+
+  Destroy();
+  return true;
+}
+
+void ObjectCameraExtendedKalmanFilter::Predict(const int track_id,
+    const double timestamp) {
+  double time_diff = timestamp - tracked_filters_[track_id].last_timestamp_;
+
+  // update transition model
```
Capri2014: @YajiaZhang You probably want to make a note here that this camera EKF is mainly for vehicle kinematics?

Capri2014: @YajiaZhang Also make a note saying x[0], x[1] is obstacle position in FLU, x[2] is theta heading angle, x[3] is velocity?

YajiaZhang: Not necessary. Although we assume the obstacles are vehicles, this works for non vehicles as well.

PR #4605, opened 2018-06-20
Planning: used a move constructor of the class hdmap::Path to improve the efficiency of object construction in the class ReferenceLine.

davidhopper2003:


davidhopper2003:
The original file "modules/canbus/vehicle/gem/gem_message_manager_test.cc" can't pass the code style checking and I have fixed it.

PR #4645, opened 2018-06-22
Prediction: add unit test for sequence_predictor.cc

kechxu:


modules/prediction/predictor/sequence/BUILD
```diff
@@ -22,4 +22,26 @@ cc_library(
...
+    ],
+    data = [
+        "//modules/prediction:prediction_data",
+        "//modules/prediction:prediction_testdata",
+    ],
+    deps = [
+        "//modules/common/util",
+        "//modules/prediction/common:kml_map_based_test",
+        "//modules/prediction/common:prediction_gflags",
+        "//modules/prediction/container/obstacles:obstacles_container",
```
Capri2014: correct the format here please

kechxu: @Capri2014  What kind of format did you refer to?

PR #4660, opened 2018-06-23
Third-party perception README.md

natashadsouza:


natashadsouza:
@Capri2014 good doc! made some small changes to it

PR #4703, opened 2018-06-26
planning: relax reference line shift condition

startcode:


startcode:
The code change passed simulation

Issue #4745, opened 2018-06-29
frame.cc: optimize CreateReferenceLineInfo method 


freeHackOfJeff:
We should break from the loop over immedinately if the reference information initization is succeed.

lianglia-apollo:
Thanks for your comments. Please make a pull request and we will review it.

Zhenni17:
Closing the issue.

PR #4749, opened 2018-06-29
frame.cc: optimize CreateReferenceLineInfo method [#4745]

freeHackOfJeff:
Break the loop over immediately once reference information is initialized successfully.


freeHackOfJeff:
Please note that, the "continue" could be removed with this change.

Issue #4764, opened 2018-06-30
sample_pnc_topics.py: fix the argument check up and use classmethod decorator


freeHackOfJeff:
1) We should show the usage as user friendly style.
2) Use classmethod instead of staticmethod decorator since the later is most comfortable resides in helper class.

natashadsouza:
Closing this issue as the changes have been made. Thanks!

PR #4765, opened 2018-06-30
sample_pnc_topics.py: fix the argument check up and use classmethod decorator [#4764]

freeHackOfJeff:
1) We should show the usage as user friendly style. 
2) Use classmethod instead of staticmethod decorator since the later is most comfortable resides in helper class.


xiaoxq:
Well done. Thank you very much!
Please sign the CLA above so we can merge it.

Issue #4768, opened 2018-07-02
Prediction::Init(): fix logic


freeHackOfJeff:
1) code style, please check if/else best practice.
2) fix the logic in case of offline_mode, otherwise it might lack of other conditions determination.

natashadsouza:
@freeHackOfJeff please submit a pull request for the changes you would like to make. Thanks!
Currently closing this issue.

PR #4769, opened 2018-07-02
Prediction init method fix [#4768]

freeHackOfJeff:
1) Code style, please check if/else best practice.
2) Fix the logic in case of offline_mode, otherwise it might lack of other conditions determination.


lianglia-apollo:
Please fix the CI error and re-submit. Thanks!

PR #4771, opened 2018-07-02
gen_pose_file.py: rewrote file [#4774]

freeHackOfJeff:
1) Argument checking up is invalid and not user friendly
2) Don't use print to write file
3) Rename pcd_file to captical letter since it is iteral, myabe I would refactor a couple of files to    consolidate this kind of stuff to another generic file in the future.
4) Rename main() to meaningful method name.


lianglia-apollo:
Please fix the CI error and re-submit.

Issue #4775, opened 2018-07-02
copy_bags_between_disks.py: code consolidation


freeHackOfJeff:
1) Don't hard code upon python interpreter version
2) Use capital letter for consistent value

PR #4776, opened 2018-07-02
Copy bags between disks fix

freeHackOfJeff:
copy_bags_between_disks.py: code consolidation #4775 
 1) Don't hard code upon python interpreter version
 2) Use capital letter for consistent value


xiaoxq:
Thanks for the great work! Could you sign the CLA so that we can merge the code?

PR #4780, opened 2018-07-02
docker: add conda and upgrade python to 2.7.15

ycool:


Capri2014:
@ycool @xiaoxq Do we plan to open the conda environment file as well?

xiaoxq:
I believe so. I can add the docker/build/installers/... later which
installs conda and as well as env config.

PR #4799, opened 2018-07-03
Optimize routine GetSubnodeByName

muleisheng:
reduce one copy operation

### DIFF
--- a/modules/perception/onboard/dag_streaming.cc
+++ b/modules/perception/onboard/dag_streaming.cc
@@ -230,7 +230,7 @@ void DAGStreamingMonitor::Run() {
   }
 }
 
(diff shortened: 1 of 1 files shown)


modules/perception/onboard/dag_streaming.cc
```diff
@@ -230,7 +230,7 @@ void DAGStreamingMonitor::Run() {
   }
 }
 
-Subnode* DAGStreaming::GetSubnodeByName(std::string name) {
+Subnode* DAGStreaming::GetSubnodeByName(const std::string &name) {
```
ycool: Thanks a lot for the contribution. The header file need be updated too to match the new change. 

Thanks.

PR #4818, opened 2018-07-03
VLS-128 velodyne driver added

velodynevls:
velodyne_vls folder is added to support Velodyne VLS-128 sensor for ROS.


velodynevls:
Hi @lianglia-apollo ,
The reason why we created a separate 'velodyne_vls' folder is that I don't know who manages 'velodyne' folder. I didn't want to make any changes or combine with someone else's work. Would it be okay if we keep 'velodyne_vls' folder under my responsibility and keep updating for our upcoming products?

PR #4857, opened 2018-07-06
control: added a flag to disable unnecessary navigation mode handling

YajiaZhang:


Capri2014:
@YajiaZhang The easies way is to set enable_navigation_mode=false in https://github.com/ApolloAuto/apollo/blob/master/modules/control/conf/control.conf for the test, right?

YajiaZhang:
There might be some parameter loading issue which we found when we were debugging in navigation mode. This flag might be overridden by other modules. The safest way is changing the parameter.

YajiaZhang:
In theory, there shouldn't be any flags that distinguish between absolute or relative frame in control. Therefore, the flag, along with the corresponding code, needs to be removed. The new flag is set for testing purpose and once tests confirm the theory, all these will be removed.

lianglia-apollo:
Yes, I thought you were changing something else..

PR #4860, opened 2018-07-07
docker: start supervisor when starting dev docker

startcode:


docker/scripts/dev_start.sh
```diff
@@ -289,6 +289,8 @@ function main(){
         docker exec apollo_dev bash -c '/apollo/scripts/docker_adduser.sh'
     fi
 
+    docker exec --user $USER apollo_dev bash -c "supervisord -c /apollo/modules/tools/supervisord/release.conf >& /tmp/supervisord.start.log"
```
zhxt: release.conf ?

And do we still need it in bootstrap.sh ? https://github.com/ApolloAuto/apollo/blob/0b1eb42abb10225928c2d3a40baa3b55d8075dfe/scripts/bootstrap.sh#L30-L33

startcode: removed in bootstrap

zhxt: Aha, great.  One more thing,  I think it's a copy-paste issue:
It should be `supervisord -c /apollo/modules/tools/supervisord/dev.conf ` instead of
`supervisord -c /apollo/modules/tools/supervisord/release.conf ` 
in dev_start.sh. Since the location of the  executable files are different in dev and release docker.

zhxt: To make it work with canary,  it also needs to be added to `docker/canary/start_container.py`.

PR #4869, opened 2018-07-08
correct a trivial hyperlink error which goes to a wrong page

DongAi:


docs/quickstart/apollo_3_0_quick_start_cn.md
```diff
@@ -7,7 +7,7 @@
 ## 内容
 - [车辆校准指南](#车辆校准指南)
 - [硬件和软件安装](#硬件和软件安装)
-- [DreamView的使用](https://github.com/ApolloAuto/apollo/new/master/docs/quickstart#dreamview%E7%9A%84%E4%BD%BF%E7%94%A8)
+- [DreamView的使用](#硬件和软件安装)
```
siyangy: #DreamView的使用?

DongAi: hello, I've done my best to translate this sentence 'DreamView Usage Table' to chinese, but I didn't find any other better choices, so, do you have any suggestions?

siyangy: I don't get that, shouldn't this just be a link to the later section? For now it's pointing to an incorrect one.

DongAi: Yes, you get the right point. It would be better if I explained it when I just committed it. there is a link bug in md file that clicking a text hyperlink can not go to the proper place labled with EN characters and CN characters. If all the characters are EN or CN, it's OK. But I could not tranlate 'DreamView Usage Table' to EN characters or CN characters. What I modified is a trick, clicking 'DreamView的使用' will go to place labled '硬件安装指南' which is always in the same page with 'DreamView的使用', it seems like we just go to the right place. I have tested several times, it won't make any confusions, you can test it by yourself. Trust me!

PR #4920, opened 2018-07-10
Dreamview: fix image topic in navigation mode

siyangy:


modules/dreamview/backend/dreamview.cc
```diff
@@ -68,7 +68,7 @@ void Dreamview::CheckAdapters() {
       << "RoutingRequestAdapter is not initialized.";
   CHECK(AdapterManager::GetRoutingResponse())
       << "RoutingResponseAdapter is not initialized.";
-  CHECK(AdapterManager::GetCompressedImage())
+  CHECK(AdapterManager::GetImageFront())
       << "CompressedImageAdapter is not initialized.";
```
mickeyouyou: ImageFrontAdapter init

siyangy: done

PR #4923, opened 2018-07-10
control: added trajectory transformation for navigation mode

YajiaZhang:


modules/control/controller/lat_controller.cc
```diff
@@ -269,8 +270,66 @@ Status LatController::ComputeControlCommand(
...
+    // new plannng trajectory
+    if (time_stamp_diff > 1.0e-6) {
+      init_vehicle_x_ = curr_vehicle_x;
+      init_vehicle_y_ = curr_vehicle_y;
+      init_vehicle_heading_ = curr_vehicle_heading;
+
+      current_trajectory_timestamp_ =
+          planning_published_trajectory->header().timestamp_sec();
+    } else {
+      auto x_diff = curr_vehicle_x - init_vehicle_x_;
```
Capri2014: @YajiaZhang We should add extra evaluation of std_dev of GPS/Localization here to determine whether we do the frame rotation here right? -- Only do this rotation when we have accurate enough localization signal.

modules/control/controller/lat_controller.cc
```diff
@@ -482,15 +541,8 @@ void LatController::ComputeLateralErrors(
     const double x, const double y, const double theta, const double linear_v,
     const double angular_v, const TrajectoryAnalyzer &trajectory_analyzer,
     SimpleLateralDebug *debug) {
-  // TODO(QiL): change this to conf.
-  TrajectoryPoint target_point;
-  if (FLAGS_use_navigation_mode) {
```
Capri2014: @YajiaZhang Similar here, Only Query by Time when 1. use_navigation_mode, 2. localization std_dev too high.

modules/control/controller/lat_controller.cc
YajiaZhang: This is equivalent to the case that control in absolute map frame with noisy localization. If we don't do the check in that case, we do not need to do this either. Also, the std dev can jump over time as well which will lead to unnecessary switching between time matching and position matching.

PR #4942, opened 2018-07-11
control: restructured lateral controller

YajiaZhang:


modules/control/controller/lat_controller.cc
```diff
@@ -514,7 +513,7 @@ void LatController::UpdateMatrixCompound() {
   matrix_bdc_.block(0, 0, basic_state_size_, 1) = matrix_bd_;
   if (preview_window_ > 0) {
     matrix_bdc_(matrix_bdc_.rows() - 1, 0) = 1;
-    // Update augument A matrix;
```
Capri2014: this is argument matrix of A

YajiaZhang: OK. Will be modified in later version.

PR #4958, opened 2018-07-12
planning: fixed the starting s issue in trajectory stitcher

YajiaZhang:


YajiaZhang:
The s value in the planning starting point was not consistent with the planned trajectory from last cycle and was not zero. Any changes to the common interface MUST be reviewed and approved by the corresponding author.

YajiaZhang:
It doesn't matter since the trajectory will be (and must be) discretized densely enough. Does it really matter that we matched current time to one point or its adjacent one? No, it does not, as long as we can get reasonable estimate of the control error and predict forward time is not too large. The accuracy you mentioned is the "false" accuracy we do not need. 

What definitely is not accurate is the linear approximation between trajectory points, which we should always avoid doing so in practice.

lianglia-apollo:
I agree it is a small difference, but it can be more accurate without additional effort, why not?

YajiaZhang:
We do not add code unless it is really necessary. In this case, we appear to make it more "accurate" (and not necessary as you agreed as well), but we need to worry about a couple of things more: 1. linear approximation between trajectory points that loses the true geometric and dynamic information from their original close-form discretization. We definitely cannot use this point as the planning starting point, we merely use it for estimation. 2. breaking the nice and natural assumption that the planned trajectory will always start from s_value = 0, therefore, in latter procedure we need to pay close attention to. This is the bug this pull request fixes.

Given these drawbacks, why should we do that :-)?

lianglia-apollo:
It won't break the assumption that planning always start at s = 0 if you add the matched point in Line 168 - Line 169.

PR #4977, opened 2018-07-13
control: fixed the control flag

YajiaZhang:


modules/control/controller/lat_controller.cc
```diff
@@ -269,6 +269,7 @@ Status LatController::ComputeControlCommand(
     const canbus::Chassis *chassis,
     const planning::ADCTrajectory *planning_published_trajectory,
     ControlCommand *cmd) {
+
   VehicleStateProvider::instance()->set_linear_velocity(chassis->speed_mps());
```
jilinzhou: - There is no need to set this speed again here. It was set in VehicleStateProvider's callback along with other chassis data. If you do a comparison before calling the setter in between VehicleStateProvider::instance()->linear_velocity() and chassis->speed_mps(), they are always the same.
- After ported Apollo 2.0 over QNX, We had been experiencing the steering control stability issue for our MKZ. To figure it out, I went through the technical report in which the dynamic model (adopted here) was derived (Automatic Steering Methods for Autonomous Automobile Path Tracking by Jarrod M. Snider). One possible cause (not 100% sure) I found is the speed used here. The chassis->mps give the speed of the center of front axle along the wheel's orientation. However, the reference frame of the dynamic model is based on the center of gravity of the car in which it expects the car's longitudinal velocity (v_x) for the lateral error rate (Section 4.1.1 Linearized Dynamic Bicycle Model, Equation (28)). The instability behaviour is quite obvious on a curvy road. After I switch it to the longitudinal velocity, the instability behaviour disappears right away but the car still slowly swings along the reference line. So I further tuned the control parameter q_4 to add some damping to the controller which ends with a acceptable result.

PR #4981, opened 2018-07-14
Planning: fixed scenario 79

lianglia-apollo:


lianglia-apollo:
Passed all daily build scenarios, where scenario 79 was passed but dangerous.

PR #4990, opened 2018-07-16
fix  usb_cam docs and unstable topic in Dreamview

mickeyouyou:
_sorry about so many history commits._

- usb_cam docs
- one camera topic display in dreamview become unstable ,when you play a rosbag that include topic ` /apollo/sensor/camera/obstacle/front_6mm`.


modules/dreamview/backend/handlers/image_handler.cc
```diff
@@ -78,11 +78,8 @@ void ImageHandler::OnImage(const sensor_msgs::CompressedImage &image) {
 }
 
 ImageHandler::ImageHandler() : requests_(0) {
-  if (FLAGS_use_navigation_mode) {
```
siyangy: We don't want to have both subscribers on as there are both the two cameras on the car and only one of them should be displayed depending on mode.

siyangy: @mickeyouyou This has been fixed by another PR, please resolve the conflict and squash all the commits into one, and then this PR is ready to merge.

mickeyouyou: OK

PR #5001, opened 2018-07-17
Revert "Planning: used condition_variable to update the reference lin…

lianglia-apollo:
…e for efficiency."

This reverts commit f3740cd794f1ca1bcc950146703cd85e88e13a5c.


lianglia-apollo:
Using vehicle state conditional variable will not be able to make the algorithm more efficient. In fact, it will make the reference smoother less efficient based on the current update frequency of vehicle state.

PR #5040, opened 2018-07-19
Planning: implemented append in piecewise jerk trajectory

kechxu:


modules/planning/lattice/trajectory1d/piecewise_jerk_trajectory1d.cc
```diff
@@ -20,8 +20,34 @@
 
 #include "modules/planning/lattice/trajectory1d/piecewise_jerk_trajectory1d.h"
 
+#include "glog/logging.h"
```
lianglia-apollo: use common/log.h

PR #5042, opened 2018-07-19
navi: added lidar based perception into navigation mode

yifeijiang:


yifeijiang:
@Capri2014 observe() is for relative map data.

PR #5073, opened 2018-07-20
planning:modify the obstacles' interface for navi_path_decider and na…

konyhl:
@yifeijiang
 modify the obstacles' interface for navi_path_decider and navi_speed_decider.

### DIFF
--- a/modules/planning/common/planning_gflags.cc
+++ b/modules/planning/common/planning_gflags.cc
@@ -180,6 +180,8 @@ DEFINE_bool(enable_nudge_decision, true, "enable nudge decision");
 DEFINE_bool(enable_nudge_slowdown, true,
             "True to slow down when nudge obstacles.");
 
+DEFINE_bool(enable_side_radar, false,
+            "If there is no radar on the side,ignore it");
 DEFINE_double(static_decision_nudge_l_buffer, 0.5, "l buffer for nudge");
 DEFINE_double(lateral_ignore_buffer, 3.0,
               "If an obstacle's lateral distance is further away than this "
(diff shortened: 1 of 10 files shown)


modules/planning/common/planning_gflags.cc
```diff
@@ -180,6 +180,8 @@ DEFINE_bool(enable_nudge_decision, true, "enable nudge decision");
 DEFINE_bool(enable_nudge_slowdown, true,
             "True to slow down when nudge obstacles.");
 
+DEFINE_bool(noenable_side_radar, true,
```
lianglia-apollo: Please use enable_side_rader

modules/planning/reference_line/reference_line_provider.cc
```diff
@@ -480,10 +480,14 @@ bool ReferenceLineProvider::GetNearestWayPointFromNavigationPath(
     double s = 0.0;
     double l = 0.0;
     if (!lane->GetProjection({point.x(), point.y()}, &s, &l)) {
+      AERROR << "get adc projection point on lane failed";
       continue;
     }
     constexpr double kEpsilon = 1e-6;
-    if (s > (lane->total_length() + kEpsilon) || (s + kEpsilon) < 0.0) {
+    if (s > (lane->total_length() + kEpsilon)) {
```
lianglia-apollo: Why removed the second condition?

konyhl: Under the relative map, the initial s does not always start from 0.0, and many times the s of the first few points is negative.
if do't remove ,have maney such error:
reference_line_provider.cc:537] failed to find nearest point x: 0 y: 0

lianglia-apollo: Is it possible to treat this issue inside navigation mode? Other modules could be affected if there are changes on shared API.

konyhl: There is no suitable local revision at present,
i will revoke this amendment.

konyhl:
@Liangliang Zhang
I have revised it according to your opinion.

PR #5102, opened 2018-07-24
type sync script

natashadsouza:


docs/howto/how_to_use_a_different_vehicle_cn.md
```diff
@@ -1,13 +1,4 @@
-# 如何选择使用不同的车辆
+# 如何在Dreamview中应用不同的车型
```
startcode: seems we still need to reset this file before checkin/

Issue #4868, opened 2018-07-08
fatal: Not a git repository (or any parent up to mount point /apollo)


liangzcamg:
root@in_dev_docker:/apollo# bash apollo.sh release
System check passed. Build continue ...
[WARNING] ESD CAN library supplied by ESD Electronics does not exist. If you need ESD CAN, please refer to third_party/can_card_library/esd_can/README.md.
INFO: Reading 'startup' options from /apollo/tools/bazel.rc: --batch_cpu_scheduling
fatal: Not a git repository (or any parent up to mount point /apollo)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

MirrorLakr:
kino@in_dev_docker:/apollo$ bash apollo.sh release
System check passed. Build continue ...
[WARNING] ESD CAN library supplied by ESD Electronics does not exist. If you need ESD CAN, please refer to third_party/can_card_library/esd_can/README.md.
INFO: Reading 'startup' options from /apollo/tools/bazel.rc: --batch_cpu_scheduling
.
fatal: Not a git repository (or any parent up to mount point /apollo)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

zhxt:
This may be a bug in apollo.sh: 

https://github.com/ApolloAuto/apollo/blob/1dc4b4c8e2087fe765914ba221a3e9bb6fcb2f40/apollo.sh#L283

https://github.com/ApolloAuto/apollo/blob/1dc4b4c8e2087fe765914ba221a3e9bb6fcb2f40/apollo.sh#L139

It does not check if it's git work tree or not before trying to get the revision with command `git rev-parse HEAD`.
So if you downloaded the Apollo source code  from the "Download ZIP' button, you'll encounter this problem, since the VCS info(.git dir) is removed from the ZIP ball by Github. 
It would be okay when you get the source with `git clone`.

Another similar issue is fixed in
https://github.com/ApolloAuto/apollo/blob/1dc4b4c8e2087fe765914ba221a3e9bb6fcb2f40/scripts/apollo_base.sh#L398-L403

And BTW, there is also a command ` git rev-parse --is-inside-work-tree` for this kind of check.

natashadsouza:
Closing this issue as a fix was submitted for it. If you are still seeing the error, feel free to open this issue.

PR #5140, opened 2018-07-26
scripts: do more check before a git rev-parse

zhxt:
Handle the situation that the source tree does not contain .git .
Fix error: 'Fatal: Not a git repository'.


zhxt:
patch for #4868

PR #5145, opened 2018-07-27
Fixes #5120

### LINKED ISSUE #5120: The missing declaration of "use_navigation_mode" in third_party_perception_flags.h causes the module build failure.
The modules/third_party_perception/conversion.cc has references to FLAGS_use_navigation_mode but it is neither declared or defined in modules/third_party_perception/common/third_party_perception_flags.h/third_party_perception_flags.cc. After adding the declaration and definition, the module can be built successfully.

jilinzhou:
The modules/third_party_perception/conversion.cc has references to FLAGS_use_navigation_mode but it is neither declared or defined in third_party_perception_flags.h or third_party_perception_flags.cc. After adding the declaration and definition, the module can be built successfully.


jilinzhou:
@lianglia-apollo :  as a note, it is needed for r2.5.0 as well.

PR #5201, opened 2018-08-01
Perception: fixed bug in Yuyv2rgb conversion for AVX non-compliant CPUs.

hadiTab:
The function Yuyv2rgb has two definitions; one uses AVX instructions and the other does not. This commit fixes a bug in the latter function definition. The ordering of the color channels in the input image were incorrect which would result in an incorrect conversion to rgb which subsequently affects the perception results. The bug has been fixed by correcting the channel order the function expects.


lianglia-apollo:
This fix looks correct according to the definition here: http://www.fourcc.org/pixel-format/yuv-yuy2/

PR #5229, opened 2018-08-03
Fix bug about time.

xinwf:
When carrying out a tf transform in localization, if I use real objects to get time, this will cause a error in time. As you can see the screenshot about the output in the log:<br>
This happens in this cpp file.
screenshot from 2018-08-03 23-02-47
<br>This is the output of the tf topic.
screenshot from 2018-08-03 23-18-49

Time is different in two pictures, because I get these two pics in different time, but I think this should be able to demonstrate that it is an error when I use real objects to get the time. It shouldn't be given a value with ros::Time.


xinwf:
you mean like this? in`modules/common/configs/config_gflags.cc` sets a bool variable
```
 if(use_ros_time)
      tf2_msg.header.stamp = ros::Time(localization.measurement_time());
 else
      tf2_msg.header.stamp = localization.measurement_time();
```

lianglia-apollo:
There is a time class at https://github.com/ApolloAuto/apollo/blob/master/modules/common/time/time.h. You can use APIs in that file.

lianglia-apollo:
Please address the comments and re-submit.

PR #5230, opened 2018-08-03
allow sim_control to initialize start point from localization

unacao:


modules/dreamview/backend/sim_control/sim_control.cc
```diff
@@ -107,6 +98,57 @@ void SimControl::OnReceiveNavigationInfo(
...
+    point.mutable_path_point()->set_x(pose.position().x());
+    point.mutable_path_point()->set_y(pose.position().y());
+    point.mutable_path_point()->set_z(pose.position().z());
+    point.mutable_path_point()->set_theta(pose.heading());
+    start_velocity = std::hypot(pose.linear_velocity().x(),
+                                pose.linear_velocity().y());
+    // Calculates the dot product of acceleration and velocity. The sign
+    // of this projection indicates whether this is acceleration or
+    // deceleration.
+    double projection = pose.linear_acceleration().x()
```
siyangy: Let's try not to override function param.

modules/dreamview/backend/sim_control/sim_control.cc
```diff
@@ -107,6 +98,57 @@ void SimControl::OnReceiveNavigationInfo(
   }
 }
 
+bool SimControl::InitStartPoint(double start_velocity,
+                                double start_acceleration) {
+  TrajectoryPoint point;
+  bool actual_position = false;
```
siyangy: acutal_position can just be replaced as start_point_set, making the state mutation easier to follow.

modules/dreamview/backend/sim_control/sim_control.cc
```diff
@@ -66,6 +66,13 @@ SimControl::SimControl(const MapService* map_service)
 
 void SimControl::Init(bool set_start_point, double start_velocity,
                       double start_acceleration) {
+  if (set_start_point && !FLAGS_use_navigation_mode) {
+    start_point_set_ = InitStartPoint(start_velocity, start_acceleration);
+  }
+  if (inited_) {
```
siyangy: There could be a better separation about exactly one time setup init (callback, timer, etc.) and start point init.

modules/dreamview/backend/sim_control/sim_control.cc
```diff
@@ -107,6 +102,56 @@ void SimControl::OnReceiveNavigationInfo(
...
 }
 
+void SimControl::InitStartPoint(double start_velocity,
+                                double start_acceleration) {
+  TrajectoryPoint point;
+  // Use the latest localization position as start point,
+  // fall back to a dummy point from map
+  AdapterManager::GetLocalization()->Observe();
+  if (AdapterManager::GetLocalization()->Empty()) {
+    start_point_set_ = false;
```
siyangy: can we have a more accurate naming here? Setting start_point_set_  to false seems to contradict with what the code actually does here.

modules/dreamview/backend/sim_control/sim_control.h
```diff
@@ -126,12 +130,18 @@ class SimControl : SimControlInterface {
...
 
+  // Whether planning has requested a re-routing.
   bool re_routing_triggered_ = false;
 
-  // Whether the sim control is enabled / initialized.
+  // Whether the sim control is enabled.
   bool enabled_ = false;
+
+  // Whether the adapter setup has been initialized.
   bool inited_ = false;
```
siyangy: likewise, inited_ seems to be too broad, maybe rename it to something more explicit?

PR #5269, opened 2018-08-07
A tool to evaluate body sensation

JackFu123:


modules/tools/rosbag/body_sensation_evaluation.py
```diff
@@ -0,0 +1,224 @@
+# -*- coding: utf-8 -*-
```
Capri2014: utf-8 ?

modules/tools/rosbag/body_sensation_evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+        """check driving mode according to timestamp"""
+        for mode in self.driving_mode:
+            if timestamp >= mode["start_time"] and timestamp <= mode["end_time"]:
+                if mode["status"] == Chassis.COMPLETE_AUTO_DRIVE:
+                    return True
+                else:
+                    return False
+        return False
+
+    def _bumps_rollback(self, bump_time):
```
Capri2014: use bumps_rollback_ instead

modules/tools/rosbag/body_sensation_evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+                            if self._check_status(t):
+                                self.auto_counts["turning"] += 1
+                            else:
+                                self.manual_counts["speed_up_2"] += 1
+                        self._turning_flag = 0
+                    elif abs(acc_x) >= -2:
+                        self._turning_flag = 1
+
+
+def main():
```
Capri2014: Add usage instruction here

modules/tools/rosbag/body_sensation_evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+                    else:
+                        self.manual_counts["bumps"] += 1
+                else:
+                    if self._speed_down_2_flag:
+                        if acc_y <= -4:
+                            self._speed_down_4_flag = 1
+                            continue
+                        if acc_y <= -2:
+                            continue
+                        if self._speed_down_4_flag == 1 and t - self._last_speed_down_4_time >=1:
```
Capri2014: Similar issue

modules/tools/rosbag/body_sensation_evaluation.py
```diff
@@ -0,0 +1,224 @@
...
+                #时间戳
+                #dt = time.strftime('%Y-%m-%d %H:%M:%S',time.localtime(_t))
+                self.timestamp = t
+                if t - self._bump_lock_time <= 3:
+                    continue
+                acc_x = msg.pose.linear_acceleration.x
+                acc_y = msg.pose.linear_acceleration.y
+                acc_z = msg.pose.linear_acceleration.z
+                diff_bump_time = t - self._last_bump_time
+                if abs(acc_z) >= 2 and diff_bump_time >= 3:
```
Capri2014: make 2 and 3 const variable

modules/tools/rosbag/body_sensation_evaluation.py
JackFu123: Could you explain why? This method should be protected.

modules/tools/rosbag/body_sensation_evaluation.py
JackFu123: Usage instruction is in the front.

modules/tools/rosbag/body_sensation_evaluation.py
JackFu123: python doesn't have "const" like cpp. Have to define a new class if const variable is needed.

modules/tools/rosbag/body_sensation_evaluation.py
JackFu123: This line has been deleted

modules/tools/rosbag/body_sensation_evaluation.py
Capri2014: similar to kMaxAccleration

modules/tools/rosbag/body_sensation_evaluation.py
Capri2014: sorry my bad here

PR #5287, opened 2018-08-07
common: fixed lint and added test for Encode64

lianglia-apollo:


modules/common/util/string_util.cc
```diff
@@ -24,13 +24,13 @@ namespace common {
...
 
-static const std::string kBase64Array =
+const char kBase64Array[] =
     "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
 // A table which maps a char to its value in Base64 mode.
 std::vector<int> Base64CodeTable() {
   std::vector<int> table(256, -1);
-  for (size_t i = 0; i < kBase64Array.length(); ++i) {
+  for (size_t i = 0; i < strlen(kBase64Array); ++i) {
```
startcode: strlen() will be called multiple times here. Save strlen first.

startcode: strlen() will be called multiple times here. Save strlen first.

xiaoxq: I made similar change, but it's good to fix things here.
My two cents:
1. Remove "auto table = kBase64Array;"
2. Unify names of DecodeBase64 and EncodeBase64.

PR #5298, opened 2018-08-08
pnc_map: avoid loop in route segments

startcode:


startcode:
Simulation 100% passed

PR #5308, opened 2018-08-09
planning: added piecewise quintic spiral path for future use

YajiaZhang:


modules/planning/math/curve1d/piecewise_quintic_spiral_path.cc
```diff
@@ -0,0 +1,64 @@
...
+
+double PiecewiseQuinticSpiralPath::DeriveKappaS(const double s) const {
+  auto index = LocatePiece(s);
+  const auto& piece = pieces_[index];
+  double ratio = (s - accumulated_s_[index]) / piece.ParamLength();
+  return piece.DeriveKappaDerivative(QuinticSpiralPath::DELTA_S, ratio);
+}
+
+std::size_t PiecewiseQuinticSpiralPath::LocatePiece(const double s) const {
+  CHECK(s >= accumulated_s_.front() && s <= accumulated_s_.back());
```
startcode: it is better to tolerate this or return an error code.

modules/planning/math/curve1d/piecewise_quintic_spiral_path.h
```diff
@@ -0,0 +1,60 @@
...
+  void Append(QuinticSpiralPath spiral_path);
+
+  double Evaluate(const std::uint32_t order,
+      const double param) const override;
+
+  double DeriveKappaS(const double s) const;
+
+  double ParamLength() const override;
+
+  std::string ToString() const override { return ""; };
```
startcode: you can return the class name here, instead of nothing.

modules/planning/math/curve1d/piecewise_quintic_spiral_path.cc
YajiaZhang: The class is dedicated for special purpose. The situation must not happen. So it is OK to have a strong assertion here.

modules/planning/math/curve1d/piecewise_quintic_spiral_path.h
YajiaZhang: OK

YajiaZhang:
We don't have any type of spiral path yet and we do not have plans to add new types in the near future. There is no need to add the general class.

PR #5315, opened 2018-08-09
Dreamview: fix a deadlock case

siyangy:


modules/dreamview/backend/handlers/websocket_handler.cc
```diff
@@ -46,20 +46,25 @@ void WebSocketHandler::handleReadyState(CivetServer *server, Connection *conn) {
...
-    {
-      std::unique_lock<std::mutex> lock(*connection_lock);
-      connections_.erase(connection);
-    }
-    AINFO << name_
-          << ": Connection closed. Total connections: " << connections_.size();
+  {
+    // Make sure there's no data being sent via the connection
+    std::unique_lock<std::mutex> lock_connection(*connection_lock);
+    std::unique_lock<std::mutex> lock(mutex_);
```
startcode: Is there a way to lock two mutexes in an atomic way?

siyangy: the other reference of this is not a scoped lock, meaning that we call ->lock, then ->unlock.

PR #5321, opened 2018-08-10
Perception: minor fixes.

csukuangfj:


modules/common/macro.h
```diff
@@ -15,7 +15,7 @@
  *****************************************************************************/
 
 /**
- * @file
+ * @file macro.h
```
startcode: you can remove macro.h
doxygen can fill in this field automatically

csukuangfj: @startcode
thank you. Fixed.

PR #5322, opened 2018-08-10
Perception: avoid unnecessary copy.

csukuangfj:


modules/perception/lib/base/registerer.h
```diff
@@ -138,7 +138,7 @@ bool GetRegisteredClasses(
       FactoryMap &map = perception::GlobalFactoryMap()[#base_class];  \
       FactoryMap::iterator iter = map.find(name);                     \
       if (iter == map.end()) {                                        \
-        for (auto c : map) {                                          \
+        for (const auto &c : map) {                                   \
```
startcode: Nice catch. Thanks for the change.
 But can you also replace `map` by a better variable name
Thanks.

csukuangfj: @startcode
thanks for the comment.

I don't think map is a bad name in this context. If you can suggest a new name, I would like to replace it.

csukuangfj: `object_factory_map` is too long and it exceeds 80 columns per row. In addition, the backslash
at the end of the line will not be aligned.

startcode: `map`  fine. it a minor thing anyway.

PR #5331, opened 2018-08-11
Dreamview: Fix Lidar point cloud visualization

andytolst:
There is a blind spot on point cloud caused by voxel grid filter
when filter is executed in-place (output is the same as input).


unacao:
I have no problem with your change. But could you help specify in what situation the output will be the same as the input? In our own tests, the output does change based on the filter size although it is done in-place. Thanks!

andytolst:
Sorry for confusion about input/output, the output will be filtered, but the result is incorrect (some points are missing). I'll try to attach a picture, the missing sector is clearly visible if you see from 'Overhead' point of view.

andytolst:
Here is a screenshot: before the fix (see the left top corner):
lidar_before

with the fix:
lidar_after

andytolst:
BTW, you might want to take a look at this:
https://github.com/PointCloudLibrary/pcl/issues/1041

and this:
https://github.com/PointCloudLibrary/pcl/pull/1042

basically this should have been fixed in the pcl library, but Apollo might use the different version ?

unacao:
Thanks for the links!

PR #5371, opened 2018-08-16
Added more targets in cibuild for integration test.

lianglia-apollo:


apollo.sh
```diff
@@ -388,6 +388,10 @@ function run_test() {
 
 function citest() {
   BUILD_TARGETS="
+  modules/routing/graph:sub_topo_graph_test
```
startcode: recommend use blacklist, instead of white list

PR #5422, opened 2018-08-20
verify_client: refactor this program via an init method

freeHackOfJeff:
1. It's better to isoldate the init_secure_upgrade of sec_api into a method.
2. Use capital letter to define the constant variable
3. Check up boolean type via 'is' rather than '=='


modules/tools/diagnostics/message.py
```diff
@@ -93,9 +93,9 @@ def callback(self, data):
             time = 0
             sequence_num = 0
 
-        if self.msg_received == True:
+        if self.msg_received:
             seq_diff = sequence_num - self.sequence_num
-            if seq_diff is not 0:
+            if seq_diff == 0:
```
kernyan: Is this change intended? "is not" changed to "=="

lianglia-apollo:
Please add the modified files and re-submit.

PR #5429, opened 2018-08-21
Cos theta qp

JasonZhou404:


modules/planning/common/planning_gflags.cc
```diff
@@ -40,6 +40,10 @@ DEFINE_string(smoother_config_filename,
               "modules/planning/conf/qp_spline_smoother_config.pb.txt",
               "The configuration file for qp_spline smoother");
 
+DEFINE_string(reOpt_smoother_config_filename,
```
startcode: don't use capital case in FLAG names

PR #5466, opened 2018-08-23
Offlineview: fix wrong decision position issue & show geolocation in map view & add 'position' section in menu layer

vlin17:


vlin17:
Two additional double-counted offset found: 
1. No need to add offset in FindNudgeRegion(), since the offset was applied in SetObstaclePolygon()
2. In UpdateMainStopDecision(), only the "normal stop" needs to apply the offset.

PR #5511, opened 2018-08-27
Docker: Move apt-get commands to earlier and add apt-get update 

redfuture:
before the install.

apt-get commands part is more stable than the installer part, which including build
steps,this may could fasten the docker image build process when there are only
installers changed.


xiaoxq:
Good point!

Actually installers/pre_install.sh did this part. Could you move the content out and remove the pre_install.sh.
Or the other way, move the commands into pre_install.sh and run it first:
```
COPY installers/pre_install.sh /tmp/installers/pre_install.sh
RUN bash /tmp/installers/pre_install.sh

COPY installers /tmp/installers
...
```

redfuture:
Thanks for your review, I didn't notice the pre_install.sh , I will make  another patch to move out the pre_install.sh content out and merge with the apt-get commands, this looks like more readable.

redfuture:
I pushed again to the same branch , please review. thanks.

docker/build/dev.x86_64.dockerfile
```diff
@@ -2,37 +2,45 @@ FROM nvidia/cuda:8.0-cudnn7-devel-ubuntu14.04
...
-   realpath \
-   shellcheck \
-   vim \
-   v4l-utils \
-   nfs-common \
-   zip && \
-   apt-get clean && rm -rf /var/lib/apt/lists/*
+RUN apt-get update -y && \
+    apt-get install -y \
+    apt-transport-https \
```
xiaoxq: Please make this list in sorted order. So that it's easier for future lookup and modification.

redfuture: thanks for your quick review, I sorted the list and pushed another commit on the same branch. Please review again.

xiaoxq:
Excellent, thanks!

PR #5523, opened 2018-08-28
created Planningcontext class to support planning status and other

startcode:
persistant storage service


startcode:
simulation result is the same as before change

PR #5555, opened 2018-08-29
Planning : Change planning base interface to de-couple from reference…

Capri2014:
… line and frame


Capri2014:
@JasonZhou404 @JasonXu12

PR #5601, opened 2018-08-31
Planning: added dummy scenarios and scenario manager.

lianglia-apollo:


modules/planning/scenarios/lane_follow/lane_follow_scenario_test.cc
```diff
@@ -0,0 +1,46 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
+ *
```
jinghaomiao: 2018

modules/planning/scenarios/scenario_test.cc
```diff
@@ -0,0 +1,46 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
+ *
```
jinghaomiao: 2018

PR #5663, opened 2018-09-05
Planning: fix no planning output when a car stops behind adc

JasonZhou404:


modules/planning/toolkits/deciders/front_vehicle.cc
```diff
@@ -302,6 +302,7 @@ std::string FrontVehicle::FindPassableObstacle(
 
 void FrontVehicle::MakeStopDecision(ReferenceLineInfo* reference_line_info) {
   const auto& adc_sl = reference_line_info->AdcSlBoundary();
+  const auto& vehicle_sl = reference_line_info->VehicleSlBoundary();
```
lianglia-apollo: Please put variable declaration as close as where you will use it.

modules/planning/common/reference_line_info.h
```diff
@@ -203,6 +204,8 @@ class ReferenceLineInfo {
 
   SLBoundary adc_sl_boundary_;
 
+  SLBoundary vehicle_sl_boundary_;
```
lianglia-apollo: what is the difference between adc_sl_boundary_ and vehicle_sl_boundary_?

JasonZhou404: adc_sl_boundary is generated by the stitching point while the latter is generated by the realtime vehicle position

ycool: can you add these comments for these two variables? thanks.

PR #5681, opened 2018-09-06
Planning: calculate front clear distance.

lianglia-apollo:


Capri2014:
should not include reference line info in ego related information

PR #5722, opened 2018-09-10
Tools/map_gen: Use list comprehension to replace multi line for loop.

redfuture:
In the old code it's missing f.close when using open.This patch use
with statement when open file.


modules/tools/map_gen/map_gen.py
```diff
@@ -92,14 +92,9 @@ def create_lane(map, id):
...
 fpath = sys.argv[1]
-f = open(fpath, 'r')
-points = []
-for line in f:
-    line = line.replace("\n", '')
-    data = line.split(',')
-    x = float(data[0])
-    y = float(data[1])
-    points.append((x, y))
+with open(fpath, 'r') as f:
```
lianglia-apollo: Complicated comprehensions or generator expressions can be hard to read. We don't recommend expressions that cannot be fit in one line. Please see ```https://github.com/google/styleguide/blob/gh-pages/pyguide.md``` for details.

redfuture: got that, thanks for pointing that out , I will submit another pull request to just use with statement with open file.

redfuture: please close this pull request.

lianglia-apollo: Thanks!

PR #5723, opened 2018-09-10
Tools/map_gen: Use with statement to open file.

redfuture:
Lacking f.close after called open in the old code, though this is
not mandatory, use with statement is better.


Pumelhumel:
5723

PR #5725, opened 2018-09-10
Planning: used std::async for reference line provider.

lianglia-apollo:


lianglia-apollo:
hold for new operation system.

lianglia-apollo:
close as this change will not be necessary in new release.

PR #5749, opened 2018-09-13
Dreamview: implement connection lost detection and consolidate uttera…

vlin17:
…nce logic


modules/dreamview/frontend/src/store/websocket/websocket_realtime.js
```diff
@@ -1,14 +1,15 @@
...
+import UTTERANCE from "store/utterance";
 import Worker from 'utils/webworker.js';
 
 export default class RosWebSocketEndpoint {
     constructor(serverAddr) {
         this.serverAddr = serverAddr;
         this.websocket = null;
         this.simWorldUpdatePeriodMs = 100;
-        this.simWorldLastUpdateTimestamp = 0;
+        this.simWorldLastUpdateTimestamp = new Date().getTime();
```
siyangy: checkMessage relies on the fact that this variable is initially 0. The connection lost alert can follow the same check.

vlin17: Good catch. Fixed the logic.

siyangy:
@vlin17 Does this need to be in stable branch as well?

vlin17:
Yes, will cherry-pick to stable branch once merged.

PR #5768, opened 2018-09-18
Update interpolation_2d.cc

lianglia-apollo:


lianglia-apollo:
will wait for new release to fix.

PR #5786, opened 2018-09-21
RelativeMap: expanded a specified number of lanes on both left and right sides based on the configuration file.

davidhopper2003:


davidhopper2003:
screenshot from 2018-09-21 10-34-41

PR #5791, opened 2018-09-23
Fix typos

SohamGhormade:


modules/prediction/network/net_layer.h
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright 2017 The Apollo Authors. All Rights Reserved.
+ * Copyright 2018 The Apollo Authors. All Rights Reserved.
```
lianglia-apollo: Please using the time the original file was uploaded.

modules/prediction/network/net_model.h
```diff
@@ -75,8 +75,7 @@ class NetModel {
   virtual void State(std::vector<Eigen::MatrixXf>* states) const {}
 
   /**
-   * @brief Set the internal state of a model
-   * @param A specified internal state in a vector of Eigen::MatrixXf
+   * @brief Reset the internal state of a network model
    */
   virtual void ResetState() const {}
```
zhuangli1987: it's better to change those virtual functions to pure virtual functions

modules/prediction/network/net_layer.h
SohamGhormade: Thanks for clarifying--if this was a new file when we would have added the current year in the copyright notice.
Will revert this part of the changeset.

modules/prediction/network/net_model.h
SohamGhormade: Agreed.Will include this change in my next commit.Thanks!

SohamGhormade: ```AutotuningMLPModel``` implements NetModel but does not override ResetState,State and SetState.
We need to implement these methods for that class and write unit tests before declaring these methods as pure virtual functions in the base class.
Looking into it.

modules/planning/tuning/autotuning_mlp_net_model.h
```diff
@@ -39,6 +39,23 @@ class AutotuningMLPModel : public prediction::network::NetModel {
    */
   void Run(const std::vector<Eigen::MatrixXf>& inputs,
            Eigen::MatrixXf* output) const override;
+
+  /**
+   * @brief Set the internal state of a network model
+   * @param A specified internal state in a vector of Eigen::MatrixXf
+   */
+   void SetState(const std::vector<Eigen::MatrixXf>& states)override{}
```
SohamGhormade: It might be better to revert to declaring the State related  functions as virtual.
Alternatively we could implement these methods and use them in unit tests.

modules/prediction/network/net_model.h
```diff
@@ -66,19 +66,18 @@ class NetModel {
    * @brief Set the internal state of a network model
    * @param A specified internal state in a vector of Eigen::MatrixXf
    */
-  virtual void SetState(const std::vector<Eigen::MatrixXf>& states) {}
+  virtual void SetState(const std::vector<Eigen::MatrixXf>& states){}
```
SohamGhormade: The spacing change was unintended but it passed the lint run locally.

SohamGhormade:
@lianglia-apollo  please take a second look and please approve and merge if the PR looks good.
Thanks!

lianglia-apollo:
> @lianglia-apollo please take a second look and please approve and merge if the PR looks good.
> Thanks!

We are temporally holding all pull requests util the next release. Thanks.

PR #5934, opened 2018-10-26
translated CONTRIBUTING.md in korean

taewankang:
I translated this file into Korean so that many Koreans could be interested.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,64 +1,59 @@
-## How to Contribute to Apollo
+## 아폴로에 기여 하는 법
 
-### Contributor License Agreements
+### 기여 허가증 동의서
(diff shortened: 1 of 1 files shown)


zhxt:
Would it be better to put the different translation to a separate file and name the file with a language suffix instead of mixing them together? 

eg. CONTRIBUTE_cn.md(or _zh) for Chinese.

zhxt:
Thanks for contributing, we almost there, please keep the default English version as it is(do not delete it), and add a link for the translated version(_LANG-SUFFIX.md) to it somewhere( maybe at the beginning after the title?) , so developers can easily select the version they want. 
I prefer this way.  How do you think? @taewankang @lianglia-apollo

PR #5971, opened 2018-11-01
Bug fixes in open-loop controller calibration tools

hadiTab:
This pull request addresses two issues in the calibration tools for the open-loop controller:

1. The data_collection.py script did not properly send brake commands when three arguments were provided as input. This has been fixed in [bd39136](https://github.com/ApolloAuto/apollo/commit/bd39136a892b8fac3da287bb0e6a0b73f22c45dc).

2. The processing.py script finds the first usable index by searching for the last entry where the brake percentage is at (exactly) 40. This is only useful for a case where the stationary vehicle always keeps the brake at 40 percent. I've added an additional check for cases where a brake percentage of 40 is not present in the data. In these cases, the velocity of the vehicle will be used to find the index of the first entry to be used for calibration. This is addressed in [3890654](https://github.com/ApolloAuto/apollo/commit/3890654aff693429760c7f70308b486bfd43f653).

### DIFF
--- a/modules/tools/calibration/data_collector.py
+++ b/modules/tools/calibration/data_collector.py
@@ -157,11 +157,11 @@ class DataCollector(object):
                 self.case = 'd'
         elif self.case == 'd':
             if self.cmd[2] > 0:
-                self.controlcmd.throttle = self.cmd[0]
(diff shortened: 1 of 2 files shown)


natashadsouza:

Issue #5272, opened 2018-08-07
misleading name ObjectCameraExtendedKalmanFilter::UpdateTransitionMatrix


csukuangfj:
See the code
https://github.com/ApolloAuto/apollo/blob/4f79a4fd0bf3551dda1eb262b5f539b25a1f6d40/modules/perception/obstacle/camera/filter/object_camera_extended_kalman_filter.cc#L174-L182

It simply computes a Jacobian matrix and should not be called a transition matrix.

freshnewegg:
I agree, this is just the jacobian -> I propose to change it to 

"GetTransitionalModelJacobian" or "GetTransitionalUpdateMatrix"

I prefer the former. If agreed upon, I can do pr

apollodevcommunity:
you shall concern about where the caller use the return value matrix.acturally   this function is not used now. ——Reply reference comes from the Apollo developer community issue master contest https://apollo.auto/developer/questiondata_cn.html?target=649

PR #5995, opened 2018-11-07
Reworded method name in extended kalman filter

freshnewegg:
https://github.com/ApolloAuto/apollo/issues/5272


jinghaomiao:
Please check into v3.1_dev branch instead of master. Thanks!

PR #6027, opened 2018-11-15
I translated demo_guide into koreans

taewankang:
I translated demo_guide into koreans so that koreans also understand.


jinghaomiao:
@taewankang

PR #6028, opened 2018-11-15
I translated CONTRIBUTING.md into koreans.

taewankang:
I translated CONTRIBUTING.md into koreans so that koreans also understand.


jinghaomiao:
@taewankang

PR #6046, opened 2018-11-18
修复使用China mirror时，启动的却是US mirror里的镜像问题

tsxuehu:


jinghaomiao:
I don't see the original implementation has any issues.

PR #6059, opened 2018-11-20
CHECK_NOTNULL to CHECK

geoffviola:
It's better to use CHECK_NOTNULL, when the pointer is being used on the same line. There are gcc warnings, when compiling DCHECK_NOTNULL with optimization: https://github.com/google/glog/blob/master/src/glog/logging.h.in#L1061.

The transformation was applied with
```
git grep -l 'CHECK_NOTNULL' | xargs sed -i -r 's/CHECK_NOTNULL\(([^)]+)\);/CHECK(\1);/g'
```


geoffviola:
Continued in new PR: https://github.com/ApolloAuto/apollo/pull/6084

PR #6073, opened 2018-11-26
Add keys to change extrinsic matrix in perception viewer

GeorgeMD:
You can change the extrinsic matrix of the camera while the perception viewer runs. It should help with calibrating the camera.
The controls are as follows:
M - change rotation mode or go back to normal mode
In rotation mode:
W,S - rotate forward, backward (around the axis perpendicular to the direction of the car)
A,D - rotate left, right (around the direction of the car)
Q.E - rotate left, right (around the vertical axis)

N - change translation mode or go back to normal mode
W,S - forward, backward
A,D - left, right
Q,E - up, down

Z - save the matrix to /apollo/extrinsics.yaml


natashadsouza:

PR #6084, opened 2018-11-30
CHECK_NOTNULL to CHECK

geoffviola:
It's better to use CHECK_NOTNULL, when the pointer is being used on the same line. There are gcc warnings, when compiling DCHECK_NOTNULL with optimization: https://github.com/google/glog/blob/master/src/glog/logging.h.in#L1061.

The transformation was applied with

```
git grep -l 'CHECK_NOTNULL' | xargs sed -i -r 's/CHECK_NOTNULL\(([^)]+)\);/CHECK(\1);/g'
```

on mac (BSD sed), it's
```
git grep -l 'CHECK_NOTNULL' | xargs sed -i '' -E 's/CHECK_NOTNULL\(([^)]+)\);/CHECK(\1);/g'
```

### DIFF
--- a/modules/common/math/line_segment2d.cc
+++ b/modules/common/math/line_segment2d.cc
@@ -74,7 +74,7 @@ double LineSegment2d::DistanceTo(const Vec2d &point) const {
 
 double LineSegment2d::DistanceTo(const Vec2d &point,
                                  Vec2d *const nearest_pt) const {
-  CHECK_NOTNULL(nearest_pt);
+  CHECK(nearest_pt);
   if (length_ <= kMathEpsilon) {
     *nearest_pt = start_;
     return point.DistanceTo(start_);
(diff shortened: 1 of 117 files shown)


xiaoxq:
Actually I suggest removing CHECK the output parameter of functions:
void Func(const InputType& input, Output* output);

IMHO it's a global practice everywhere and in most occasions it's the caller's responsibility to guarantee a valid output pointer. So we can simply remove these checks to make code shorter.

geoffviola:
The intention of the current MR is to remove warnings when building with optimization.

@xiaoxq What about input/output parameters? Should they also not have a check? This change touches a lot of files, it might be good to change things incrementally.

lianglia-apollo:
> The intention of the current PR is to remove warnings when building with optimization.
> 
> @xiaoxq What about input/output parameters? Should they also not have a check? This change touches a lot of files, it might be good to change things incrementally.

The warning from DCHECK is intended, as we don't need those checks in online environment, but need them in off-line mode.

geoffviola:
Don't you want to eliminate the false positive warnings so that other more important one can pop up? For example, "-Wuninitialized" is more active when optimization is enabled.

PR #6127, opened 2018-12-14
dreamview: fix the rss info.

kiboliu:
Separate rss info from monitor items, and remove rss marker boolean from adc.


modules/dreamview/frontend/src/renderer/adc.js
```diff
@@ -27,8 +27,8 @@ export default class AutoDrivingCar {
         this.rssUnsafeMarker = new THREE.Object3D();
         this.rssUnsafeMarker.add(rssUnsafeMesh);
         this.rssUnsafeMarker.mesh = rssUnsafeMesh;
-        this.rssUnsafeMarkerAddedToScene = false;
-        this.scene = scene;
+        this.rssUnsafeMarker.visible = false;
```
vlin17: directly assign "this.rssUnsafeMarker = rssUnsafeMesh;" would be good enough

PR #6132, opened 2018-12-14
Planning& Dreamview: Idle gear shift time

JasonZhou404:
1. Reverse Path point theta in Planning when in reverse gear
2. Idle Time for gear shifting
3. related fix on Sim control "PublishLocalization" by reverting the reverse of theta when in reverse gear because of point 1 above @siyangy @unacao


Capri2014:
LGTM, @siyangy @unacao Please also review

unacao:
is this tested in Dreamview already? Does the car show the correct forward heading with backward motion when in reverse gear?

PR #6147, opened 2018-12-15
Perception: Give a some vote for confidence 0 detection

techoe:


techoe:
This change will solve unknown traffic light detection when the ego-vehicle is very close to a traffic light.

KaWaiTsoiBaidu:
LGTM

PR #6148, opened 2018-12-15
Tools: modified calibration script to allow starting from stationary position as well

hadiTab:
The processing.py script finds the first usable index by searching for the last entry where the brake percentage is at (exactly) 40. This is only useful for a case where the stationary vehicle always keeps the brake at 40 percent. I've added an additional check for cases where a brake percentage of 40 is not present in the data. In these cases, the velocity of the vehicle will be used to find the index of the first entry to be used for calibration.


hadiTab:
This is the second part of https://github.com/ApolloAuto/apollo/pull/5980

PR #6158, opened 2018-12-15
Data: Retire rosbag_to_record tool.

xiaoxq:


xiaoxq:
> Do we want to retire it now? Or our partners might still need it as of now?

Actually this tool is not a full implementation with all channels converted. It's misleading to provide an incomplete tool. For those who need such a tool:
1. They are supposed to continue working on the ROS based Apollo to process the data.
2. Or, they can easily implement their own tool to convert bags to records.

Capri2014:
I see, got it. Thanks!

PR #6192, opened 2018-12-18
Prediction: implement keep lane evaluator for cyclist

kechxu:


modules/prediction/evaluator/cyclist/cyclist_keep_lane_evaluator.cc
```diff
@@ -0,0 +1,73 @@
...
+  LaneGraph* lane_graph_ptr =
+      latest_feature_ptr->mutable_lane()->mutable_lane_graph();
+  CHECK_NOTNULL(lane_graph_ptr);
+  if (lane_graph_ptr->lane_sequence_size() == 0) {
+    AERROR << "Obstacle [" << id << "] has no lane sequences.";
+    return;
+  }
+
+  std::string curr_lane_id =
+      latest_feature_ptr->lane().lane_feature().lane_id();
```
panjiacheng: In the current implementation, in some very occasional cases, it might be possible that lane_feature()'s lane_id is not used to construct lane_graph. This might result in all lanes having 0.0 probability. I don't know if it's better that, when this happens, we assign 1.0 to some arbitrary lane_sequence.

kechxu: In that case, the obstacle will not be regarded as "OnLane", thus it won't have lane feature. In obstacle.cc Line 961 and Line 969, lane id is guaranteed.

PR #6197, opened 2018-12-18
Monitor: Support multi-parts process keyword to avoid miss-hit.

xiaoxq:


xiaoxq:
This change will fix the problem that the recorder is shown as running when we are actually running "cyber_recorder play ...".

PR #6215, opened 2018-12-19
Planning: A few refactors

panjiacheng:
- Replaced all the obstacle analyzing logics with a uniformed one to resolve any inconsistency.
- Made the side-pass path generation more robust.


panjiacheng:
Reduced the failed cases from 124 to 113.

modules/planning/common/obstacle_blocking_analyzer.cc
```diff
@@ -35,14 +35,18 @@ using apollo::hdmap::HDMapUtil;
 // TODO(all): if possible, transform as many function parameters into GFLAGS.
 bool IsBlockingObstacleToSidePass(
     const Frame& frame, const Obstacle* obstacle,
-    double block_obstacle_min_speed, double min_front_sidepass_distance,
-    bool enable_obstacle_blocked_check) {
+    double block_obstacle_min_speed,
+    double min_front_sidepass_distance,
+    bool enable_obstacle_blocked_check,
+    double* const distance_between_adc_and_obstacle) {
```
lianglia-apollo: It is better to re-design the function to make it do one thing at a time. Getting the distance and checking blocking seem to be two tasks. Can they be separated into two functions?

modules/planning/scenarios/side_pass/side_pass_scenario.cc
```diff
@@ -193,138 +203,46 @@ bool SidePassScenario::IsFarFromIntersection(const Frame& frame) {
...
-        }
+    double distance_between_adc_and_obstacle = 0.0;
+    if (IsBlockingObstacleToSidePass(
+            frame, obstacle,
+            side_pass_context_.scenario_config_.block_obstacle_min_speed(),
+            side_pass_context_.scenario_config_.min_front_obstacle_distance(),
+            side_pass_context_.scenario_config_.enable_obstacle_blocked_check(),
+            &distance_between_adc_and_obstacle)) {
+      exists_a_blocking_obstacle = true;
+      if (distance_to_closest_blocking_obstacle < 0 ||
```
lianglia-apollo: Use 0.0 when comparing doubles.

modules/planning/scenarios/side_pass/side_pass_scenario.cc
```diff
@@ -193,138 +203,46 @@ bool SidePassScenario::IsFarFromIntersection(const Frame& frame) {
...
+          distance_between_adc_and_obstacle <
+              distance_to_closest_blocking_obstacle) {
+        // Only record the closest front blocking obstacle.
+        distance_to_closest_blocking_obstacle =
+            distance_between_adc_and_obstacle;
+        front_blocking_obstacle_id_ = obstacle->Id() + "_0";
```
lianglia-apollo: Modifying the obstacle id is dangerous as multiple places are using id to match them.

modules/planning/scenarios/side_pass/stage_approach_obstacle.cc
```diff
@@ -113,103 +118,62 @@ Stage::StageStatus StageApproachObstacle::Process(
...
     }
   }
-
-  if ((front_obstacle_distance) < 0) {
+  if (!exists_a_blocking_obstacle) {
+    next_stage_ = ScenarioConfig::NO_STAGE;
+    ADEBUG << "There is no blocking obstacle.";
+    return Stage::FINISHED;
+  }
+  if ((distance_to_closest_blocking_obstacle) < 0) {
```
lianglia-apollo: Extra braces

modules/planning/toolkits/deciders/side_pass_path_decider.cc
```diff
@@ -128,6 +128,25 @@ bool SidePassPathDecider::DecideSidePassDirection(
   }
 }
 
+bool SidePassPathDecider::GetLaneInfoFromPoint(
+    double point_x, double point_y, double point_z, double point_theta,
```
lianglia-apollo: This seems to be a functon in hdmap class.

modules/planning/scenarios/side_pass/side_pass_scenario.cc
panjiacheng: Before this refactor, the obstacle id was modified like this. I wasn't quite sure why obstacle's id should be appended by a "_0"; therefore, I retained this convention.

modules/planning/toolkits/deciders/side_pass_path_decider.cc
panjiacheng: Yes, key to this function is the GetNearestLaneWithHeading function from hdmap. The reason why I wrap it with an extra function here is because: side_pass_path_decider called this multiple times and I wanted to make sure that every time it is called, the other parameters such as the search angle range, etc. are always consistent.

modules/planning/common/obstacle_blocking_analyzer.cc
panjiacheng: Getting the distance seems more like a "side" task because it is only one-line of code. We can, of course, separate them into two. But I'm a little worried of the consistency problem: there can be multiple different definitions of the distance between adc and obstacle: the spacing between centers, the spacing between edges, etc.

PR #6230, opened 2018-12-20
Planning: OpenSpace: fix reverse gear logic

JasonZhou404:


Capri2014:
@JasonZhou404  Please add a  TODO to refactor later and deprecate "traj_point_moving_direction" since this can be represented with velocity sign and other information.

PR #6235, opened 2018-12-20
submitted the conf file, launch file and the main function for V2X module. 

Yangjialin01:


Martin20:
How can I use V2X for communicating between two cars?

PR #6236, opened 2018-12-20
perception: add v2x

ghdawn:


Martin20:
How can I use v2x module to have communication between two cars?

wutianze:
Same question, how to communicate with different host

PR #6247, opened 2018-12-20
Control: reverse gear conf for lat controller

JasonZhou404:


Capri2014:
@JasonZhou404  Quite a few conf needs to be updated as well, including the one for navigation and the ones in testdata

PR #6274, opened 2018-12-22
Perception: Copy lidar parameters into the correct path

techoe:


techoe:
Added one more copy to a perception directory since fusion is using that directory.

Issue #6280, opened 2018-12-24
How to enable velodyne HDL32E for rc3.5?


simon750324:
How do I enable velodyne HDL32E?
I find there are only related files for velodyne128 or velodyne16, even velodyne64 . However, There is no related launch, dag, conf file for velodyne32.

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 14.04)**: 14.04
- **Apollo installed from (source or binary)**: source
- **Apollo version (1.0, 1.5, 2.0, 2.5, 3.0)**: rc3.5 branch

mickeyouyou:
Right now, you can refer to the other model as a reference, put follow content in the file `modules/drivers/velodyne/conf/velodyne32_conf.pb.txt`, and use this link as a module config in dag It's ok .
```
frame_id: "velodyne32"
scan_channel: "/apollo/sensor/velodyne32/VelodyneScan"
rpm: 600.0
model: 32E
mode: STRONGEST
prefix_angle: 18000
firing_data_port: 2368
use_sensor_sync: false
max_range: 130.0
min_range: 0.4 
use_gps_time: true
calibration_online: true
calibration_file: "/apollo/modules/drivers/velodyne/params/HDL32E_calibration_example.yaml"
organized: false
convert_channel_name: "/apollo/sensor/velodyne32/PointCloud2"
use_poll_sync: true
is_main_frame: true
```

simon750324:
Hi # @mickeyouyou 

Thanks, it works.

PR #6291, opened 2018-12-25
Drivers:Create velodyne32_conf.pb.txt

mickeyouyou:
Add configration file for launch the device velodyne 32E


mickeyouyou:
Updated by #6280

PR #6314, opened 2018-12-28
Localization: Modified imu msg source to FilteredImu in predictor output for LMD

Michael-Liu516:
*Enabled FilteredIMU in predictor_output


Michael-Liu516:
Merged in PR6312

PR #6405, opened 2019-01-05
 Planning: OpenSpace: move more params into configs

JasonZhou404:


modules/planning/open_space/trajectory_partition/trajectory_partitioner.cc
```diff
@@ -323,8 +338,12 @@ void TrajectoryPartitioner::GenerateGearShiftTrajectory(
   trajectory_pb->clear_trajectory_point();
 
   // TODO(QiL): move this to config after finalize the logic
-  constexpr double max_t = 3.0;
-  constexpr double unit_t = 0.02;
+  double gear_shift_max_t =
```
deanyanqing: Better to keep the const qualifier

PR #6454, opened 2019-01-09
CANBus: add the WEY vehicle

macDure:
We add a new car named WEY. The car used acceleration control command instead of throttle/brake pedal, so we do a flag to switch the acceleration cmd or throttle/brake pedal cmd, default is throttle/brake pedal cmd. Also the teleop has been modified. The code has checked, tested, and linted.

### DIFF
--- a/modules/canbus/vehicle/wey/protocol/ads1_111.h
+++ b/modules/canbus/vehicle/wey/protocol/ads1_111.h
@@ -131,9 +131,9 @@ class Ads1111 : public ::apollo::drivers::canbus::ProtocolData<
  private:
   Ads1_111::Ads_dectostopType ads_dectostop_;
   Ads1_111::Ads_modeType ads_mode_;
-  double ads_taracce_;
+  double ads_taracce_ = 0.0;
   Ads1_111::Ads_driveoff_reqType ads_driveoff_req_;
-  double ads_aeb_taracce_;
+  double ads_aeb_taracce_ = 0.0;
   Ads1_111::Ads_aeb_tgtdecel_reqType ads_aeb_tgtdecel_req_;
 };
 
--- a/modules/canbus/vehicle/wey/protocol/ads1_111.h
+++ b/modules/canbus/vehicle/wey/protocol/ads1_111.h
@@ -1,0 +1,143 @@
+/******************************************************************************
+ * Copyright 2019 The Apollo Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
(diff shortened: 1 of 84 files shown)


modules/control/controller/lon_controller.cc
```diff
@@ -255,24 +255,35 @@ Status LonController::ComputeControlCommand(
...
+      brake_cmd = std::abs(calibration_value) > brake_deadzone
                     ? std::abs(calibration_value)
                     : brake_deadzone;
+    }
+      cmd->set_throttle(throttle_cmd);
+      cmd->set_brake(brake_cmd);
+  } else {
+      cmd->set_acceleration(acceleration_cmd);
+    if (acceleration_cmd >= 0) {
+      cmd->set_throttle(acceleration_cmd);
```
Capri2014: why you set acceleration_cmd to throttle or brake here, there is reversed field for acceleration command already   optional double acceleration = 10;

Capri2014: https://github.com/ApolloAuto/apollo/blob/443318b4e123f2b87f7676828d4dd386fc6cece7/modules/control/proto/control_cmd.proto#L45

macDure: > why you set acceleration_cmd to throttle or brake here, there is reserved field for acceleration command already optional double acceleration = 10;

Because there is a frame in canbus modules: vehicle_control.cc has defined the throttle and brake to implement cmd from control modules, so we do not change  the original structure.
[/modules/canbus/vehicle/vehicle_controller.cc](url)
I set the acceleration_cmd to throttle/brake just for keep original canbus （vehicle_controller.cc） structure.

macDure: original canbus structure as follow：
[https://github.com/ApolloAuto/apollo/blob/master/modules/canbus/vehicle/vehicle_controller.cc](url)
`if (driving_mode_ == Chassis::COMPLETE_AUTO_DRIVE ||
      driving_mode_ == Chassis::AUTO_SPEED_ONLY) {
    Gear(control_command.gear_location());
    Throttle(control_command.throttle());
    Brake(control_command.brake());
    SetEpbBreak(control_command);
    SetLimits();
  }`

modules/canbus/vehicle/wey/protocol/ads1_111.h
```diff
@@ -0,0 +1,143 @@
...
+  // 'ADS_AEB_TgtDecel_Req', 'is_signed_var': False, 'offset': 0.0,
+  // 'physical_range': '[0|1]', 'bit': 31, 'type': 'enum', 'order': 'motorola',
+  // 'physical_unit': ''}
+  void set_p_ads_aeb_tgtdecel_req(uint8_t* data,
+                      Ads1_111::Ads_aeb_tgtdecel_reqType ads_aeb_tgtdecel_req);
+
+ private:
+  Ads1_111::Ads_dectostopType ads_dectostop_;
+  Ads1_111::Ads_modeType ads_mode_;
+  double ads_taracce_;
```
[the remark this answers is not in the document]
macDure: Hi, this ads1_111.h file is auto-generated by Apollo tools <gen.py>.[https://github.com/macDure/apollo/blob/master/modules/tools/gen_vehicle_protocol/gen.py](url)
 so we do not change the original style, the similar style is also in ads_eps_113.h and so on. [https://github.com/macDure/apollo/blob/master/modules/canbus/vehicle/wey/protocol/ads_eps_113.h](url) , should we change it in all these files?

macDure:
> we should not be using "use_acceleration" based on the discussion.

yes, I have rebase my commit to change it, please check it.

PR #6462, opened 2019-01-10
 Prediction: offline pipeline ListDataForLearning

HongyiSun:


modules/prediction/common/prediction_system_gflags.cc
```diff
@@ -32,6 +32,8 @@ DEFINE_string(prediction_data_dir,
               "Prefix of files to store feature data");
 DEFINE_string(offline_feature_proto_file_name, "",
               "The bin file including a series of feature proto messages");
+DEFINE_string(extract_feature_type, "",
```
panjiacheng: I don't know if it's better to use boolean for such 2-choose-1 stuff, rather than use string.

modules/prediction/proto/offline_features.proto
```diff
@@ -21,10 +21,12 @@ message DataForLearning {
...
   // The features for learning algorithms:
   repeated double features_for_learning = 3;
-
   // The ground-truth labels:
   repeated double labels = 4;
 }
+
+message ListDataForLearning {
+  repeated DataForLearning data_for_learning = 1;
+}
```
YajiaZhang: Add a new empty line after this for compiling with the coding style.

modules/prediction/common/prediction_system_gflags.cc
YajiaZhang: I think it is okay to use a string as we might support more feature type in future.

modules/prediction/feature_proto_file_to_model_features.cc
```diff
@@ -75,19 +77,76 @@ void OfflineProcessFeatureProtoFile(
...
+    const std::string file_name = features_proto_file_name + ".bin";
+    apollo::common::util::SetProtoToBinaryFile(
+        list_data_for_learning, file_name);
   }
 }
 
 int main(int argc, char *argv[]) {
   google::ParseCommandLineFlags(&argc, &argv, true);
   std::string file_name = FLAGS_offline_feature_proto_file_name;
+  std::string feature_type = FLAGS_extract_feature_type;
```
YajiaZhang: In future, pass the file name and feature type as arguments of the main function, and check whether the user provides enough arguments.

PR #6488, opened 2019-01-15
Prediction: implemented the feature-extraction part of lane-scanning evaluator.

panjiacheng:
Lane-scanning evaluator uses symmetric function to be able to scan all possible lane-sequences (variable input size).


panjiacheng:
The feature-extraction will cost an extra ~5msec.

PR #6503, opened 2019-01-16
framework: fix implicit conversions

zhxt:
error: conversion to 'int' from 'chtype {aka long unsigned int}' may
alter its value [-Werror=conversion]
       int ch = mvinch(lineNo + highlight_direction_, x);
                ^

### DIFF
--- a/cyber/tools/cyber_monitor/screen.cc
+++ b/cyber/tools/cyber_monitor/screen.cc
@@ -164,7 +164,7 @@ void Screen::HighlightLine(int lineNo) {
   if (IsInit() && lineNo < Height()) {
     SetCurrentColor(WHITE_BLACK);
     for (int x = 0; x < Width(); ++x) {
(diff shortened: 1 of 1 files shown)


cyber/tools/cyber_monitor/screen.cc
```diff
@@ -164,15 +164,15 @@ void Screen::HighlightLine(int lineNo) {
   if (IsInit() && lineNo < Height()) {
     SetCurrentColor(WHITE_BLACK);
     for (int x = 0; x < Width(); ++x) {
-      int ch = mvinch(lineNo + highlight_direction_, x);
+      chtype = mvinch(lineNo + highlight_direction_, x);
```
jinghaomiao: no variable name?

zhxt: sorry，missed during copy-paste，will add it later .

PR #6505, opened 2019-01-16
Prediction: unify FeatureOutput for all evaluator

HongyiSun:


modules/prediction/prediction_component.cc
```diff
@@ -257,9 +276,6 @@ void PredictionComponent::OnPerception(
         << diff.count() * 1000 << " msec.";
 
   // Make predictions
-  if (FLAGS_prediction_offline_mode) {
```
panjiacheng: Why delete this part?

PR #6553, opened 2019-01-18
Planning: OpenSpace: auto param tuning by rand sampling

JasonZhou404:


modules/tools/open_space_visualization/auto_param_tuning.py
```diff
@@ -0,0 +1,118 @@
...
+
+    tmp = sorted(tmp)
+    optimal_params = params_lists[tmp[0][1]]
+    optimal_evaluation = tmp[0][0]
+    return optimal_params, optimal_evaluation
+
+
+if __name__ == '__main__':
+    random.seed(666)
+    original_file_path = "/apollo/modules/planning/conf/planner_open_space_config.pb.txt"
```
Capri2014: Expand to take the original file path and destination file path as input argument as well for better integration in future.

JasonZhou404: Done

PR #6592, opened 2019-01-22
RTK_player: add flag enabling cnter translation to COM

JasonZhou404:


modules/tools/record_play/rtk_player.py
```diff
@@ -223,6 +223,13 @@ def publish_planningmsg(self):
             adc_point.path_point.theta = self.data['theta'][i]
             adc_point.path_point.s = self.data['s'][i]
 
+            if CHANGE_TO_COM:
```
Capri2014: load values from config to be adaptive to different vehicles. https://github.com/ApolloAuto/apollo/blob/master/modules/common/data/vehicle_param.pb.txt

PR #6593, opened 2019-01-22
Prediction: move callback functions outside prediction component for more flexible usage

kechxu:


modules/prediction/common/message_process.h
```diff
@@ -0,0 +1,46 @@
...
+ public:
+  static void OnPerception(
+      const perception::PerceptionObstacles &perception_obstacles,
+      PredictionObstacles* const prediction_obstacles);
+
+  static void OnLocalization(
+      const localization::LocalizationEstimate &localization);
+
+  static void OnPlanning(const planning::ADCTrajectory &adc_trajectory);
+};
```
YajiaZhang: Add "MessageProces() = delete;" if all functions are static.

kechxu: Done.

PR #6609, opened 2019-01-23
planning: update adolc install shell.

deidaraho:


wanglei828:
I think the generated adolc.so file still cannot enble openmp. Let's wait for @Capri2014 confirmation.

deidaraho:
> I think the generated adolc.so file still cannot enble openmp. Let's wait for @Capri2014 confirmation.

No, this shell only fixed the ColPack installation issue.

PR #6617, opened 2019-01-24
planning: renamed std_planning to on_lane_planning

YajiaZhang:


YajiaZhang:
Renamed a class name to be more meaningful.

PR #6718, opened 2019-02-01
planning: open space hand-derived gradient for cost functino with test case

deidaraho:


modules/planning/common/planning_gflags.cc
```diff
@@ -492,3 +492,5 @@ DEFINE_bool(enable_parallel_open_space_smoother, false,
             "True to enable open space smoother parallel implementation.");
 DEFINE_bool(enable_derivative_check, false,
             "True to enable derivative check inside open space planner.");
+DEFINE_bool(enable_hand_derivative, false,
```
Capri2014: @deidaraho Please consider moving those planner specific flags to conf in the future.

modules/planning/open_space/trajectory_smoother/distance_approach_ipopt_interface_test.cc
```diff
@@ -169,5 +169,27 @@ TEST_F(DistanceApproachIPOPTInterfaceTest, eval_jac_g_par) {
         "values differ at index " << i;
   }
 }
+
+TEST_F(DistanceApproachIPOPTInterfaceTest, eval_grad_f_hand) {
+  int n = 1274;
+  double x[1274];
+  std::fill_n(x, n, 1.2);
+  int new_x = 0;
```
Capri2014: this new_x is a boolean type

modules/planning/open_space/trajectory_smoother/distance_approach_ipopt_interface_test.cc
```diff
@@ -169,5 +169,27 @@ TEST_F(DistanceApproachIPOPTInterfaceTest, eval_jac_g_par) {
...
+  std::fill_n(grad_f_adolc, n, 0.0);
+  std::fill_n(grad_f_hand, n, 0.0);
+
+  bool res = ptop_->eval_grad_f(n, x, new_x, grad_f_adolc);
+  EXPECT_TRUE(res);
+  bool res2 = ptop_->eval_grad_f_hand(n, x, new_x, grad_f_hand);
+  EXPECT_TRUE(res2);
+
+  for (int i = 0; i < n; ++i) {
+    EXPECT_EQ(grad_f_adolc[i], grad_f_hand[i]) <<
```
Capri2014: Also save a copy of result to testdata in case later we want to re-formulate the problem and both ADOLC / Hand writing results need to be changed.

modules/planning/common/planning_gflags.cc
deidaraho: they are in planning.conf now.

modules/planning/open_space/trajectory_smoother/distance_approach_ipopt_interface_test.cc
deidaraho: I have fixed this

modules/planning/open_space/trajectory_smoother/distance_approach_ipopt_interface_test.cc
deidaraho: I am trying to use std lib to save the vector into csv file, but the csv is not in the target folder. Do you have any demo code about how to save file inside Apollo?

Issue #6741, opened 2019-02-05
Leveldb header file missing during dev container build


aboarya:
**Describe the bug**
The build_dev.sh script uses a number of install scripts.  The script [install_python_modules.sh](https://github.com/ApolloAuto/apollo/blob/master/docker/build/installers/install_python_modules.sh) fails with:

plyvel/_plyvel.cpp:359:24: fatal error: leveldb/db.h: No such file or directory

It needs to install the package `libleveldb-dev`

**To Reproduce**
1. Remove any old docker images to start fresh
2. run ./build_dev.sh dev.x86_64.dockerfile from docker/build

**Screenshots**

![screenshot from 2019-02-05 13-21-20](https://user-images.githubusercontent.com/2354387/52295619-48f75c00-294a-11e9-8084-71ee4b64ffc0.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 
**Additional context**
I created a [pull request](https://github.com/ApolloAuto/apollo/pull/6742) to fix this issue

natashadsouza:
Closing this issue as it seems to be resolved. Thanks!

PR #6742, opened 2019-02-05
Added the leveldb header dependency to install_python_modules.sh

aboarya:
updated the dev container's install_python_modules.sh script to install the 
`libleveldb-dev` development headers in addition to what's already being installed

created Issue [6741](https://github.com/ApolloAuto/apollo/issues/6741) to track this.


xiaoxq:
Thanks for the heads up!

I think it's because of dirty py27_requirements.txt file. We have already migrated the modules/common/kv_db lib from leveldb to sqlite.
We can simply remove the plyvel in the py27_requirements.txt file. Could you have a try?

PR #6762, opened 2019-02-06
planning: check smoother constraints inside open space for further study

deidaraho:


Capri2014:
@deidaraho Make sure to move algorithm specific gfalgs to configs instead in next iteration.

PR #6765, opened 2019-02-07
planning: fix open space smoother time sampling initial value bug

deidaraho:


deidaraho:
Also I transfer the enable_constraint_check from a flag to a config parameter inside open space smoother.

PR #6779, opened 2019-02-08
planning: 1. removed unused fallback strategy 2. merged s-t related code

YajiaZhang:


modules/planning/open_space_planning.cc
```diff
@@ -585,10 +585,6 @@ void OpenSpacePlanning::FillPlanningPb(const double timestamp,
   ptr_trajectory_pb->mutable_routing_header()->CopyFrom(
       local_view_.routing->header());
 
-  if (FLAGS_use_planning_fallback &&
```
Capri2014: @YajiaZhang We do need certain fall back mechanisms in case planning fails, do we have new ones to substitute this one after this got removed?

YajiaZhang: Yes, in the future, not too long...

PR #6787, opened 2019-02-08
Planning: rename reference_line_decider to lateral_bounds_decider to make for sense

kechxu:
…


YajiaZhang:
Changing the name to PathBoundaryDecider will be more accurate.

PR #6790, opened 2019-02-08
Perception: Show sub-type of object in visualizer

techoe:


modules/perception/camera/tools/offline/visualizer.cc
```diff
@@ -132,6 +132,60 @@ void Visualizer::SetDirectory(const std::string &path) {
...
+    case apollo::perception::base::ObjectType::VEHICLE:
+     return "VEH";
+    default:
+     return "WRNG";
+  }
+  return "WRNG";
+}
+
+std::string Visualizer::sub_type_to_string(
+  const apollo::perception::base::ObjectSubType type) {
```
ycool: protobuf provides functions to convert enum to string directly. i.e. try PerceptionObstacle_SubType_Name(type)

techoe: @ycool, Thank you for the suggestion and good to know it! However, in this PR, I reduced the string of the subtypes to have only 4 letters (e.g. TRAFFICCONE -> CONE, UNKNOWN_UNMOVABLE->UNMO, etc). If PerceptionObstacle_SubType_Name is used, the string is too long to read on the small image.

PR #6808, opened 2019-02-12
Common: add vehicle model for trajectory stitcher

JasonZhou404:
Test would fail now as many init point has a nonzero speed.
Plz have a review before I make any change to testdata.

### DIFF
--- a/modules/common/configs/config_gflags.cc
+++ b/modules/common/configs/config_gflags.cc
@@ -38,6 +38,11 @@ DEFINE_string(vehicle_config_path,
               "/apollo/modules/common/data/vehicle_param.pb.txt",
               "the file path of vehicle config file");
 
+DEFINE_string(
+    vehicle_model_config_filename,
+    "/apollo/modules/common/vehicle_model/conf/vehicle_model_config.pb.txt",
+    "the file path of vehicle model config file");
+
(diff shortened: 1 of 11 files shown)


Capri2014:
@JasonZhou404 @ycool @YajiaZhang We may want  to merge VehicleState and VehicleModel since they share similar functionalities.

Capri2014:
Also suggest add "Preset" interface to pre-set the initial vehicle model state, and also "Reset" interface. This should be help resolve the failed cases problems mentioned above.

YajiaZhang:
@Capri2014 Actually VehicleModel implemented here is part of VehicleState. It is fine to be integrated in the VehicleState class. 
For the failure cases, the past implementation is wrong and we should not resolve the failure cases due to a wrong model other than updating it.

YajiaZhang:
@JasonZhou404 I will merge the pull request and the modifications can be implemented in future commits.

JasonZhou404:
Maybe wait unitl I make change to the testdata tomorrow?

Yajia Zhang <notifications@github.com>于2019年2月11日 周一21:25写道：

JasonZhou404:
Oh, ok, it is not master anyway

Jinyun Zhou <zhoujiny404@gmail.com>于2019年2月11日 周一21:26写道：

PR #6828, opened 2019-02-13
planning: add scenario dispatcher in ScenarioManager

jmtao:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -222,7 +229,42 @@ void ScenarioManager::Observe(const Frame& frame) {
...
+  if (SelectChangeLaneScenario(ego_point, frame)) {
+    ADEBUG << "Use change lane scenario (temporarily use LANE_FOLLOW)";
+    return;
+  }
+
+  // TODO(all): to be added
+}
+
+void ScenarioManager::Scenario_Voter(
+    const common::TrajectoryPoint& ego_point,
```
YajiaZhang: It seems the naming is a bit misleading. It is more like a class name. Change the name to VoteScenario? The same applies to "Scenario_Dispatcher".

jmtao: addressed. thx

PR #6830, opened 2019-02-13
Planning: implemented skeleton and a few functions of path-bound decider.

panjiacheng:


modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -16,19 +16,174 @@
...
 
+using apollo::common::ErrorCode;
 using apollo::common::Status;
+using apollo::common::VehicleConfigHelper;
+
+constexpr double kPathBoundsDeciderHorizon = 100.0;
+constexpr double kPathBoundsDeciderResolution = 0.1;
+constexpr double kDefaultLaneWidth = 5.0;
+constexpr double kDefaultRoadWidth = 20.0;
+constexpr double kRoadEdgeBuffer = 0.2;
```
YajiaZhang: Move the parameters to gflags. Some of the parameters are already in gflags. It is better to merge them other than defining new ones.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -16,19 +16,174 @@
...
+  // TODO(all): update frame with path bound info.
   return Status::OK();
 }
 
+bool PathBoundsDecider::InitPathBoundaries(
+    const ReferenceLine& reference_line,
+    const common::TrajectoryPoint& planning_start_point,
+    std::vector<std::tuple<double, double, double>>* const path_boundaries) {
+  // Sanity checks.
+  CHECK_NOTNULL(path_boundaries);
```
YajiaZhang: In which case this function will return false? If not, it is cleaner to return the path_boundaries directly other than passing it as a parameter.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -16,19 +16,174 @@
...
+  for (double curr_s = adc_frenet_position.s();
+       curr_s < std::min(adc_frenet_position.s() + kPathBoundsDeciderHorizon,
+                         reference_line.Length());
+       curr_s += kPathBoundsDeciderResolution) {
+    std::tuple<double, double, double> curr_bound = std::make_tuple(
+        curr_s,
+        std::numeric_limits<double>::lowest(),
+        std::numeric_limits<double>::max());
+    path_boundaries->emplace_back(curr_bound);
+  }
```
YajiaZhang: emplace_back works as emplace_back(arg0, arg1, ...). The benefit is initializing the object directly in the container. The usage here is no difference than push_back.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -16,19 +16,174 @@
...
+
+  return true;
+}
+
+bool PathBoundsDecider::GetBoundariesFromRoadsAndADC(
+    const ReferenceLine& reference_line,
```
YajiaZhang: Return value issue. Same here.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -16,19 +16,174 @@
...
+
+bool PathBoundsDecider::GetBoundariesFromStaticObstacles(
+    const IndexedList<std::string, Obstacle>& indexed_obstacles,
+    std::vector<std::tuple<double, double, double>>* const path_boundaries) {
+
+  return true;
+}
+
+double PathBoundsDecider::GetBufferBetweenAdcCeterAndEdge() {
+  double adc_half_width =
```
YajiaZhang: Center? Also, "Adc" to "ADC"

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -16,19 +16,174 @@
...
 Status PathBoundsDecider::Process(
     Frame *const frame, ReferenceLineInfo *const reference_line_info) {
-  // TODO(all) implement
+  // Sanity checks.
+  CHECK_NOTNULL(frame);
+  CHECK_NOTNULL(reference_line_info);
+
+  // The decided path bounds should be in the format of: (s, l_min, l_max).
+  std::vector<std::tuple<double, double, double>> path_boundaries;
+
```
YajiaZhang: The s value can be easily obtained by using index * s_resolution. You can use a pair instead of a tuple.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: Yes, I intended to do that once I finish the entire implementation. That way we know what old gflags can be deleted. I just didn't want to add too many gflags without deleting any.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: corrected.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: corrected.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: The reason why tuple is used is in case later, when considering side-pass, etc., the sequence may be disturbed. If we keep a record of "s", then we always have a chance to sort.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: Sometimes, when there is some failure inside the function, we may want to return false. For example, line 132 and line 144 currently output warning msg and use some temporary remedy, but later we may change them to "return false".

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: Same reason as below.

modules/planning/tasks/deciders/path_bounds_decider.cc
YajiaZhang: OK.

PR #6831, opened 2019-02-13
planning: added empty speed bounds decider.

yifeijiang:


modules/planning/tasks/deciders/speed_bounds_decider.h
```diff
@@ -0,0 +1,40 @@
...
+namespace planning {
+
+class SpeedBoundsDecider : public Decider {
+ public:
+  explicit SpeedBoundsDecider(const TaskConfig& config);
+
+ private:
+  apollo::common::Status Process(
+      Frame* frame, ReferenceLineInfo* reference_line_info) override;
+};
```
YajiaZhang: Add "const" to make the function declaration and implementation consistent.

PR #6860, opened 2019-02-14
add driver to support new GPS

muleisheng:


xiaoxq:
Plz run "./apollo.sh build & ./apollo.sh lint" to check errors.

PR #6863, opened 2019-02-14
Planning: add soft_speed_limit in dp_st cost calc

JasonZhou404:


modules/planning/tasks/optimizers/dp_st_speed/dp_st_cost.cc
```diff
@@ -187,6 +188,15 @@ double DpStCost::GetSpeedCost(const STPoint& first, const STPoint& second,
     cost += config_.low_speed_penalty() * config_.default_speed_cost() *
             -det_speed * unit_t_;
   }
+
+  double soft_det_speed = (speed - soft_speed_limit) / soft_speed_limit;
+  if (soft_det_speed > 0) {
+    cost += config_.exceed_soft_speed_penalty() * config_.default_speed_cost() *
```
YajiaZhang: Change to 0.0 to avoid implicit type casting. You can change it in next commit.

Issue #6869, opened 2019-02-14
Canbus module: the can_sender message should be updated within the PowerSendThreadFunc, not OnControlCommand


liangyd:
**Describe the bug**
Apollo's canbus module updates the can_sender when it receives control data. In this case, it fails to handle the use case when a CAN Tx message has an alive counter and its message trasmit rate is lower than control data update rate. This will probably cause a CAN bus communication issue for certain type of vehicles.

I am using canbus module to send a can message whose last data byte needs to increment every time when it is sent out successfully. In this case, the expected CAN output sequence is:
canid: 123 candata: 00 00 00 00 00 00 00 F1
canid: 123 candata: 00 00 00 00 00 00 00 F2
canid: 123 candata: 00 00 00 00 00 00 00 F3
canid: 123 candata: 00 00 00 00 00 00 00 F4
...

However, the output is:
canid: 123 candata: 00 00 00 00 00 00 00 F1
canid: 123 candata: 00 00 00 00 00 00 00 F3
...
And the discontinued alivecounter triggers an error in the vehicle. 

Then, I noticed that the the can_sender.update() is called by OnControlCommand which runs at a faster rate than my target can message rate.

If I comment out the can_sender.update() in the OnControlCommand() function and add message.Update() inside the for loop within the PowerSendThread, the issue will be resolved..

**To Reproduce**
1. write a new can message inside the protocol folder which has an incremented alive counter. Every time it is called, the data is incremented.  
2. run the canbus module separately, or run the full-stack. 
3. log the can message output.
4. change the can message's period, and log the can messages again.
5. When the can message's period is low and the control_command's update rate is higher, I will see the discontinued alivecounter. 

**Expected behavior**
When the can tx message's period is low and the control_command's update rate is higher, I will see the discontinued alivecounter in the can output logfile. 

**Screenshots**

**Desktop (please complete the following information):**
ubuntu 16.04

**Smartphone (please complete the following information):**
none

**Additional context**

Capri2014:
@liangyd Please update your message rate accordingly, see https://github.com/ApolloAuto/apollo/blob/f2c7896c43616bf4010e028328e7c0cc39598b23/modules/canbus/vehicle/lincoln/protocol/gear_66.cc#L32 for the reference.

Capri2014:
@liangyd If your canbus rate is lower than the control rate, a better choice might be to lower your control rate instead.

liangyd:
[...]

PR #6875, opened 2019-02-15
move the can_sender message update into PowerSendThreadFunc()

liangyd:
Please see issue:  #6869 
Thank you.


jinghaomiao:
Did you get a chance to try it on board? Or at least write a unit test to test your changes? Thank you.

liangyd:
> Did you get a chance to try it on board? Or at least write a unit test to test your changes? Thank you.

Yes. I tried it on board. It was actually tested on a different type of vehicle and I implemented a new list of protocols. The goal is to resolve the issue #6869  that I found out during the test. 

I will try to write a simple unit test when I have time. thank you :-)

PR #6882, opened 2019-02-15
planning: switch to scenario central dispatcher

jmtao:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -418,9 +419,12 @@ void ScenarioManager::ScenarioDispatch(
     scenario_type = SelectSidePassScenario(frame);
   }
 
-  ADEBUG << "select scenario: "
+  ADEBUG << "select cenario: "
```
YajiaZhang: scenario

yifeijiang:
Seems ScenarioConfig::APPROACH is not implemented yet. Or using ScenarioConfig::LANE_FOLLOW?

PR #6883, opened 2019-02-15
Control: merge lead/lag control into the longitudinal controller

kevin-y-wang:


modules/control/proto/control_conf.proto
```diff
@@ -35,6 +35,8 @@ message ControlConf {
   optional bool enable_navigation_mode_position_update = 17 [default = true];
   optional bool enable_persistent_estop = 18 [default = true];
 
+  optional bool enable_reverse_leadlag_compensation = 37 [default = false];
```
Capri2014: why have same configs added in both flags and conf? Try using this one instead and remove the ones in flags

kevin-y-wang: Ok. I will remove the redundant flags

kevin-y-wang: I put the leadlag enable flag into the lon_controller_conf, since the usage scope is for either specific controller. In the future, the newly added flags will be assigned to either control_conf or single controller's conf, depends on the usage scope.

PR #6892, opened 2019-02-15
planning: changed to safer interface for member variable access

YajiaZhang:


modules/planning/common/frame.h
```diff
@@ -118,10 +119,16 @@ class Frame {
       const double planning_start_time,
       prediction::PredictionObstacles *prediction_obstacles);
 
-  ADCTrajectory *mutable_trajectory() { return &trajectory_; }
+  void set_last_planned_trajectory(ADCTrajectory last_planned_trajectory) {
+    last_planned_trajectory_ = std::move(last_planned_trajectory);
+  }
 
-  const ADCTrajectory &trajectory() const { return trajectory_; }
+  ADCTrajectory last_planned_trajectory() const {
```
ycool: s/ADCTrajectory/const ADCTrajectory &/

YajiaZhang: In open space planner, it will change the trajectory_ using mutable_trajectory function. We intend to use this function to substitute the mutable function for safer access of Frame member variables.

PR #6894, opened 2019-02-15
planning: introduce dual warm up debug mode to compare three warm ups in open space

deidaraho:


modules/planning/open_space/trajectory_smoother/dual_variable_warm_start_ipopt_interface.cc
```diff
@@ -115,15 +115,15 @@ bool DualVariableWarmStartIPOPTInterface::get_starting_point(
   // 1. lagrange constraint l, obstacles_edges_sum_ * (horizon_+1)
   for (int i = 0; i < horizon_ + 1; ++i) {
     for (int j = 0; j < obstacles_edges_sum_; ++j) {
-      x[l_index] = 0.5;
+      x[l_index] = 0.0;
```
Capri2014: why change those init values? Did you get better performance with these start values?

deidaraho: > why change those init values? Did you get better performance with these start values?

I think so, original point is inside current optimization problem's feasible set.

modules/planning/conf/planner_open_space_config.pb.txt
```diff
@@ -32,6 +32,7 @@ dual_variable_warm_start_config {
     ipopt_recalc_y: "yes"
   }
   qp_format: OSQP
+  min_safety_distance: 0.0
```
Capri2014: Theoretically this should be greater than 0.05 to avoid quantization error. Or we should specifically add a quantization error field so a lot of our parameters should be greater than it. Otherwise it might look good in planning/simulation but fail in vehicle. --- I will approve this as of now and let's take the discussion offline.

deidaraho: > Theoretically this should be greater than 0.05 to avoid quantization error. Or we should specifically add a quantization error field so a lot of our parameters should be greater than it. Otherwise it might look good in planning/simulation but fail in vehicle. --- I will approve this as of now and let's take the discussion offline.

Sure, we will talk about this later.

PR #6915, opened 2019-02-20
Planning: remove code related to front obs distance

zwccharlie:


panjiacheng:
It would be great if this change can be tested by simulation, because I'm a little worried that there may be some bugs in the path_bound_decider.
path_bound_decider is just code-complete, and hasn't been tested in simulation yet.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -84,6 +84,7 @@ Status PathBoundsDecider::Process(
 
   // Update the path boundary info to the frame.
   // TODO(all): update frame with path bound info.
+  reference_line_info->SetReachableS(std::get<0>(path_boundaries.back()));
```
panjiacheng: path_bound_decider hasn't been tested in simulation. If this would be the first time using anything from path_bound_decider, it will be great if simulation can be run before this change takes effect on regular ADC tests everyday.

zwccharlie:
> It would be great if this change can be tested by simulation, because I'm a little worried that there may be some bugs in the path_bound_decider.
> path_bound_decider is just code-complete, and hasn't been tested in simulation yet.

No need to worry. We don't have to merge right now. We can do more tests this week.

modules/planning/tasks/deciders/path_bounds_decider.cc
zwccharlie: > path_bound_decider hasn't been tested in simulation. If this would be the first time using anything from path_bound_decider, it will be great if simulation can be run before this change takes effect on regular ADC tests everyday.

Also need to mention that this bound only used on fallback speed.

YajiaZhang:
I think the solution is on the right track. As @panjiacheng said, the path bound decider has not been tested and really used in the code. Let's hold the pull request for some time and do some testing first @zwccharlie  But I'm really glad to see the front_obs_distance stuff will soon be gone :-)

jinghaomiao:
@zwccharlie  Good job. Please also resolve the conflicts after passing all necessary tests.

zwccharlie:
Will resolve conflicts once tests done by Jiacheng. Otherwise some commits will still cause conflicts.

PR #6917, opened 2019-02-20
Control: merge lead/lag control into the lateral controller

kevin-y-wang:


modules/control/proto/lat_controller_conf.proto
```diff
@@ -25,4 +26,6 @@ message LatControllerConf {
   optional double max_lateral_acceleration = 15;  // limit aggressive steering
   optional apollo.control.GainScheduler lat_err_gain_scheduler = 16;
   optional apollo.control.GainScheduler heading_err_gain_scheduler = 17;
+  optional LeadlagConf reverse_leadlag_conf = 18;
```
Capri2014: The corresponding control_conf.pb.txt changes?

kevin-y-wang: Yes, the control_conf.pb.txt will change accordingly so as to reflect the new configurations. I did not include the pb.txt into this commit only because the variable values are not fixed. Do you think we should anyway add the pb.txt into this commit and may adjust the variable values later? Thanks.

Capri2014: @kevin-y-wang Yes, please add the field in conf file with default values set so the overall gain is 1 (turned off).

kevin-y-wang: Get it.

jinghaomiao:
LGTM, @Capri2014 Please take a second look.

PR #6928, opened 2019-02-21
Prediction: tmp disable building lane graph for ego vehicle

kechxu:


YajiaZhang:
What is the reason of temp. disabling that? When will we enable that back?

PR #6936, opened 2019-02-21
Prediction: caution level for cruise change lane

kechxu:


modules/prediction/scenario/prioritization/obstacles_prioritizer.cc
```diff
@@ -33,6 +35,51 @@ using common::adapter::AdapterConfig;
...
+    std::string lane_id = lane_segment.lane_id();
+    if (ego_trajectory_container->IsLaneIdInReferenceLine(lane_id)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+int NearestFrontObstacleIdOnLaneSequence(const LaneSequence& lane_sequence) {
+  int nearest_front_obstacle_id = -10;
```
YajiaZhang: Why the default value is -10?

kechxu: Just want to distinguish other obstacles and ego vehicle

YajiaZhang: It is better to use a flag. Hard coded magic number is not recommended.

kechxu: I don't this it is necessary to make a flag only for this case.

kechxu: I used inf instead of the magic number -10. Please checkout the latest commit.

YajiaZhang: Okay.

PR #6947, opened 2019-02-22
Docs: Dreanview features in Apollo 3.5

natashadsouza:


modules/dreamview/README.md
```diff
@@ -15,3 +16,19 @@ Dreamview or Apollo's HMI module provides a web application that helps developer
   
 ## Output
   A web-based dynamic 3D rendering of the monitored messages in a simulated world.
+
+## Features introduced in Apollo 3.5
```
vlin17: Shall we put this content in the usage table? Because there are other features in apollo 3.5, such as the customizable graph, but this feature is presented in FAQ instead.

natashadsouza: Makes sense. Will send the update shortly

PR #6948, opened 2019-02-22
Prediction: fixed a bug in BuildLaneGraphFromLeftToRight.

panjiacheng:


modules/prediction/container/obstacles/obstacle.cc
```diff
@@ -1162,6 +1163,20 @@ void Obstacle::BuildLaneGraphFromLeftToRight() {
     }
   }
 
+  // Remove repeated lane_segments
+  std::set<std::string> existing_lane_ids;
+  std::vector<std::string> lane_ids_ordered_temp = lane_ids_ordered;
```
YajiaZhang: You can use unordered_set for efficiency gain.

PR #6958, opened 2019-02-22
Prediction: acquire more neighbor lanes while constructing lane graphs.

panjiacheng:


modules/prediction/container/obstacles/obstacle.cc
```diff
@@ -1120,6 +1121,65 @@ void Obstacle::SetLaneSequenceStopSign(LaneSequence* lane_sequence_ptr) {
   }
 }
 
+void Obstacle::GetNeighborLaneSegments(
+    std::shared_ptr<const LaneInfo> center_lane_info,
+    bool is_left,
+    int recursion_depth,
+    std::list<std::string>* const lane_ids_ordered,
```
YajiaZhang: Add "const" to parameters that will not be changed.

PR #6976, opened 2019-02-25
Planning: renaming for better semantic understanding

JasonZhou404:


modules/planning/common/frame.h
```diff
@@ -121,11 +121,11 @@ class Frame {
       const double planning_start_time,
       prediction::PredictionObstacles *prediction_obstacles);
 
-  void set_last_planned_trajectory(ADCTrajectory last_planned_trajectory) {
+  void set_current_frame_planned_trajectory(ADCTrajectory last_planned_trajectory) {
```
xiaoxq: Should be either "const ADCTrajectory& last_planned_trajectory" or "ADCTrajectory&& last_planned_trajectory", or else it does several copies.

modules/planning/common/frame.h
```diff
@@ -121,11 +121,11 @@ class Frame {
...
     last_planned_trajectory_ = std::move(last_planned_trajectory);
   }
 
-  ADCTrajectory last_planned_trajectory() const {
+  ADCTrajectory current_frame_planned_trajectory() const {
```
xiaoxq: Return "const ADCTrajectory&". Let the user to decide if they want a copy or a ready only reference.

modules/planning/open_space_planning.cc
```diff
@@ -247,7 +247,7 @@ Status OpenSpacePlanning::Plan(
   Status trajectory_partition_status;
 
   if (status == Status::OK()) {
-    auto trajectory_after_stitching_point = frame_->last_planned_trajectory();
+    auto trajectory_after_stitching_point = frame_->current_frame_planned_trajectory();
```
xiaoxq: For example, this is a copy.

modules/planning/tasks/smoothers/smoother.cc
```diff
@@ -95,7 +95,7 @@ apollo::common::Status Smoother::Smooth(
     AWARN << msg;
     return Status(ErrorCode::PLANNING_ERROR, msg);
   }
-  const auto& previous_planning = previous_frame->last_planned_trajectory();
+  const auto& previous_planning = previous_frame->current_frame_planned_trajectory();
```
xiaoxq: This is also a copy.

modules/planning/common/frame.h
JasonZhou404: Yeah, will fix it.

PR #7000, opened 2019-02-27
Third_party: libtf2.so rebuilt on ubuntu 18.04

fengliu00:
Rebuilt it on Ubuntu 18.04


ycool:
18.04 libtf2.so  can't be merged into master yet for current master still runs in 14.04 env.

PR #7018, opened 2019-02-27
planning: implement open_space_roi_decider based on open_space_ROI

deidaraho:


Capri2014:
please add test for it as well

deidaraho:
> please add test for it as well

We have included an unit test before, which follows 'side_pass_path_decider_test.cc'

PR #7065, opened 2019-03-01
planning: scenario manager: add a map to avoid traversing vector here…

jmtao:
… and there

### DIFF
--- a/modules/planning/scenarios/scenario_manager.cc
+++ b/modules/planning/scenarios/scenario_manager.cc
@@ -294,6 +294,30 @@ bool ScenarioManager::SelectScenario(
 void ScenarioManager::Observe(const Frame& frame) {
   // read traffic light signal info
   ReadTrafficLight(frame);
+
+  // init first_encountered_overlap_map_
+  first_encountered_overlap_map_.clear();
+  const auto& reference_line_info = frame.reference_line_info().front();
+  const auto& first_encountered_overlaps =
+      reference_line_info.FirstEncounteredOverlaps();
+  for (const auto& overlap : first_encountered_overlaps) {
+    switch (overlap.first) {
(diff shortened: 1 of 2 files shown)


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -294,6 +294,30 @@ bool ScenarioManager::SelectScenario(
...
+        first_encountered_overlap_map_[ReferenceLineInfo::STOP_SIGN] =
+            overlap.second;
+        break;
+      case ReferenceLineInfo::SIGNAL:
+        first_encountered_overlap_map_[ReferenceLineInfo::SIGNAL] =
+            overlap.second;
+        break;
+      case ReferenceLineInfo::PNC_JUNCTION:
+        first_encountered_overlap_map_[ReferenceLineInfo::PNC_JUNCTION] =
+            overlap.second;
```
xiaoxq: Simply first_encountered_overlap_map_[overlap.first] = overlap.second; for these three types?

YajiaZhang: I guess there might be some customized assignment coming.

xiaoxq: Then might be like:

    switch (overlap.first) {
      case ReferenceLineInfo::STOP_SIGN:
      case ReferenceLineInfo::SIGNAL:
      case ReferenceLineInfo::PNC_JUNCTION:
          first_encountered_overlap_map_[overlap.first] = overlap.second;
          break;

Or put them to default block, and use cases to handle customized ones :)

YajiaZhang: Sure, I'm fine with either way. Let @jmtao decide.

jmtao: haha. good catch! fixed, amended

PR #7081, opened 2019-03-04
tools: update mapshow to be able to display pnc-junctions

jmtao:


modules/tools/mapshow/libs/map.py
```diff
@@ -207,6 +237,16 @@ def _draw_lane_central(lane, ax, color_val):
...
 
+    @staticmethod
+    def _draw_pnc_boundary(pnc_junction, ax, color_val):
+        """draw boundary"""
+        px = []
+        py = []
+        for point in pnc_junction.polygon.point:
+            px.append(point.x)
+            py.append(point.y)
+	    ax.plot(px, py, ls='-', c=color_val, alpha=0.5)
```
xiaoxq: Indents.

modules/tools/mapshow/libs/map.py
```diff
@@ -147,6 +147,36 @@ def _draw_lane_details(self, lane, ax, color_val):
             arrowprops=dict(arrowstyle='-|>', connectionstyle='arc3,rad=-0.2',
                             fc=color_val, ec=color_val, alpha=0.5))
 
+    def draw_pnc_junctions(self, ax):
+        cnt = 1
```
xiaoxq: cnt = 1 + len(self.map_pb.pnc_junction) ?

modules/tools/mapshow/libs/map.py
jmtao: fixed

modules/tools/mapshow/libs/map.py
jmtao: cnt is a counter, incrementing, used to decide/differentiate label color

PR #7082, opened 2019-03-04
Planning: remove code related to front obs distance

zwccharlie:
Do not merge unless Yajia and JIacheng both approved


YajiaZhang:
+1, hold for now.

PR #7090, opened 2019-03-05
Planning: interface for path point speed decision guide in speed_boun…

JasonZhou404:
…ds_decider


modules/planning/proto/speed_bounds_decider_config.proto
```diff
@@ -17,4 +17,5 @@ message SpeedBoundsDeciderConfig {
   optional double static_obs_nudge_speed_ratio = 12;
   optional double dynamic_obs_nudge_speed_ratio = 13;
   optional double centri_jerk_speed_coeff = 14;
+  optional bool path_point_decision_guide_flag = 15 [default = false];
```
YajiaZhang: Add a new line

PR #7095, opened 2019-03-05
Cyber: Fix the memory issue of cyber_record

KitFung:
The current `cyber_recorder play`  will hold all the loaded message in the memory.
It cause memory issue while playing many bags.

Example:
`cyber_recorder play -f /path/to/data//20190304154029.record.0000*`

This commit is aims at fixing this issue:
 - Clear the data of the RecordReader if that file is already finish playing

This fix isn't the best solution (will still having memory issue while reading a single super large data), but it is a usable solution for general case. (A lot of sequential small data)

The below is the memory statistic before this change
screenshot from 2019-03-05 15-31-07

The below is the memory statistic after this change
screenshot from 2019-03-05 17-52-47

The peak memory have been reduced from 7.x GB to 2.x GB

### DIFF
--- a/cyber/record/record_viewer.cc
+++ b/cyber/record/record_viewer.cc
@@ -28,8 +28,11 @@ namespace record {
 RecordViewer::RecordViewer(const RecordReaderPtr& reader, uint64_t begin_time,
                            uint64_t end_time,
                            const std::set<std::string>& channels)
-    : begin_time_(begin_time), end_time_(end_time), channels_(channels) {
-  readers_.emplace_back(reader);
+    : begin_time_(begin_time),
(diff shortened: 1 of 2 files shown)


cyber/record/record_viewer.cc
```diff
@@ -146,7 +151,19 @@ bool RecordViewer::FillBuffer() {
       this_end_time = end_time_;
     }
 
-    for (auto& reader : readers_) {
+    for (size_t i = 0; i < readers_.size(); ++i) {
+      if (!readers_finished_[i] &&
```
gruminions: thanks for your optimization!

PR #7100, opened 2019-03-05
Prediction: add mask to junction_labels

HongyiSun:


modules/tools/prediction/data_pipelines/common/online_to_offline.py
```diff
@@ -436,6 +436,7 @@ def LabelJunctionExit(self):
                 # Construct dictionary of all exit with dict[exit_lane_id] = np.array(exit_position)
                 exit_dict = dict()
                 exit_pos_dict = dict()
+                mask = [0 for idx in range(12)]
```
xiaoxq: [0] * 12

PR #7108, opened 2019-03-06
Planning: OpenSpace:  separate open_space and on_lane logics in on_lane_planning.cc

JasonZhou404:


Capri2014:
@JasonZhou404 Make sure to rename the on_lane_planning.cc if we intend to use this to host both the on_lane and open_space planning logic.

PR #7113, opened 2019-03-06
Dreamview: Trigger calibration monitor when changing mode.

xiaoxq:


modules/dreamview/backend/hmi/hmi.cc
```diff
@@ -102,6 +102,14 @@ void HMI::RegisterMessageHandlers() {
           // Reload lidar params for point cloud service.
           PointCloudUpdater::LoadLidarHeight(FLAGS_lidar_height_yaml);
           SendVehicleParam();
+        } else if (hmi_action == HMIAction::CHANGE_MODE) {
+          static constexpr char kCalibrationMode[] = "Mkz Calibration";
+          if (value == kCalibrationMode) {
+            // TODO(vivian): Finish.
```
Capri2014: Is this the one for control/dynamic model calibration?

Capri2014: If it is, shall we re-name it to "Vehicle Calibration", this mode would be needed for calibration for other vehicles as well.

xiaoxq: Per offline discussion. Do this in future commits.

PR #7136, opened 2019-03-06
planning: implement open space fall back decider

deidaraho:
The testing cases have been generated before.

### DIFF
--- a/modules/planning/common/open_space_info.h
+++ b/modules/planning/common/open_space_info.h
@@ -192,14 +192,18 @@ class OpenSpaceInfo {
     return &chosen_paritioned_trajectory_;
   }
 
-  bool *mutable_fallback_flag() { return &fallback_flag_; }
-
   const bool &fallback_flag() const { return fallback_flag_; }
+
+  void set_fallback_flag(bool flag) { fallback_flag_ = flag; }
 
   TrajGearPair *mutable_fallback_trajectory() { return &fallback_trajectory_; }
 
   const TrajGearPair &fallback_trajectory() const {
     return fallback_trajectory_;
+  }
+
+  void set_fallback_trajectory(const TrajGearPair& traj_gear_pair) {
+    fallback_trajectory_ = traj_gear_pair;
   }
 
   std::pair<PublishableTrajectory, canbus::Chassis::GearPosition>
--- a/modules/planning/proto/open_space_fallback_decider_config.proto
+++ b/modules/planning/proto/open_space_fallback_decider_config.proto
@@ -2,4 +2,11 @@ syntax = "proto2";
 
 package apollo.planning;
 
-message OpenSpaceFallBackDeciderConfig {}
+message OpenSpaceFallBackDeciderConfig {
+  // prediction time horizon for prediction obstacles
+  optional double open_space_prediction_time_period = 1 [default = 5.0];
+  // fallback collision check distance
+  optional double open_space_fall_back_collision_distance = 2 [default = 5.0];
+  // fallback stop trajectory safety gap to obstacles
+  optional double open_space_fall_back_stop_safety_gap = 3 [default = 1.0];
+}
--- a/modules/planning/tasks/deciders/open_space_fallback_decider.cc
+++ b/modules/planning/tasks/deciders/open_space_fallback_decider.cc
@@ -22,13 +22,136 @@
 
 namespace apollo {
 namespace planning {
-
 using apollo::common::Status;
+using apollo::common::math::Box2d;
+using apollo::common::math::Vec2d;
+using apollo::common::TrajectoryPoint;
+using apollo::planning::TrajGearPair;
 
 OpenSpaceFallbackDecider::OpenSpaceFallbackDecider(const TaskConfig& config)
     : Decider(config) {}
 
-Status OpenSpaceFallbackDecider::Process(Frame* frame) { return Status::OK(); }
+Status OpenSpaceFallbackDecider::Process(Frame* frame) {
+  std::vector<std::vector<common::math::Box2d>> predicted_bounding_rectangles;
+  double collision_distance;
 
+  BuildPredictedEnvironment(frame->obstacles(),
+      predicted_bounding_rectangles);
+
+  if (!IsCollisionFreeTrajectory(
+        frame->open_space_info().chosen_paritioned_trajectory(),
+        predicted_bounding_rectangles, &collision_distance)) {
+    // change gflag
+    frame_->mutable_open_space_info()->set_fallback_flag(true);
+
+    // generate fallback trajectory base on current partition trajectory
+    // vehicle speed is decreased to zero inside safety distance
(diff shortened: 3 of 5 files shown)


modules/planning/common/open_space_info.h
```diff
@@ -192,16 +192,20 @@ class OpenSpaceInfo {
...
 
+  void set_fallback_flag(bool flag) { fallback_flag_ = flag; }
+
   TrajGearPair *mutable_fallback_trajectory() { return &fallback_trajectory_; }
 
   const TrajGearPair &fallback_trajectory() const {
     return fallback_trajectory_;
   }
 
+  void set_fallback_trajectory(TrajGearPair traj_gear_pair) {
```
JasonZhou404: use mutable_fallback_trajectory to avoid useless copying

modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -22,13 +22,125 @@
...
+    BuildPredictedEnvironment(frame->obstacles());
+
+    if (!IsCollisionFreeTrajectory(
+      frame->open_space_info().chosen_paritioned_trajectory())) {
+      // change gflag
+      frame_->mutable_open_space_info()->set_fallback_flag(true);
+
+      // generate fallback trajectory base on current partition trajectory
+      // vehicle speed is decreased to zero inside safety distance
+      frame_->mutable_open_space_info()->set_fallback_trajectory(
```
JasonZhou404: use:
const auto &open_space_info = frame_->_open_space_info()
auto* open_space_info_ptr = frame_->mutable_open_space_info()
to make codes shorter

modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -22,13 +22,125 @@
...
+          frame_->mutable_open_space_info()->
+            mutable_chosen_paritioned_trajectory()->
+            first[i].set_v(next_v);
+          frame_->mutable_open_space_info()->
+            mutable_chosen_paritioned_trajectory()->
+            first[i].set_a(-accelerate);
+          current_v = next_v;
+        }
+      } else {
+        size_t temp_horizon = frame_->open_space_info().
```
Capri2014: In the "else" logic, should you set the fallback_flag back to "false" ?

modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -22,13 +22,125 @@
...
+    while (relative_time <
+        config_.open_space_fallback_decider_config().
+          open_space_prediction_time_horizon()) {
+      std::vector<Box2d> predicted_env;
+      for (const Obstacle* obstacle : obstacles) {
+          TrajectoryPoint point = obstacle->GetPointAtTime(relative_time);
+          Box2d box = obstacle->GetBoundingBox(point);
+          predicted_env.push_back(std::move(box));
+    }
+    predicted_bounding_rectangles_.push_back(std::move(predicted_env));
```
JasonZhou404: use emplace_back  rather than push_back

modules/planning/common/open_space_info.h
xiaoxq: const TrajGearPair& traj_gear_pair

modules/planning/proto/open_space_fallback_decider_config.proto
```diff
@@ -2,4 +2,11 @@ syntax = "proto2";
 
 package apollo.planning;
 
-message OpenSpaceFallBackDeciderConfig {}
\ No newline at end of file
+message OpenSpaceFallBackDeciderConfig {
+  // prediction time horizon for prediction obstacles
+  optional double open_space_prediction_time_horizon = 1 [default = 1.0];
```
Capri2014: This is time (double in seconds) not time horizon (integer No.), please change the name to avoid confusion. Also, prediction time of default 1.0s might be too small.

modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -22,13 +22,125 @@
...
+          double next_v = std::max(0.0, current_v -
+              accelerate * frame_->open_space_info().
+                  chosen_paritioned_trajectory().first.
+                  TrajectoryPointAt(i).relative_time());
+          frame_->mutable_open_space_info()->
+            mutable_chosen_paritioned_trajectory()->
+            first[i].set_v(next_v);
+          frame_->mutable_open_space_info()->
+            mutable_chosen_paritioned_trajectory()->
+            first[i].set_a(-accelerate);
```
xiaoxq: Save frame_->mutable_open_space_info()->mutable_chosen_paritioned_trajectory() to a variable, make code shorter and cleaner.

modules/planning/tasks/deciders/open_space_fallback_decider.h
```diff
@@ -20,18 +20,43 @@
...
+  // bool IsCollisionFreeTrajectory(const ADCTrajectory& trajectory_pb);
+
+  void BuildPredictedEnvironment(const std::vector<const Obstacle*>& obstacles);
+
+  bool IsCollisionFreeTrajectory(const TrajGearPair& trajectory_pb);
+
+ private:
+  std::vector<std::vector<common::math::Box2d>> predicted_bounding_rectangles_;
+
+  double collision_distance_;
```
JasonZhou404: I think It's better pass the collision_distance_  and predicted_bounding_rectangles_ between "check collision"  and "process" rather than put in private data field

modules/planning/tasks/deciders/open_space_fallback_decider.cc
deidaraho: > In the "else" logic, should you set the fallback_flag back to "false" ?

The logic is if the vehicle has enough distance to the obstacle, it decrease and stop following the previous trajectory, otherwise, stop at the current point.

modules/planning/common/open_space_info.h
deidaraho: > const TrajGearPair& traj_gear_pair

fixed

modules/planning/proto/open_space_fallback_decider_config.proto
deidaraho: > This is time (double in seconds) not time horizon (integer No.), please change the name to avoid confusion. Also, prediction time of default 1.0s might be too small.

fixed

modules/planning/tasks/deciders/open_space_fallback_decider.cc
deidaraho: > In the "else" logic, should you set the fallback_flag back to "false" ?

fixed

modules/planning/tasks/deciders/open_space_fallback_decider.h
deidaraho: > I think It's better pass the collision_distance_ and predicted_bounding_rectangles_ between "check collision" and "process" rather than put in private data field
your four comments are fixed

PR #7139, opened 2019-03-06
 Planning: OpenSpace: Constructing valet parking scenarios and stages 

JasonZhou404:


modules/planning/scenarios/valet_parking/BUILD
```diff
@@ -6,44 +6,12 @@ cc_library(
     name = "valet_parking",
     srcs = [
         "valet_parking_scenario.cc",
```
Capri2014: break them into small ones

modules/planning/scenarios/valet_parking/valet_parking_scenario.cc
```diff
@@ -17,21 +17,206 @@
...
+    AERROR << "miss scenario specific config";
+    return false;
+  }
+  context_.scenario_config.CopyFrom(config_.valet_parking_config());
+  return true;
 }
 
 bool ValetParkingScenario::IsTransferable(const Scenario& current_scenario,
                                           const Frame& frame) {
+  // TODO(all) Implement avaliable parking spot detection by preception results
```
Capri2014: You can add a dummy function "IsParkingSpotAvailable"  and always returns true as of now.

modules/planning/scenarios/valet_parking/BUILD
JasonZhou404: Current architect must have them depends on each other

modules/planning/scenarios/valet_parking/valet_parking_scenario.cc
JasonZhou404: Will do that later, if we don't take available parking spot from routing message

PR #7175, opened 2019-03-08
Planning: separate  CUDA and non CUDA distance approach interface in open space planner

Capri2014:


modules/planning/open_space/trajectory_smoother/BUILD
```diff
@@ -98,6 +98,7 @@ cc_library(
     ],
     deps = [
         "distance_approach_ipopt_interface",
+        "distance_approach_ipopt_cuda_interface",
```
xiaoxq: Should add leading ":"

modules/planning/open_space/trajectory_smoother/distance_approach_interface.h
```diff
@@ -0,0 +1,127 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
xiaoxq: 2019, as well as other new files.

Capri2014: done

modules/planning/open_space/trajectory_smoother/BUILD
Capri2014: done

PR #7204, opened 2019-03-11
optimization for arm64(TX2), only update a portion of path in relative map&navigation mode

DongAi:
This feature can improve performance significantly in navigation mode, it can reduce the resource consumption of RelativeMap, Planning, Dreamview. For the sake of safety, it's only available on arm64 now.


modules/map/relative_map/navigation_lane.cc
```diff
@@ -425,8 +425,17 @@ bool NavigationLane::ConvertNavigationLineToPath(const int line_index,
...
       }
+      int path_size = navigation_path.path_point_size();
+      int path_size_ahead = path_size;
+#ifdef __aarch64__
+      path_size_ahead = current_project_index +
+                        FLAGS_relative_map_path_frame_ahead;
+      if (path_size_ahead > path_size) {
+        path_size_ahead = path_size;
+      }
+#endif
```
xiaoxq: Make the logic shorter and cleaner, such as:
```
      int path_size = navigation_path.path_point_size();
#ifdef __aarch64__
      int path_size_ahead = std::min(current_project_index + FLAGS_relative_map_path_frame_ahead, path_size);
#else 
      int path_size_ahead = path_size;
#endif
```

xiaoxq: #ifdef __aarch64__
      const int path_size_ahead = std::min(
          current_project_index + FLAGS_relative_map_path_frame_ahead,
          navigation_path.path_point_size());
#else 
      const int path_size_ahead = navigation_path.path_point_size();
#endif

PR #7211, opened 2019-03-11
Prediction: also AssignCautionLevelByEgoReferenceLine in cruise

HongyiSun:


modules/prediction/scenario/prioritization/obstacles_prioritizer.cc
```diff
@@ -337,13 +338,53 @@ void ObstaclesPrioritizer::AssignCautionLevelByEgoReferenceLine() {
...
   double accumulated_s = 0.0;
+  // first loop of lane_ids to findout ego_vehicle_s
+  for (const std::string& lane_id : lane_ids) {
+    std::shared_ptr<const LaneInfo> lane_info_ptr =
+        PredictionMap::LaneById(lane_id);
+    if (lane_info_ptr == nullptr) {
+      AERROR << "Null lane info pointer found.";
+      continue;
+    }
+    double s, l;
```
kechxu: suggest to use:
double s = 0.0;
double l = 0.0;

modules/prediction/scenario/prioritization/obstacles_prioritizer.cc
```diff
@@ -337,13 +338,53 @@ void ObstaclesPrioritizer::AssignCautionLevelByEgoReferenceLine() {
...
+    PredictionMap::GetProjection({pose_x, pose_y}, lane_info_ptr, &s, &l);
```
kechxu: Check if GetProjection returns true

PR #7244, opened 2019-03-13
add auto lock class

muleisheng:


xiaoxq:
What is the usage?

muleisheng:
> What is the usage?

I have added logic to use this auto lock.

modules/common/util/auto_lock.h
```diff
@@ -0,0 +1,88 @@
...
+void InternalUnLock(T *t);
+
+/**
+ * @brief override these two routines for other synchronization methods.
+ */
+void InternalLock(std::mutex *t) noexcept(false) {
+    if (!t)
+        throw std::errc::bad_address;
+    else
+        t->lock();
```
xiaoxq: 1. Always use {} for sub-scopes.
2. Fix indents to 2 spaces.

modules/common/util/auto_lock.h
```diff
@@ -0,0 +1,88 @@
...
+void InternalLock(T *t);
+template<typename T>
+void InternalUnLock(T *t);
+
+/**
+ * @brief override these two routines for other synchronization methods.
+ */
+void InternalLock(std::mutex *t) noexcept(false) {
+    if (!t)
+        throw std::errc::bad_address;
```
xiaoxq: Avoid "throw" if possible. Simply CHECK.

modules/common/util/auto_lock.h
```diff
@@ -0,0 +1,88 @@
...
+        if (!t_)
+            throw std::errc::bad_address;
+        if (!lock_) {
+            InternalLock(t_);
+            lock_ = true;
+        }
+    }
+
+ private:
+    T *t_;
```
fengqikai1414: Use `T& t_` may be better

modules/common/util/auto_lock.h
muleisheng: OK ， I replace throw by CHECK

modules/common/util/auto_lock_test.cc
```diff
@@ -0,0 +1,44 @@
...
+#include "gtest/gtest.h"
+
+#define MAX_COUNT       10000000
+#define EXPECTED_RESULT 20000000
+
+std::mutex g_mutex;
+int g_int = 0;
+
+void function() {
+    for (int i = 0; i < MAX_COUNT; i++) {
```
xiaoxq: Indents as 2.

modules/common/util/auto_lock.h
```diff
@@ -0,0 +1,83 @@
...
+ public:
+  explicit AutoLock(std::mutex *t) :
+    t_(t),
+    lock_(false) {
+    Lock();
+  }
+
+  ~AutoLock() {
+    UnLock();
+  }
```
xiaoxq: Could you try std::lock_guard? It's pretty similar with your implementation.

muleisheng: auto_lock supplies unlock by manual when we need.

xiaoxq: Personally I prefer a scoped lock_guard in such case. It's pretty straight forward and easy to understand by everybody.

PR #7253, opened 2019-03-13
Prediction: add 15 degree to classification zones

HongyiSun:


modules/tools/prediction/data_pipelines/common/online_to_offline.py
```diff
@@ -449,7 +449,7 @@ def LabelJunctionExit(self):
                         exit_pos_dict[junction_exit.exit_lane_id] = exit_pos
                         delta_pos = exit_pos - curr_pos
                         angle = math.atan2(delta_pos[1], delta_pos[0]) - heading
-                        d_idx = int((angle / (2.0 * np.pi)) * 12 % 12)
+                        d_idx = int((angle / (2.0 * np.pi) + 1.0 / 24) * 12 % 12)
```
xiaoxq: Use static_cast

PR #7272, opened 2019-03-14
planning: update scenario proto

jmtao:


modules/planning/proto/planning_config.proto
```diff
@@ -101,6 +101,13 @@ message ScenarioSidePassConfig {
   optional double stop_fence_distance_to_blocking_obstacle = 10 [default = 6.0];
 }
 
+message ScenarioBareIntersectionConfig {
+  optional double start_bare_intersection_scenario_distance = 1 [default = 40.0];  // meter
+  optional double min_pass_s_distance = 2 [default = 3.0];  // meter
+  optional float stop_timeout = 3 [default = 8.0];  // sec
+  optional float creep_timeout = 4 [default = 10.0];  // sec
```
xiaoxq: Put unit in field name would be more clear for users, such as creep_timeout_sec.

jmtao: that'll be too much code changes.  will do that in another commit later.

PR #7301, opened 2019-03-15
Perception: Save manually calibrated camera parameters

techoe:
Note: Temporarily camera parameters are saved by streaming. Please let me know if we have save_yaml function.


modules/perception/camera/tools/offline/visualizer.cc
```diff
@@ -383,6 +386,244 @@ bool Visualizer::reset_key() {
   return true;
 }
 
+double Visualizer::regularize_angle(const double &radian_angle) {
```
xiaoxq: Simply pass value of primitive types, instead of reference. Plz also fix others.

techoe: I think reference consumes less memory than copying double type.

xiaoxq: It takes as long as a pointer, so generally there is no benefit. Mordern compiler could use register to store the scala types which even gives better performance.

techoe: OK. I will change double reference to call by value.

techoe: I updated the code to have to call by value of double type!

PR #7302, opened 2019-03-15
planning: proto cleanup

jmtao:


modules/planning/proto/decider_config.proto
```diff
@@ -42,10 +36,9 @@ message RuleOpenSpaceStopConfig {
 }
 
 message DeciderRuleBasedStopConfig {
-  optional RuleCrosswalkConfig crosswalk = 1;
-  optional RuleStopSignConfig stop_sign = 2;
-  optional RuleTrafficLightConfig traffic_light = 3;
-  optional RuleOpenSpaceStopConfig open_space = 4;
+  optional RuleStopSignConfig stop_sign = 1;
+  optional RuleTrafficLightConfig traffic_light = 2;
+  optional RuleOpenSpaceStopConfig open_space = 3;
```
xiaoxq: Is this proto for text config only or ever recorded somewhere? If there is stale data, better to keep the index number unchanged.

jmtao: this is internal to planning.

PR #7305, opened 2019-03-15
planning: 1st check-in of bare-intersection-protected scenario

jmtao:


modules/planning/scenarios/bare_intersection/protected/stage_approach.cc
```diff
@@ -0,0 +1,66 @@
...
+ *****************************************************************************/
+
+/**
+ * @file
+ **/
+
+#include <vector>
+
+#include "modules/planning/scenarios/bare_intersection/protected/stage_approach.h"
+
```
YajiaZhang: Move the class header before system header? Specific class header -> system headers -> other headers

jmtao: thx. amended

PR #7307, opened 2019-03-15
Planning: add end_pose of each frame into debug in open space planner

Capri2014:


Capri2014:
To debug for the end pose support for different vehicles.

PR #7316, opened 2019-03-18
Add a new docs file of M2 GPS device

lishuanghua:
Add a new docs file of M2 GPS device which supports static initialization


mickeyouyou:
After a detailed comparison, the data structure designed in the file newtonm2_message.h is completely duplicated, and the existing types of novatel_message.h can be reused.

PR #7323, opened 2019-03-18
Planning: update path info for speed decider.

panjiacheng:


modules/planning/common/path/path_data.h
```diff
@@ -36,12 +36,11 @@ namespace planning {
...
  public:
   enum class PathPointType {
-    UNKNOWN,
-    OFF_ROAD,
     IN_LANE,
     OUT_ON_FORWARD_LANE,
     OUT_ON_REVERSE_LANE,
-    BACK_TO_IN_LANE,
+    OFF_ROAD,
+    UNKNOWN,
```
xiaoxq: Just a reminder that be careful about changing enum value. It breaks if some existing code is using its integer value, and the implicite default value which is the first entry.

modules/planning/tasks/deciders/BUILD
```diff
@@ -79,12 +80,29 @@ cc_library(
...
+    name = "path_decider_obstacle_utils",
+    srcs = [
+        "path_decider_obstacle_utils.cc",
+    ],
+    hdrs = [
+        "path_decider_obstacle_utils.h",
+    ],
+    copts = [
+        "-DMODULE_NAME=\\\"planning\\\"",
+    ],
```
xiaoxq: AFAIK, this definition is only necessary for the final cc_binary target.

modules/planning/tasks/deciders/path_assessment_decider.cc
```diff
@@ -93,8 +101,17 @@ bool PathAssessmentDecider::IsValidFallbackPath(
...
+  // Go through every path_point, and label its:
+  //  - in-lane/out-of-lane info
+  //  - distance to the closest obstacle.
+  std::vector<PathPointDecision> path_decision;
+  InitPathPointDecision(*path_data, &path_decision);
+  SetPathPointType(reference_line_info, *path_data, &path_decision);
+  SetObstacleDistance(reference_line_info, *path_data, &path_decision);
+
+  path_data->SetPathPointDecisionGuide(path_decision);
   return;
```
xiaoxq: Remove `return`, as well as all other places.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -123,15 +124,15 @@ Status PathBoundsDecider::Process(
...
     Frame* frame, ReferenceLineInfo* reference_line_info,
     std::vector<std::tuple<double, double, double>>* const path_boundaries) {
   // Sanity checks.
   CHECK_NOTNULL(frame);
   CHECK_NOTNULL(reference_line_info);
 
   // 1. Initialize the path boundaries to be an indefinitely large area.
-  if (!InitPathBoundaries(reference_line_info->reference_line(),
+  if (!InitPathBoundary(reference_line_info->reference_line(),
                           frame->PlanningStartPoint(), path_boundaries)) {
```
xiaoxq: Fix alignment when you changed corresponding code.

modules/planning/tasks/deciders/path_decider_obstacle_utils.cc
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+
+#include "modules/planning/tasks/deciders/path_decider_obstacle_utils.h"
+
+namespace apollo {
+namespace planning {
+
+constexpr double kStaticObstacleSpeedThreshold = 0.5;
```
xiaoxq: This becomes a global variable.
Use `static` or anonymous namespace to make it local.

panjiacheng: Actually, this is meant to be global. Later I will put all constexpr into GFLAGS.

modules/planning/common/path/path_data.h
panjiacheng: Thanks. I just wanted to change the sequence to make it more meaningful and didn't add/delete anything.

panjiacheng: Thanks. This enum has not been used yet. I talked with Jinyun and we both agreed to delete this BACK_TO_IN_LANE one.

modules/planning/tasks/deciders/BUILD
panjiacheng: deleted.

modules/planning/tasks/deciders/path_assessment_decider.cc
panjiacheng: removed all such things.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: fixed alignments.

PR #7324, opened 2019-03-18
planning: update ReferenceLineInfo.GetPathTurnType()

jmtao:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -191,9 +191,11 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectTrafficLightScenario(
   const auto& reference_line_info = frame.reference_line_info().front();
   const double adc_front_edge_s = reference_line_info.AdcSlBoundary().end_s();
   const bool right_turn =
-      (reference_line_info.GetPathTurnType() == hdmap::Lane::RIGHT_TURN);
+      (reference_line_info.GetPathTurnType(adc_front_edge_s) ==
+          hdmap::Lane::RIGHT_TURN);
   const bool left_turn =
-      (reference_line_info.GetPathTurnType() == hdmap::Lane::LEFT_TURN);
+      (reference_line_info.GetPathTurnType(adc_front_edge_s) ==
+          hdmap::Lane::LEFT_TURN);
```
xiaoxq: They are only used for debug information. Better to use "hdmap::Lane::LaneType_Name(reference_line_info.GetPathTurnType(adc_front_edge_s))" to get the readable string directly.

jmtao: they are not for debug use only. we need the type.

PR #7344, opened 2019-03-19
planning: added orientated bounding polygon implementation for sl bou…

YajiaZhang:
…ndary

### DIFF
--- a/modules/planning/reference_line/reference_line.cc
+++ b/modules/planning/reference_line/reference_line.cc
@@ -41,6 +41,7 @@ namespace planning {
 using MapPath = hdmap::Path;
 using apollo::common::SLPoint;
 using apollo::common::math::CartesianFrenetConverter;
+using apollo::common::math::Vec2d;
 using apollo::common::util::DistanceXY;
 using apollo::hdmap::InterpolatedIndex;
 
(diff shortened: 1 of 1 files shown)


modules/planning/reference_line/reference_line.cc
```diff
@@ -626,18 +627,56 @@ bool ReferenceLine::GetSLBoundary(const common::math::Box2d& box,
...
+  // the order must be counter-clockwise
+  std::vector<SLPoint> sl_corners;
   for (const auto& point : corners) {
     SLPoint sl_point;
     if (!XYToSL(point, &sl_point)) {
       AERROR << "failed to get projection for point: " << point.DebugString()
              << " on reference line.";
       return false;
     }
+    sl_corners.push_back(std::move(sl_point));
```
xiaoxq: Cause undefined behavior as the sl_point will be still in use later.

YajiaZhang: No, it will not. The variable only exists in the for loop and will not be accessed after the line.

YajiaZhang: My bad, copy-pasted some code after the line. Will fix.

PR #7349, opened 2019-03-20
Planning: OpenSpace: use last frame lane info to generate roi

JasonZhou404:


modules/planning/common/open_space_info.h
```diff
@@ -87,6 +87,14 @@ class OpenSpaceInfo {
     return &target_parking_spot_;
   }
 
+  const hdmap::LaneInfoConstPtr target_parking_lane() const {
+    return target_parking_lane_;
+  }
+
+  hdmap::LaneInfoConstPtr *mutable_target_parking_lane() {
+    return &target_parking_lane_;
```
xiaoxq: It's already a shared pointer. Not safe to get its pointer which breaks the ownership rule.

PR #7363, opened 2019-03-20
Planning: fix fail safe for single thread reference line provider

JasonZhou404:


modules/planning/reference_line/reference_line_provider.cc
```diff
@@ -239,26 +239,25 @@ bool ReferenceLineProvider::GetReferenceLines(
...
-    UpdateReferenceLine(*reference_lines, *segments);
-    double end_time = Clock::NowInSeconds();
-    last_calculation_time_ = end_time - start_time;
   }
+
+  AWARN << "Reference line is NOT ready.";
+  if (reference_line_history_.empty()) {
+    return false;
+  }
+  reference_lines->assign(reference_line_history_.back().begin(),
```
Capri2014: Add a log indicating we are using reference line from history

PR #7372, opened 2019-03-21
Prediction: add __init__ for python import

HongyiSun:


modules/tools/prediction/data_pipelines/__init__.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+
+###############################################################################
+# Copyright 2018 The Apollo Authors. All Rights Reserved.
+#
```
YajiaZhang: 2019

PR #7394, opened 2019-03-22
Planning: multiple updates.

panjiacheng:


modules/planning/tasks/deciders/path_assessment_decider.cc
```diff
@@ -46,26 +46,49 @@ Status PathAssessmentDecider::Process(
...
+  }
+
+  // Pick the optimal path
+  std::sort(valid_path_data.begin(), valid_path_data.end(),
+            [](const PathData& lhs, const PathData& rhs) {
+              if (lhs.path_label() != rhs.path_label()) {
+                return lhs.path_label() == "regular";
+              }
+              return true;
+            });
```
xiaoxq: if lhs.path_label() == rhs.path_label(),
compare(lhs, rhs) equals to compare(rhs, lhs), which is confusing.

panjiacheng: fixed this.

PR #7408, opened 2019-03-22
add hash code to prevent dreamview from updating NavigationInfo topic with the same hash code

DongAi:
In navigation mode, /apollo/navigation topic is published in a very low rate, even only once in a long time. But, dreamview gets /apollo/navigation topic and updates navigation paths in every frame. So add hash code to identify different /apollo/navigation with different paths. If the hash code of /apollo/navigation topic is the same with that of the latest topic, dreamview do not its paths. 
On TX2, this feature reduces CPU usage of dreamview by 15%-20%.


modules/map/relative_map/relative_map.cc
```diff
@@ -289,6 +289,15 @@ void RelativeMap::LoadNavigationPath() {
...
   }
 
+  // Generate the hash code, dreamview does not update the navigation data
+  // having the same hash code with the latest navigation data.
+  // Generate hash code before filling proto header.
+  std::string proto_content;
+  navigation_info.SerializeToString(&proto_content);
+  static std::hash<std::string> hash_func;
+  std::size_t hash_code = hash_func(proto_content);
+  navigation_info.set_hashcode(hash_code);
```
xiaoxq: https://github.com/ApolloAuto/apollo/blob/master/modules/common/util/message_util.h#L92

siyangy:
Thanks for the investigation! It seems comparing the timestamp is better here.

modules/map/relative_map/relative_map.cc
DongAi: The hashcode is generated before setting header field, so it's just generated based on the topic's content, or path info.
This pull request is going to merge to branch v3.1_dev, and there is no file message_util.h in branch v3.1_dev, so hashcode generating function is declared locally.

DongAi:
> comparing

modules/map/relative_map/relative_map.cc
DongAi: @siyangy This pull request tries to prevent topics being updated with the same path info, and timestamps could distinguish topics rather than the content of topics. If we publish the two topics with the same path info on different time, both of them will be updated once.

modules/dreamview/backend/simulation_world/simulation_world_service.cc
```diff
@@ -1013,6 +1013,20 @@ void SimulationWorldService::UpdateSimulationWorld(
...
+    const auto hashcode = navigation_info.hashcode();
+    if (topic_hashcode_map_.find(topic_name) !=
+        topic_hashcode_map_.end()) {
+      // it seems to be the same topic, ignore this update
+      if (topic_hashcode_map_[topic_name] == hashcode) {
+        return;
+      }
+    }
+
+    topic_hashcode_map_[topic_name] = hashcode;
```
siyangy: why do you need a map if the topic_name is always "NavigationInfo"?

DongAi: I considered that it's more convenience for further extension. What if another topic also needs to be distinguished via hashcode? So I used an unorderd_map to store hashcodes.

siyangy: I don't see a reason to optimize prematurely here, normally a module is not supposed to resend identical data. Also what you are really doing here is to dedup consecutive message with same content, just having a local static to store previous hash_code suffices.

DongAi: Thanks for your suggestion. Using timestamp to check and using local variable to store timestamp also meet the needs of this feature, simply and clean enough^^. I'll modify and commit today.

modules/dreamview/backend/simulation_world/simulation_world_service.cc
```diff
@@ -1013,6 +1013,14 @@ void SimulationWorldService::UpdateSimulationWorld(
 template <>
 void SimulationWorldService::UpdateSimulationWorld(
     const NavigationInfo &navigation_info) {
+  static double previous_timestamp = 0.0;
+  double topic_timestamp = navigation_info.header().timestamp_sec();
+  if (previous_timestamp == topic_timestamp) {
+    // it seems to be the same topic, ignore this update
+    return;
+  }
```
xiaoxq: Not safe to compare float numbers to be exactly equal.

DongAi: Fixed, please review again. THX.^^

PR #7417, opened 2019-03-22
Prediction: use simple-version of move_sequence_predictor

kechxu:


kechxu:
@panjiacheng The sampling-based method will be used by caution-level obstacles later.

PR #7421, opened 2019-03-22
planning: openspace, disable export chart flag for conflict issue

deidaraho:
remove redundant debug flag in open space

### DIFF
--- a/modules/planning/common/open_space_info.h
+++ b/modules/planning/common/open_space_info.h
@@ -229,7 +229,7 @@ class OpenSpaceInfo {
   }
 
   const std::vector<TrajGearPair> &paritioned_trajectories() const {
-    // to-do(Runxin): export to chart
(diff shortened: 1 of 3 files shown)


modules/planning/common/open_space_info.h
```diff
@@ -229,6 +229,7 @@ class OpenSpaceInfo {
   }
 
   const std::vector<TrajGearPair> &paritioned_trajectories() const {
+    // to-do: export to chart
```
xiaoxq: TODO(Who): xxx.

deidaraho: > TODO(Who): xxx.

fixed

xiaoxq: Not yet. Should exactly start with `TODO(Renxin): `.

PR #7423, opened 2019-03-22
Prediction: start new right_of_way function

HongyiSun:


modules/prediction/proto/lane_graph.proto
```diff
@@ -60,6 +60,7 @@ message LaneSequence {
   optional double lane_s = 17;
   optional double lane_l = 18;
   optional StopSign stop_sign = 20;
+  optional int32 right_of_way = 21 [default = 50];
```
xiaoxq: 1. Change to unsigned int if it's supposed to be.
2. Comment about the supposed range.

modules/prediction/scenario/BUILD
```diff
@@ -10,6 +10,7 @@ cc_library(
         "//modules/prediction/scenario/analyzer:scenario_analyzer",
         "//modules/prediction/scenario/feature_extractor",
         "//modules/prediction/scenario/prioritization:obstacles_prioritizer",
+        "//modules/prediction/scenario/right_of_way:right_of_way",
```
xiaoxq: Simply "//modules/prediction/scenario/right_of_way" in such case.

modules/prediction/scenario/right_of_way/right_of_way.h
```diff
@@ -0,0 +1,40 @@
...
+   */
+  RightOfWay() = delete;
+
+  /**
+   * @brief Set right_of_way for all lane_sequence
+   * @return Scenario features
+   */
+  static void Analyze();
+
+ private:
```
xiaoxq: Remove

PR #7431, opened 2019-03-23
improve initial functions

muleisheng:


modules/perception/onboard/component/lane_detection_component.cc
```diff
@@ -181,30 +189,11 @@ bool LaneDetectionComponent::Init() {
...
-  if (InitMotionService() != cyber::SUCC) {
-    AERROR << "InitMotionService() failed.";
-    return false;
-  }
-  if (InitCameraListeners() != cyber::SUCC) {
-    AERROR << "InitCameraListeners() failed.";
-    return false;
+  for (auto &itor : LaneDetectionComponent::init_func_arry_) {
+    if ((this->*(itor.init_function_))() != cyber::SUCC) {
+      AERROR << itor.init_fun_name_<< "() failed.";
```
xiaoxq: IMHO, it's not that straight forward for people to understand this piece of code.
Add Perception folks to review.

muleisheng: I want to encapsulate these code like as boost::function to adapter our special requirement in future. I want to remove  duplicated code as far as possible

muleisheng: I will add some basic tool classes to make it easy to use. This may implement some code or logic to reuse

yg13: Hi Mulei,  thanks for the suggestion. However, because all the other components kind of follow the old convention, I prefer the keep the old way.

muleisheng:
encapsulated these code as a macro may make it easy to use:)

PR #7441, opened 2019-03-25
planning:enable stage config check in each stage

mickeyouyou:
1. enable stage config check; finish current stage if `enable = false`;
2. expand scope to all scenarios


modules/common/util/factory.h
```diff
@@ -65,7 +65,7 @@ class Factory {
    * @param id Identifier of the class being registered
    * @param creator Function returning a pointer to an instance of
    * the registered class
-   * @return True iff the key id is still available
+   * @return True if the key id is still available
```
xiaoxq: `iff` sometimes means `if and only if`. I'm not sure if it applies here.

mickeyouyou: Let us confirm

modules/planning/scenarios/bare_intersection/unprotected/stage_approach.cc
```diff
@@ -41,6 +41,11 @@ Stage::StageStatus BareIntersectionUnprotectedStageApproach::Process(
 
   scenario_config_.CopyFrom(GetContext()->scenario_config);
 
+  if (!config_.enabled()) {
+    ADEBUG << "stage Approach finished because not enable.";
+    return Stage::FINISHED;
+  }
+
```
jmtao: this is not how it works.
for each stage, after it finishes, it shall set next_stage_ , for the stage machine to move on. 
only setting up stage status doesn't help in that way. 
And secondly, although we have enable flag as conf for stages, we'd better not allow skip stages in general in code implementation, to avoid bad conf by mistakes.

modules/planning/scenarios/bare_intersection/unprotected/stage_intersection_cruise.cc
```diff
@@ -38,6 +38,11 @@ Stage::StageStatus BareIntersectionUnprotectedStageIntersectionCruise::Process(
   ADEBUG << "stage: IntersectionCruise";
   CHECK_NOTNULL(frame);
 
+  if (!config_.enabled()) {
+    ADEBUG << "stage IntersectionCruise finished because not enable.";
+    return FinishStage();
+  }
+
```
jmtao: same as above

modules/planning/scenarios/bare_intersection/unprotected/stage_approach.cc
mickeyouyou: Thanks for your reminding.

mickeyouyou:
there is a bug for this pull request, close it for now.

PR #7453, opened 2019-03-25
Planning: add backside caution-level vehicle to St-boundary

JasonXu12:


modules/planning/common/obstacle.h
```diff
@@ -124,6 +124,10 @@ class Obstacle {
 
   static bool IsValidTrajectoryPoint(const common::TrajectoryPoint& point);
 
+  const bool IsCautionLevelObstacle() const {
+      return is_caution_level_obstacle_;
+    }
+
```
xiaoxq: Fix indents.

modules/planning/common/reference_line_info.cc
```diff
@@ -385,6 +385,8 @@ bool ReferenceLineInfo::AddObstacles(
 }
 
 bool ReferenceLineInfo::IsUnrelaventObstacle(const Obstacle* obstacle) {
+  if (obstacle->IsCautionLevelObstacle())
+    return false;
```
xiaoxq: Always apply brackets to scopes.

modules/planning/traffic_rules/backside_vehicle.cc
```diff
@@ -36,8 +36,9 @@ void BacksideVehicle::MakeLaneKeepingObstacleDecision(
       adc_sl_boundary.end_s() - adc_sl_boundary.start_s();
   for (const auto* obstacle : path_decision->obstacles().Items()) {
     if (obstacle->PerceptionSLBoundary().end_s() >=
-        adc_sl_boundary.end_s()) {  // don't ignore such vehicles.
-      continue;
+        adc_sl_boundary.end_s() || obstacle->IsCautionLevelObstacle()) {
+        // don't ignore such vehicles.
+        continue;
```
xiaoxq: Indents as 2.

JasonXu12: Fixed. Thanks!

modules/planning/common/reference_line_info.cc
JasonXu12: Fixed. Thanks!

modules/planning/common/obstacle.h
JasonXu12: Fixed. Thanks!

modules/planning/common/obstacle.h
```diff
@@ -124,6 +124,10 @@ class Obstacle {
 
   static bool IsValidTrajectoryPoint(const common::TrajectoryPoint& point);
 
+  const bool IsCautionLevelObstacle() const {
```
xiaoxq: 1. Don't return a `const` copy which has no benefit but harm.
2. Better to declare `inline` for functions in header file, or you may run into "redefinition" error.

JasonXu12: Fixed. Thanks!

PR #7478, opened 2019-03-26
Planning: plot the output path on dreamview.

panjiacheng:


modules/planning/common/planning_context.h
```diff
@@ -104,7 +104,8 @@ class PlanningContext {
 
   static const LaneFollowInfo& lane_follow_info() { return lane_follow_info_; }
 
-  static LaneFollowInfo* mutable_lane_follow_info() { return &lane_follow_info_; }
+  static LaneFollowInfo* mutable_lane_follow_info() {
+      return &lane_follow_info_; }
```
xiaoxq: No need to change

modules/planning/tasks/deciders/path_assessment_decider.cc
```diff
@@ -69,13 +73,16 @@ Status PathAssessmentDecider::Process(
     const std::string msg = "Neither regular nor fallback path is valid.";
     AERROR << msg;
     return Status(ErrorCode::PLANNING_ERROR, msg);
+  } else {
+    ADEBUG << "There are " << valid_path_data.size() << " valid path data.";
```
xiaoxq: No need to put in `else`

modules/planning/common/planning_context.h
panjiacheng: line was too long and violated lint.

modules/planning/tasks/deciders/path_assessment_decider.cc
panjiacheng: fixed.

PR #7480, opened 2019-03-26
planning: move traffic-light stop fence creation from decider to traf…

jmtao:
…fic-rule

### DIFF
--- a/modules/planning/traffic_rules/traffic_light.h
+++ b/modules/planning/traffic_rules/traffic_light.h
@@ -20,6 +20,12 @@
 
 #pragma once
 
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "modules/perception/proto/traffic_light_detection.pb.h"
 #include "modules/planning/traffic_rules/traffic_rule.h"
 
 namespace apollo {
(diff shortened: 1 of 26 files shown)


modules/planning/traffic_rules/traffic_light.h
```diff
@@ -0,0 +1,49 @@
...
+
+ private:
+  void MakeDecisions(Frame* const frame,
+                     ReferenceLineInfo* const reference_line_info);
+  int BuildStopDecision(Frame* const frame,
+                        ReferenceLineInfo* const reference_line_info,
+                        const hdmap::PathOverlap& traffic_light_overlap);
+
+ private:
+  static constexpr char const* const TRAFFIC_LIGHT_VO_ID_PREFIX = "TL_";
```
xiaoxq: If it's only used in this component, simply put in the `.cc` file and close to where it's used. Header file is some kind of `interface` to external components while this information is unused to them. So I would prefer minimum scope.

jmtao: This variable used to be a gflag to be "configurable". But we haven't really touched it ever since so we moved it inside the class, to make it less flexible.
But still, it is still not really a "local" variable, or should be kept unseen. So I think it's more appropriate to put it in the header file. So we can still be aware of it just in case if we want to change it. 
it's a private member.

PR #7494, opened 2019-03-27
Common: add processing time in header for performance evaluation

kechxu:


modules/common/proto/header.proto
```diff
@@ -32,4 +32,7 @@ message Header {
   optional StatusPb status = 8;
 
   optional string frame_id = 9;
+
+  // The processing time in ms
+  optional double processing_time_ms = 10;
```
xiaoxq: How about in seconds? As the Clock::Now() returns seconds.

kechxu: We can use std::chrono::system_clock::now(), instead

PR #7495, opened 2019-03-27
Planning: updated side_pass_scenario

panjiacheng:


modules/planning/common/planning_context.h
```diff
@@ -40,6 +40,8 @@
 namespace apollo {
 namespace planning {
 
+constexpr int kPathScenarioTransitionHysteresisFrame = 5;
```
xiaoxq: Local vars should be either `static` or in anonymous namespace to avoid global namespace pollution.

modules/planning/scenarios/side_pass/side_pass_scenario.h
```diff
@@ -61,6 +61,12 @@ class SidePassScenario : public Scenario {
 
   bool IsFarFromDestination(const Frame& frame);
 
+  bool IsWithinSidePassingSpeedADC(const Frame& frame);
+
+  bool IsSidePassableObstacle(
+    const Frame& frame, const ReferenceLineInfo& reference_line_info,
+    const std::string& blocking_obstacle_id);
```
xiaoxq: `bool function() const;` if it doesn't change the instance.

modules/planning/common/planning_context.h
panjiacheng: I will put all such thing into GFLAGS at the very end when dev is finished. They are meant to be used globally.

modules/planning/scenarios/side_pass/side_pass_scenario.h
panjiacheng: Fixed this and other similar occurrences.

PR #7501, opened 2019-03-27
tools: record analyzer added two more metrics for planning evaluation.

yifeijiang:


modules/tools/record_analyzer/module_planning_analyzer.py
```diff
@@ -69,15 +70,18 @@ def put(self, adc_trajectory):
...
             if adc_trajectory.debug.planning_data.HasField('init_point'):
                 self.init_point_curvature.append(
                     abs(adc_trajectory.debug.planning_data.init_point.path_point.kappa))
                 self.init_point_dcurvature.append(
                     abs(adc_trajectory.debug.planning_data.init_point.path_point.dkappa))
+                accel = adc_trajectory.debug.planning_data.init_point.a
+                if accel > 0:
+                    self.init_point_accel.append(accel)
+                if accel < 0:
+                    self.init_point_decel.append(abs(accel))
```
xiaoxq: Simply `self.init_point_decel.append(abs(accel))`?

PR #7509, opened 2019-03-28
Planning: Finish porting and enable all debug chart for open space planner

Capri2014:
Will optimize and re-org of the charts later.

### DIFF
--- a/modules/planning/on_lane_planning.cc
+++ b/modules/planning/on_lane_planning.cc
@@ -610,8 +610,8 @@ void OnLanePlanning::ExportOpenSpaceChart(
   if (FLAGS_enable_record_debug) {
     AddOpenSpaceOptimizerResult(debug_info, debug_chart);
     AddPartitionedTrajectory(debug_info, debug_chart);
-    // AddStitchSpeedProfile(debug);
-    // AddPublishedSpeed(debug, ptr_trajectory_pb);
+    AddStitchSpeedProfile(debug_chart);
+    // AddPublishedSpeed(debug_info, ptr_trajectory_pb);
     // AddPublishedAcceleration(debug, ptr_trajectory_pb);
   }
 }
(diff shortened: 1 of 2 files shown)


modules/planning/on_lane_planning.cc
```diff
@@ -705,13 +705,12 @@ void OnLanePlanning::AddPartitionedTrajectory(
   }
   chart->set_title("Open Space Partitioned Trajectory");
   // has to define chart boundary first
-  auto chosen_trajectory =
-      open_space_debug.chosen_trajectory().trajectory()[0];
+  auto chosen_trajectory = open_space_debug.chosen_trajectory().trajectory()[0];
```
xiaoxq: `const auto&`?

modules/planning/on_lane_planning.cc
```diff
@@ -746,5 +745,160 @@ void OnLanePlanning::AddPartitionedTrajectory(
...
+
+  // auto smoothed_trajectory = open_space_debug.smoothed_trajectory();
+  auto* speed_profile = chart->add_line();
+  speed_profile->set_label("Speed Profile");
+  for (const auto& point : trajectory_pb->trajectory_point()) {
+    auto* point_debug = speed_profile->add_point();
+    point_debug->set_x(point.relative_time() +
+                       trajectory_pb->header().timestamp_sec());
+    if (trajectory_pb->gear() == canbus::Chassis::GEAR_DRIVE)
+      point_debug->set_y(point.v());
```
xiaoxq: Add brackets to all scopes.

modules/planning/on_lane_planning.cc
Capri2014: Done.

modules/planning/on_lane_planning.cc
Capri2014: Done.

modules/planning/on_lane_planning.cc
```diff
@@ -745,5 +746,161 @@ void OnLanePlanning::AddPartitionedTrajectory(
...
+  auto* speed_profile_properties = speed_profile->mutable_properties();
+  (*speed_profile_properties)["borderWidth"] = "2";
+  (*speed_profile_properties)["pointRadius"] = "0";
+  (*speed_profile_properties)["lineTension"] = "0";
+  (*speed_profile_properties)["fill"] = "false";
+  (*speed_profile_properties)["showLine"] = "true";
+}
+
+void OnLanePlanning::AddPublishedSpeed(ADCTrajectory* const trajectory_pb,
+                                       planning_internal::Debug* debug_chart) {
```
xiaoxq: Accept const reference as input. Don't let others break you with a NULL or wild pointer.

Capri2014: Done with other minor cleaning up to avoid confusion.

PR #7534, opened 2019-03-29
Prediction: initialize extention predictor and more integration of lane scanning evaluator

kechxu:


kechxu:
> "extention" is a misspelling version of "extension". Also, it might be better to use "extrapolation" or some thing like that.

PR #7591, opened 2019-04-01
planning: make SidePassScenario::IsTransferable() static so that it c…

jmtao:
…an be called without an instance. Creating an instance just to call this function caused a bug

### DIFF
--- a/modules/planning/scenarios/scenario_manager.cc
+++ b/modules/planning/scenarios/scenario_manager.cc
@@ -221,8 +221,8 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectTrafficLightScenario(
   const std::vector<PathOverlap>& traffic_light_overlaps =
       reference_line_info.reference_line().map_path().signal_overlaps();
   for (const auto& traffic_light_overlap : traffic_light_overlaps) {
-    const double dist =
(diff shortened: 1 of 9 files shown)


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -221,8 +221,8 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectTrafficLightScenario(
   const std::vector<PathOverlap>& traffic_light_overlaps =
       reference_line_info.reference_line().map_path().signal_overlaps();
   for (const auto& traffic_light_overlap : traffic_light_overlaps) {
-    const double dist =
-        traffic_light_overlap.start_s - first_encountered_traffic_light.start_s;
+    const double dist = traffic_light_overlap.start_s -
+        first_encountered_traffic_light.start_s;
```
xiaoxq: Seems like the clang-format tool prefer the existing style.

jmtao: I'll leave it to your bot to update it. 
too much work to keep it unchanged after rebasing...

xiaoxq: :100:

PR #7593, opened 2019-04-01
Planning: minor change for potential bug fix.

panjiacheng:


modules/planning/tasks/deciders/path_assessment_decider.cc
```diff
@@ -99,6 +100,9 @@ Status PathAssessmentDecider::Process(
   // 3. Pick the optimal path.
   std::sort(valid_path_data.begin(), valid_path_data.end(),
             [](const PathData& lhs, const PathData& rhs) {
+              if (lhs.Empty() && rhs.Empty()) {
+                return false;
+              }
```
xiaoxq: if (lhs.Empty()) return false;
if (rhs.Empty()) return true;

PR #7594, opened 2019-04-01
Planning: update logic to fix bug for open space ROI

Capri2014:


modules/planning/tasks/deciders/open_space_roi_decider.cc
```diff
@@ -46,6 +46,13 @@ OpenSpaceRoiDecider::OpenSpaceRoiDecider(const TaskConfig &config)
 Status OpenSpaceRoiDecider::Process(Frame *frame) {
   // TODO(Jinyun) only run GetOpenSpaceROI() at the first frame of
   // open space planner to save computation effort
+
+  if (frame == nullptr) {
+    const std::string msg =
+        "Invalid frame, fail to process the OpenSpaceRoiDecider.";
+    return Status(ErrorCode::PLANNING_ERROR, msg);
```
xiaoxq: Simply put the msg into the statement to reduce a local variable.

Capri2014: Oh, I forgot to add an AERROR for the message.

PR #7597, opened 2019-04-02
planning: open space fixed wrong fallback trajectory

deidaraho:
TODO(Runxin): find a better fallback trajectory.

### DIFF
--- a/modules/planning/tasks/deciders/open_space_fallback_decider.cc
+++ b/modules/planning/tasks/deciders/open_space_fallback_decider.cc
@@ -33,13 +33,13 @@ OpenSpaceFallbackDecider::OpenSpaceFallbackDecider(const TaskConfig& config)
 
 Status OpenSpaceFallbackDecider::Process(Frame* frame) {
   std::vector<std::vector<common::math::Box2d>> predicted_bounding_rectangles;
-  double collision_distance;
+  double obstacle_to_vichcle_distance = 0.0;
 
   BuildPredictedEnvironment(frame->obstacles(), predicted_bounding_rectangles);
 
   if (!IsCollisionFreeTrajectory(
           frame->open_space_info().chosen_paritioned_trajectory(),
-          predicted_bounding_rectangles, &collision_distance)) {
+          predicted_bounding_rectangles, &obstacle_to_vichcle_distance)) {
     // change gflag
     frame_->mutable_open_space_info()->set_fallback_flag(true);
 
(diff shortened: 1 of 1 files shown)


modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -33,7 +33,7 @@ OpenSpaceFallbackDecider::OpenSpaceFallbackDecider(const TaskConfig& config)
 
 Status OpenSpaceFallbackDecider::Process(Frame* frame) {
   std::vector<std::vector<common::math::Box2d>> predicted_bounding_rectangles;
-  double collision_distance;
```
Capri2014: Load this one from configs

modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -48,35 +48,25 @@ Status OpenSpaceFallbackDecider::Process(Frame* frame) {
     *(frame_->mutable_open_space_info()->mutable_fallback_trajectory()) =
         frame->open_space_info().chosen_paritioned_trajectory();
     auto fallback_trajectory_vec =
```
JasonZhou404: use another variable name. It's a traj and gear pair actually

deidaraho: > use another variable name. It's a traj and gear pair actually

fixed

modules/planning/tasks/deciders/open_space_fallback_decider.cc
deidaraho: > Load this one from configs

I change the varaible's name, it is the distance between obstacle and vehicle

mickeyouyou: > > Load this one from configs
> 
> I change the varaible's name, it is the distance between obstacle and vehicle

It is a variable mistake about  `obstacle_to_vichcle_distance` => `obstacle_to_vehicle_distance` ? or other expressions of vehicle?

Capri2014: @mickeyouyou Thanks for the catch and I submitted a commit to fix it.

mickeyouyou: > @mickeyouyou Thanks for the catch and I submitted a commit to fix it.

👍

PR #7611, opened 2019-04-02
Planning: reduce running time in path_bounds decider.

panjiacheng:


modules/planning/tasks/deciders/path_assessment_decider.cc
```diff
@@ -41,6 +41,8 @@ using apollo::hdmap::HDMapUtil;
 // PointDecision contains (s, PathPointType, distance to closest obstacle).
 using PathPointDecision = std::tuple<double, PathData::PathPointType, double>;
 
+constexpr double kMinObstacleArea = 1e-4;
```
xiaoxq: This is a global variable. Mark either `static` or anonymous namespace to make it local to this file.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -245,9 +248,10 @@ std::string PathBoundsDecider::GenerateFallbackPathBound(
   // PathBoundsDebugString(*path_bound);
 
   // 2. Decide a rough boundary based on lane info and ADC's position
-  if (!GetBoundaryFromLanesAndADC(reference_line_info.reference_line(),
+  std::string dummy;
```
Capri2014: Do not name it dummy, make the name more self-explaining, this string is for "borrow_lane_type" right?

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -315,6 +322,38 @@ bool PathBoundsDecider::GetBoundaryFromLanesAndADC(
 
     // 2. Get the neighbor lane widths at the current point.
     double curr_neighbor_lane_width = 0.0;
+    hdmap::Id neighbor_lane_id;
+    if (lane_borrow_info == LaneBorrowInfo::LEFT_BORROW) {
+      // Borrowing left neighbor lane.
+      if (reference_line_info.GetNeighborLaneInfo(
+              ReferenceLineInfo::LaneType::LeftForward,
```
xiaoxq: `using ...::LaneType;` at the beginning to shorten your code, as it's used many times.

modules/planning/tasks/deciders/path_assessment_decider.cc
panjiacheng: This is a global variable but in .cc only.
Others won't include a .cc file.

panjiacheng: I added anonymous namespace anyway.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: Modified.

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: Thx. I updated it to be dummy_borrow_lane_type. It is for "borrow lane type" but it is a dummy variable, too.

PR #7616, opened 2019-04-02
 Planning: openspace: trajectory partition debug

JasonZhou404:


modules/planning/tasks/optimizers/open_space_trajectory_partition/open_space_trajectory_partition.cc
```diff
@@ -219,11 +239,19 @@ Status OpenSpaceTrajectoryPartition::Process() {
       vehicle_state.gear() == canbus::Chassis::GEAR_REVERSE
           ? NormalizeAngle(vehicle_state.heading() + M_PI)
           : vehicle_state.heading();
+  // TODO(Jinyun) move to confs
+  constexpr double abs_static_linear_velocity = 0.2;
```
Capri2014: @JasonZhou404 This might needs to be consistent with the ones in Control, maybe change this one as a global config or use the one in vehicle_params?

PR #7623, opened 2019-04-03
Control: compute the acceleration, jerk variables for Lon/Lat Controller

kevin-y-wang:
Verified by the bag files generated from dreamland simulation

### DIFF
--- a/modules/control/controller/lat_controller.cc
+++ b/modules/control/controller/lat_controller.cc
@@ -493,13 +493,14 @@ void LatController::UpdateState(SimpleLateralDebug *debug) {
   if (FLAGS_use_navigation_mode) {
     ComputeLateralErrors(
         0.0, 0.0, driving_orientation_, vehicle_state->linear_velocity(),
-        vehicle_state->angular_velocity(), trajectory_analyzer_, debug);
+        vehicle_state->angular_velocity(), vehicle_state->linear_acceleration(),
+        trajectory_analyzer_, debug);
   } else {
     const auto &com = vehicle_state->ComputeCOMPosition(lr_);
-    ComputeLateralErrors(com.x(), com.y(), driving_orientation_,
-                         vehicle_state->linear_velocity(),
-                         vehicle_state->angular_velocity(),
-                         trajectory_analyzer_, debug);
+    ComputeLateralErrors(
+        com.x(), com.y(), driving_orientation_,
+        vehicle_state->linear_velocity(), vehicle_state->angular_velocity(),
+        vehicle_state->linear_acceleration(), trajectory_analyzer_, debug);
   }
 
   // State matrix update;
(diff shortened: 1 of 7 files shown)


kevin-y-wang:
Add some necessary comments into the control_cmd.proto for clarifying the terminology

modules/control/controller/lat_controller.cc
```diff
@@ -610,18 +611,37 @@ void LatController::ComputeLateralErrors(
...
+  previous_lateral_acceleration_ = lateral_error_dot_dot;
+
+  debug->set_heading_rate(angular_v);
+  debug->set_ref_heading_rate(target_point.path_point().kappa() *
+                              target_point.v());
+
+  debug->set_heading_error_rate(debug->heading_rate() -
+                                debug->ref_heading_rate());
+
+  debug->set_heading_acceleration(
```
Capri2014: 1. Not sure if calling "heading jerk" is a good idea, but I see the point of calling it here, so please add detailed calculation in notes and also saying A.K.A "yaw jerk". 

2. Since we add both "ref heading rate", "reference heading rate" as well as "heading error rate", why not add all them for both acceleration and jerk level? So we and have a total profiling for both the ones from planning and from the real vehicle feedback?

Capri2014: I see the comments been added. Thanks.

modules/control/proto/control_cmd.proto
```diff
@@ -80,40 +80,65 @@ message SimpleLongitudinalDebug {
...
-  optional double calibration_value = 15;
-  optional double throttle_cmd = 16;
-  optional double brake_cmd = 17;
-  optional bool is_full_stop = 18;
-  optional double slope_offset_compensation = 19;
-  optional double current_station = 20;
-  optional double path_remain = 21;
-  optional int32 pid_saturation_status = 22;
-  optional int32 leadlag_saturation_status = 23;
+  optional double acceleration_reference = 10;
```
Capri2014: Do not change the order of exsiting protos, please do ONLY append.

modules/control/proto/control_cmd.proto
```diff
@@ -80,40 +80,65 @@ message SimpleLongitudinalDebug {
...
+  optional double heading_error_rate = 10;
+  // heading_acceleration, as known as yaw acceleration, is the time derivative
+  // of heading rate,  in rad/s^2
+  optional double ref_heading_acceleration = 11;
+  optional double heading_acceleration = 12;
+  optional double heading_error_acceleration = 13;
+  // heading_jerk, as known as yaw jerk, is the second time derivative of
+  // heading rate, in rad/s^3
+  optional double ref_heading_jerk = 14;
+  optional double heading_jerk = 15;
```
Capri2014: same here.

PR #7628, opened 2019-04-03
esd can extended frame format

panxuechao:


modules/drivers/canbus/can_client/esd/BUILD
```diff
@@ -13,6 +13,7 @@ cc_library(
     deps = [
         "//modules/drivers/canbus/can_client",
         "//third_party/can_card_library/esd_can",
+        "//modules/drivers/canbus:sensor_gflags",
```
xiaoxq: Make it sorted.

modules/drivers/canbus/sensor_gflags.cc
```diff
@@ -32,3 +32,7 @@ DEFINE_double(sensor_freq, 100,
 
 // System gflags
 DEFINE_string(sensor_node_name, "", "Sensor node name.");
+
+// esd can extended frame gflags
+DEFINE_bool(esd_can_extended_frame, false,
+                "check esd can exdended frame enabled or not");
```
xiaoxq: Align with first argument.

modules/tools/gen_vehicle_protocol/extract_dbc_meta.py
```diff
@@ -21,7 +21,7 @@
 import shlex
 import re
 
-MAX_CAN_ID = 2048
+MAX_CAN_ID = 4096000000 #2048
```
xiaoxq: Two spaces before comments.

modules/tools/gen_vehicle_protocol/gen_protocols.py
```diff
@@ -66,7 +66,11 @@ def gen_report_cpp(car_type, p, output_dir):
         fmt_val["protocol_name_lower"] = p["name"]
         classname = p["name"].replace('_', '').capitalize()
         fmt_val["classname"] = classname
-        fmt_val["id_upper"] = p["id"].upper()
+        a = int(p["id"].upper(),16)
```
xiaoxq: 1. Space after `,`
2. Avoid single-char variable names. Please also fix the `p`.

modules/tools/gen_vehicle_protocol/gen_protocols.py
```diff
@@ -449,6 +457,15 @@ def gen_protocols(protocol_conf_file, protocol_dir):
...
 
+def gen_esd_can_extended(s):
+    """
+        id string:
+    """
+    a = int(s,16)
+    a = a & 0x1FFFFFFF
+    a = a | 0x20000000
+    s = hex(a).replace('0x','')
+    return s
```
xiaoxq: ditto

modules/tools/gen_vehicle_protocol/template/control_protocol.cc.tpl
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright 2018 The Apollo Authors. All Rights Reserved.
+ * Copyright 2019 The Apollo Authors. All Rights Reserved.
```
xiaoxq: Don't change year of existing files.

Capri2014:
@panxuechao  Nice work to extend the can message format to support extended frame! One thing is that this change would be applicable to all CAN card that supports extended frame (esd, hermes etc) and all CAN connected devices (vehicle, radar, etc), so could you rename the function name and remove the "esd" keyword to avoid confusion?

Capri2014:
my bad, this is only for esd canard.

PR #7638, opened 2019-04-03
Prediction: reduce the complexity of collision cost in interaction predictor

kechxu:


modules/prediction/predictor/interaction/interaction_predictor.cc
```diff
@@ -259,21 +259,51 @@ double InteractionPredictor::CentripetalAccelerationCost(
...
-    cost_abs_sum += std::abs(cost);
-    cost_sqr_sum += cost * cost;
+    double curr_s = GetSByConstantAcceleration(
+        speed, acceleration, relative_time);
+    double delta_s = curr_s - prev_s;
+    remained_s += delta_s;
+    while (lane_seg_idx < num_lane_segment) {
+      const LaneSegment& lane_segment =
+          lane_sequence.lane_segment(lane_seg_idx);
+      std::string lane_id = lane_segment.lane_id();
```
xiaoxq: const &

PR #7639, opened 2019-04-03
Docker: Save docker volume dependency info into meta.ini.

xiaoxq:


Capri2014:
Should we record this information also in master?

PR #7644, opened 2019-04-03
Dreamview: plotting vehicle bounding box to chart drawing element

vlin17:
car_with_bounding_box


Capri2014:
wait, why this one is not a rectangle ....

PR #7646, opened 2019-04-04
planning: stage Approach in BareIntersectionUnprotected scenario

jmtao:


modules/planning/scenarios/bare_intersection/unprotected/stage_approach.cc
```diff
@@ -42,16 +45,64 @@ Stage::StageStatus BareIntersectionUnprotectedStageApproach::Process(
...
 
-  // TODO(all): to be implemented
-  return Stage::RUNNING;
-}
+  const auto& reference_line_info = frame->reference_line_info().front();
 
-Stage::StageStatus BareIntersectionUnprotectedStageApproach::FinishScenario() {
-  next_stage_ = ScenarioConfig::NO_STAGE;
-  return Stage::FINISHED;
+  const std::string pnc_junction_overlap_id =
```
YajiaZhang: const std::string& ?

PR #7648, opened 2019-04-04
Common: add vehicle brake/throttle deadzone to common/data/vehicle_param

JasonXu12:


Capri2014:
@vlin17 updated per discussion.

PR #7666, opened 2019-04-04
Control: load brake/throttle deadzone params from common for lon/mpc …

JasonXu12:
…controller

### DIFF
--- a/modules/common/configs/proto/vehicle_config.proto
+++ b/modules/common/configs/proto/vehicle_config.proto
@@ -65,9 +65,9 @@ message VehicleParam {
   optional float max_abs_speed_when_stopped = 18;
 
   // minimum value get from chassis.brake, in percentage
-  optional float brake_deadzone = 19;
+  optional double brake_deadzone = 19;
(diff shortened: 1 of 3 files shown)


modules/common/configs/proto/vehicle_config.proto
```diff
@@ -65,9 +65,9 @@ message VehicleParam {
   optional float max_abs_speed_when_stopped = 18;
 
   // minimum value get from chassis.brake, in percentage
-  optional float brake_deadzone = 19;
+  optional double brake_deadzone = 19;
   // minimum value get from chassis.throttle, in percentage
-  optional float throttle_deadzone = 20;
+  optional double throttle_deadzone = 20;
```
xiaoxq: Is this proto ever published and recorded in data? If yes, this change will break compatibility.

JasonXu12: No, it hasn't. This is a non-recorded proto and the throttle/brake_deadzone are newly added to it.

PR #7671, opened 2019-04-04
planning: open space fix rigid transform bug when extracting obstacles

deidaraho:
1, remove virtual obstacles from open space
2, the rigid transfer to obstacles should be according to original point instead of the obstacle's center.

### DIFF
--- a/modules/planning/tasks/deciders/open_space_roi_decider.cc
+++ b/modules/planning/tasks/deciders/open_space_roi_decider.cc
@@ -83,6 +83,11 @@ bool OpenSpaceRoiDecider::VPresentationObstacle() {
 
   if (config_.open_space_roi_decider_config().enable_perception_obstacles()) {
     size_t perception_obstacles_num = obstacles_by_frame_->Items().size();
+    for (const auto &ob : obstacles_by_frame_->Items()) {
+      if (ob->IsVirtual()) {
+        --perception_obstacles_num;
+      }
+    }
 
     frame_->mutable_open_space_info()->set_obstacles_num(
         perception_obstacles_num + parking_boundaries_num);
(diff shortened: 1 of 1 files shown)


modules/planning/tasks/deciders/open_space_roi_decider.cc
```diff
@@ -83,6 +83,11 @@ bool OpenSpaceRoiDecider::VPresentationObstacle() {
 
   if (config_.open_space_roi_decider_config().enable_perception_obstacles()) {
     size_t perception_obstacles_num = obstacles_by_frame_->Items().size();
+    for (const auto &ob : obstacles_by_frame_->Items()) {
+      if (ob->IsVirtual()) {
+        perception_obstacles_num--;
```
xiaoxq: --perception_obstacles_num;

modules/planning/tasks/deciders/open_space_roi_decider.cc
```diff
@@ -107,13 +112,18 @@ bool OpenSpaceRoiDecider::VPresentationObstacle() {
     const auto &origin_point = frame_->open_space_info().origin_point();
     const auto &origin_heading = frame_->open_space_info().origin_heading();
     for (const auto &obstacle : obstacles_by_frame_->Items()) {
+      if (obstacle->IsVirtual()) { continue; }
```
xiaoxq: Expand

modules/planning/tasks/deciders/open_space_roi_decider.cc
deidaraho: > --perception_obstacles_num;

fixed.

modules/planning/tasks/deciders/open_space_roi_decider.cc
deidaraho: > Expand

fixed

PR #7677, opened 2019-04-05
modified the calibration scripts

wfgsy101:
modified the calibration scripts

### DIFF
--- a/modules/tools/calibration/process.py
+++ b/modules/tools/calibration/process.py
@@ -48,12 +48,12 @@ def get_start_index(data):
     if np.all(data['vehicle_speed'] == 0):
         return 0
 
-    start_ind = np.where(data['brake_percentage'] == 40)[0][0]
+    start_ind = np.where(data['vehicle_speed'] == 0.0)[0][0]
 
     ind = start_ind
     while ind < len(data):
-        if data['brake_percentage'][
-                ind] == 40:  #or data['vehicle_speed'][ind] == 0.0:
+        if data['vehicle_speed'][
+                ind] == 0.0:  #or data['vehicle_speed'][ind] == 0.0:
             ind = ind + 1
         else:
             break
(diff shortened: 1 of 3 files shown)


modules/tools/calibration/process.py
```diff
@@ -48,12 +48,12 @@ def get_start_index(data):
...
 
-    start_ind = np.where(data['brake_percentage'] == 40)[0][0]
+    start_ind = np.where(data['vehicle_speed'] == 0.0)[0][0]
 
     ind = start_ind
     while ind < len(data):
-        if data['brake_percentage'][
-                ind] == 40:  #or data['vehicle_speed'][ind] == 0.0:
+        if data['vehicle_speed'][
+                ind] == 0.0:  #or data['vehicle_speed'][ind] == 0.0:
```
xiaoxq: Put comment in another line to make the code easy to read.

modules/tools/calibration/process_data.sh
```diff
@@ -20,7 +20,7 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}"  )" && pwd  )"
 rm $DIR/result.csv
-for f in `ls ${1}/*_recorded.csv`
+for f in `ls ${1}/*r0.csv`
```
xiaoxq: Avoid single-char variable name.

modules/tools/calibration/process.py
```diff
@@ -48,12 +48,12 @@ def get_start_index(data):
     if np.all(data['vehicle_speed'] == 0):
         return 0
 
-    start_ind = np.where(data['brake_percentage'] == 40)[0][0]
+    start_ind = np.where(data['vehicle_speed'] == 0.0)[0][0]
```
xiaoxq: Not safe to compare `equal` on float numbers.

wfgsy101: modified as your require

modules/tools/calibration/process_data.sh
wfgsy101: has been changed to t*r0.csv

PR #7688, opened 2019-04-05
tools: mapshow added function to plot x,y coord.

yifeijiang:


modules/tools/mapshow/mapshow.py
```diff
@@ -98,5 +102,9 @@ def draw(map):
         localization.load(args.loc)
         localization.plot_vehicle(plt)
 
+    if args.position is not None:
+        x = float(args.position.split(",")[0])
+        y = float(args.position.split(",")[1])
```
xiaoxq: Split once and reuse. Eg:
`x, y = args.position.split(",")`
`x, y = float(x), float(y)`

yifeijiang: Thanks for pointing this out! Changing it now.

PR #7689, opened 2019-04-05
Planning: use self path when speed fail in side pass

JasonZhou404:


modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
```diff
@@ -31,6 +31,7 @@
 #include "modules/planning/tasks/deciders/speed_bounds_decider/speed_limit_decider.h"
 #include "modules/planning/tasks/deciders/speed_bounds_decider/st_boundary_mapper.h"
 
+#define ADEBUG AERROR
```
panjiacheng: This is for debugging purpose only. Should remove this line in the commit because otherwise the DreamLand will fail due to memory limits.

modules/planning/tasks/deciders/path_assessment_decider.cc
```diff
@@ -63,8 +63,8 @@ Status PathAssessmentDecider::Process(
   // 1. Remove invalid path.
   std::vector<PathData> valid_path_data;
   for (const auto& curr_path_data : candidate_path_data) {
-    // RecordDebugInfo(curr_path_data, curr_path_data.path_label(),
-    //                 reference_line_info);
+    RecordDebugInfo(curr_path_data, curr_path_data.path_label(),
+                    reference_line_info);
```
panjiacheng: This is for debugging purpose. Should comment this part in the commit. Otherwise, there will be too many paths printed on the simulator.

modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
JasonZhou404: Ok

modules/planning/tasks/deciders/path_assessment_decider.cc
JasonZhou404: ok

PR #7696, opened 2019-04-05
Add install_fast_rtp.sh installer and patch file to set up fast_rtp i…

kyleApollo:
…n docker env.


docker/build/installers/FastRTPS_1.5.0.patch
```diff
@@ -0,0 +1,1124 @@
+diff -Npur -X ex.file a/cmake/dev/check_configuration.cmake b/cmake/dev/check_configuration.cmake
```
xiaoxq: Patch is hard to maintain. Please make a fork of the upstream repo, submit the code into the fork, and refer to it from Apollo. So that everybody can see what is changed, and it's easy to rebase to newer version of upstream.

PR #7711, opened 2019-04-07
revise discrete matrix B

purewater0901:
### Discrete Matrix Error in `apollo/modules/control/controller/mpc_controller.cc`  

I think apollo use trapezoidal integration to discrete MPC model matrix.  
This method should be like following equations.  

* original equations  
(dx/dt) = Ax + Bu   

* discrete equations  
(x(k+1)-x(k))/&delta;t = A * (x(k+1) + x(k)) / 2 +Bu  
  
x(k+1) = (I-A*&delta;t/2)<sup>-1</sup>*(I+A*&delta;t/2)*x(k) + (I-A*&delta;t/2)<sup>-1</sup> * B * &delta;t * u  
  
therefore, the discrete matrix will be 

Ad = (I-A*&delta;t/2)<sup>-1</sup>*(I+A*&delta;t/2)  
Bd = (I-A*&delta;t/2)<sup>-1</sup> * B * &delta;t     

However, discrete matrix B  in `apollo/modules/control/controller/mpc_controller.cc`  is  
Bd = B * &delta;t     

Please check the discrete matrix.  I think it should be revised.


Capri2014:
@purewater0901 Thanks for the PR, and sorry for the confusion. --- Previously we were lazy (mainly me) and only apply the bilinear transformation to matrix A and the zero order hold for matrix B for the approximation (and simply to keep it stochastic with respect to u(t) as well). (Since A is more sensitive to with respect to system parameter change.) -- This might be the root cause of all the confusion. We will look into it and keep you updated.

purewater0901:
@Capri2014 Thanks for your replying. I am looking forward from hearing from you.

PR #7716, opened 2019-04-07
planning: open space plot fallback trajecotry for debug

deidaraho:


modules/planning/common/open_space_info.cc
```diff
@@ -64,6 +64,16 @@ void OpenSpaceInfo::RecordDebug(apollo::planning_internal::Debug* ptr_debug) {
...
       ->mutable_open_space()
       ->set_is_fallback_trajectory(fallback_flag_);
+
+  // 5, record fallback trajectory if needed
+  if (fallback_flag_) {
+    auto* ptr_fallback_trajectory = ptr_debug->mutable_planning_data()
+                                    ->mutable_open_space()
+                                    ->mutable_fallback_trajectory()
+                                    ->add_trajectory();
+    auto fallback_trajectory = fallback_trajectory_.first;
```
xiaoxq: `const auto&`?

deidaraho: > `const auto&`?

fixed

PR #7735, opened 2019-04-08
apollo.sh experimental_multi_threaded_digest

Fla3inH0tCheet0s:
enable bazel's feature to compute md5 checksums in parallel.

`docker/scripts/dev_start.sh` brings in over a GB of map data. Then, the maps are globbed and turned into a `file_group` in https://github.com/ApolloAuto/apollo/blob/master/modules/map/BUILD#L3. bazel server will compute the md5 sums and keep it in memory the first time `bazel build` or `bazel test` runs on `modules/map`. Digesting these files takes longer as file sizes grow.

Running in parallel is much faster and doesn't seem to cause any problems.

### DIFF
--- a/apollo.sh
+++ b/apollo.sh
@@ -698,6 +698,7 @@ function main() {
   check_esd_files
 
   DEFINES="--define ARCH=${MACHINE_ARCH} --define CAN_CARD=${CAN_CARD} --cxxopt=-DUSE_ESD_CAN=${USE_ESD_CAN}"
+  # Enable bazel's feature to compute md5 checksums in parallel
   DEFINES="${DEFINES} --experimental_multi_threaded_digest"
 
   if [ ${MACHINE_ARCH} == "x86_64" ]; then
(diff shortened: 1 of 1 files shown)


apollo.sh
```diff
@@ -698,6 +698,7 @@ function main() {
   check_esd_files
 
   DEFINES="--define ARCH=${MACHINE_ARCH} --define CAN_CARD=${CAN_CARD} --cxxopt=-DUSE_ESD_CAN=${USE_ESD_CAN}"
+  DEFINES="${DEFINES} --experimental_multi_threaded_digest"
```
xiaoxq: Thanks for the contribution!
1. Could you add some comment to explain this flag?
2. Have you tried and evaluated the speeding up?

xiaoxq: I mean, add comment in code. It could be similar to the pull-request explanation above.

Fla3inH0tCheet0s: 1. I pushed a new commit. If you want, I can squash it.
2. The travis CI build times are noisy. The best benchmarks can be best seen on an already built project *without* changes. My processor runs at 4 Ghz. It has 4 physical cores and 4 virtual cores. After I run `docker/scripts/dev_start.sh` on 6d159867dab292373cdd272655b6d5ee6fba8e4a, it takes 146 seconds to run `apollo.sh build`. After I ran `docker/scripts/dev_start.sh` on my patch, it took 55 seconds `apollo.sh build`.

PR #7738, opened 2019-04-09
planning: open space fallback, bug fix

deidaraho:
fallback trajectory's start point issue fixed

### DIFF
--- a/modules/planning/tasks/deciders/open_space_fallback_decider.cc
+++ b/modules/planning/tasks/deciders/open_space_fallback_decider.cc
@@ -33,18 +33,19 @@ OpenSpaceFallbackDecider::OpenSpaceFallbackDecider(const TaskConfig& config)
 
 Status OpenSpaceFallbackDecider::Process(Frame* frame) {
   std::vector<std::vector<common::math::Box2d>> predicted_bounding_rectangles;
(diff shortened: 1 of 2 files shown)


modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -55,63 +56,40 @@ Status OpenSpaceFallbackDecider::Process(Frame* frame) {
...
-          ptr_fallback_trajectory_pair->first[0].path_point();
-      for (size_t i = 0; i < temp_horizon; ++i) {
-        ptr_fallback_trajectory_pair->first[i].set_v(0.0);
+        collision_point.path_point().s() - previous_point.path_point().s();
+    double stopping_distance = relative_collision_distance;
+    // the accelerate = -v0^2 / (2*s), where s is slowing down distance
+    size_t temp_horizon =
+        frame_->open_space_info().fallback_trajectory().first.NumOfPoints();
+    const double accelerate =
+        -std::abs(previous_point.v()) * previous_point.v()
```
JasonZhou404: why std::abs here

deidaraho: > why std::abs here

The accelerate is the -sign(current speed), so it is - |v| * v / (2 * s)

PR #7743, opened 2019-04-09
Planning: initial implementation for lane-change urgency checking

JasonXu12:
Current logic is simply stopping and waiting. Future implementation will be based on this.

### DIFF
--- a/modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
+++ b/modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
@@ -92,6 +92,11 @@ Status SpeedBoundsDecider::Process(
     mutable_side_pass_info->change_lane_stop_flag = set_stop_fence;
   }
 
+  // Check the lane change urgency at current frame
+  if (FLAGS_enable_lane_change_urgency_checking) {
+    CheckLaneChangeUrgency(frame);
+  }
+
   // 2. Map obstacles into st graph
   StBoundaryMapper boundary_mapper(adc_sl_boundary, speed_bounds_config_,
                                    reference_line, path_data,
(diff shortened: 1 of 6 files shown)


modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
```diff
@@ -159,6 +164,51 @@ Status SpeedBoundsDecider::Process(
...
+              if (distance_to_passage_end < \
+                  speed_bounds_config_.approach_distance_for_lane_change()) {
+                // Set a stop fence and wait for chances to change lane
+                const std::string stop_wall_id = "lane_change_stop";
+                std::vector<std::string> wait_for_obstacles;
+                DeciderRuleBasedStop::BuildStopDecision(
+                       stop_wall_id, sl_point.s(),
+                       speed_bounds_config_.urgent_distance_for_lane_change(),
+                       StopReasonCode::STOP_REASON_LANE_CHANGE_URGENCY,
+                       wait_for_obstacles, frame, &reference_line_info);
```
xiaoxq: Try to reduce indents. Some practices:

1. You can check the negative logic and call `continue`.
```
if (is_clear_to_change_lane_) {
  continue;
}
// Other logic.
```

2. You can wrap inside logic into function to make it more readable:
```
if (!is_clear_to_change_lane_) {
  DealWithIt();
}
```

modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
```diff
@@ -159,6 +164,51 @@ Status SpeedBoundsDecider::Process(
...
+        is_clear_to_change_lane_ = \
+            ChangeLaneDecider::IsClearToChangeLane(&reference_line_info);
+      } else {
+        // If lane change is blocked
+        if (!is_clear_to_change_lane_) {
+          // Get the end point of current routing
+          const auto& route_end_waypoint = \
+          reference_line_info.Lanes().RouteEndWaypoint();
+          if (route_end_waypoint.lane) {
+            auto point = \
```
xiaoxq: No need to use `\` in C++.

modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
JasonXu12: Solved. Thanks!

modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
JasonXu12: Solved. Thanks!

PR #7752, opened 2019-04-09
Planning: skip side pass stop with self path

JasonZhou404:


modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
```diff
@@ -208,6 +151,76 @@ Status SpeedBoundsDecider::Process(
...
+            reference_line_info);
+      }
+      if (side_pass_info.check_clear_flag &&
+          CheckClearDone(*reference_line_info,
+                         side_pass_info.change_lane_stop_path_point)) {
+        mutable_side_pass_info->check_clear_flag = false;
+      }
+    }
+    mutable_side_pass_info->change_lane_stop_flag = set_stop_fence;
+}
```
xiaoxq: 1. Mis-aligned brackets.
2. Refactor code to `return` early and reduce nested `if`, `else`.

modules/planning/tasks/deciders/speed_bounds_decider/speed_bounds_decider.cc
```diff
@@ -60,64 +60,7 @@ Status SpeedBoundsDecider::Process(
...
-            reference_line_info);
-      }
-      if (side_pass_info.check_clear_flag &&
-          CheckClearDone(*reference_line_info,
-                         side_pass_info.change_lane_stop_path_point)) {
-        mutable_side_pass_info->check_clear_flag = false;
-      }
-    }
-    mutable_side_pass_info->change_lane_stop_flag = set_stop_fence;
+      StopOnSidePass(frame, reference_line_info);
```
xiaoxq: Seems like this is the miss-indented code.

PR #7755, opened 2019-04-09
planning: fix open space fallback shaking issue

deidaraho:


modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -47,7 +47,7 @@ Status OpenSpaceFallbackDecider::Process(Frame* frame) {
           predicted_bounding_rectangles, &current_idx, &first_collision_idx)) {
     // change gflag
     frame_->mutable_open_space_info()->set_fallback_flag(true);
-
+    AERROR << "aaaaaaaaaa";
```
Capri2014: please clean this up before doing a pr

Capri2014:
Which case this expect to fix?

deidaraho:
> Which case this expect to fix?

Scenario 562. when stopped by fallback trajectory, the vehicle is shaking at the end point.

modules/planning/tasks/deciders/open_space_fallback_decider.cc
```diff
@@ -72,16 +71,14 @@ Status OpenSpaceFallbackDecider::Process(Frame* frame) {
...
 
-      if (ptr_fallback_trajectory_pair->first[i].relative_time() >
-          relative_stopping_time) {
-        ptr_fallback_trajectory_pair->first[i].mutable_path_point()->set_x(
-            previous_point.path_point().x());
-        ptr_fallback_trajectory_pair->first[i].mutable_path_point()->set_y(
-            previous_point.path_point().y());
+      if (ptr_fallback_trajectory_pair->first[i].relative_time()
+          > relative_stopping_time) {
+        double tmp_rt = ptr_fallback_trajectory_pair->first[i].relative_time();
```
xiaoxq: Put it above so you can use it in the `if`

PR #7773, opened 2019-04-10
Tools: update data extractor. 

gchen-apollo:
(1) switch to protobuf text_format as configuration handler 
    (2) add configuration example for sensor data extraction   
    (3) enable simple one-line command
    (4) save point timestamp in second in pcd


modules/tools/sensor_calibration/proto/extractor_config.proto
```diff
@@ -0,0 +1,46 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
freeHackOfJeff: Copyright 2019, and better to embrace our copyright as usual format: '#' :)

PR #7784, opened 2019-04-10
Dreamview: guard against nan ego car pose

siyangy:


modules/dreamview/frontend/src/renderer/index.js
```diff
@@ -349,7 +349,13 @@ class Renderer {
     }
 
     updateWorld(world) {
-        this.adc.update(this.coordinates, world.autoDrivingCar);
+        const adcPose = world.autoDrivingCar;
+        this.adc.update(this.coordinates, adcPose);
+        if (isNaN(adcPose.positionX) || isNaN(adcPose.positionY)) {
```
vlin17: isNaN(null) returns false. Do we want to check number using lodash's _.isNumber() instead?

PR #7785, opened 2019-04-10
Docs: memory requirement update

natashadsouza:


martins-mozeiko:
Seeing "HeapDumpOnOutOfMemoryError" during building is normal with bazel. It does not mean machine runs out of memory. "HeapDumpOnOutOfMemoryError" is just an argument bazel puts on commandline for all bazel invocations.

natashadsouza:
> Seeing "HeapDumpOnOutOfMemoryError" during building is normal with bazel. It does not mean machine runs out of memory. "HeapDumpOnOutOfMemoryError" is just an argument bazel puts on commandline for all bazel invocations.

Fixed in latest PR. @martins-mozeiko thanks for the input. Please verify the current PR

PR #7786, opened 2019-04-10
Planning: openspace: minor improvements

JasonZhou404:


modules/planning/tasks/optimizers/open_space_trajectory_partition/open_space_trajectory_partition.cc
```diff
@@ -303,7 +303,7 @@ void OpenSpaceTrajectoryPartition::PartitionTrajectory(
     }
   }
   // Partition trajectory points into each trajectory
-  constexpr double kGearShiftEpsilon = 0.5;
+  constexpr double kGearShiftEpsilon = 0.0;
```
xiaoxq: A zero epsilon seems useless?

PR #7787, opened 2019-04-10
planning: planning_path_debug branch merge back to master

YajiaZhang:
planning_path_debug branch merge back to master with new feature switched off.


modules/planning/conf/planning.conf
```diff
@@ -14,6 +14,8 @@
 --default_cruise_speed=20.00
 --enable_scenario_side_pass
 --enable_scenario_traffic_light
+--enable_collision_detection
```
Capri2014: @YajiaZhang This flag has been deprecated via commit https://github.com/ApolloAuto/apollo/commit/1eff2859c13803037d62599cb14415a0f6b4f066  Please change accordingly.

YajiaZhang: Done.

modules/planning/math/piecewise_jerk/BUILD
```diff
@@ -3,12 +3,12 @@ load("//tools:cpplint.bzl", "cpplint")
 package(default_visibility = ["//visibility:public"])
 
 cc_library(
-    name = "fem_1d_qp_problem",
+    name = "piecewise_jerk_problem",
```
Capri2014: If we decided to fully deprecate "fem_1d_qp_problem" files, a better way is to remove the corresponding source files as well. --- If the depreciation is depending on the test result and just to change the default as "piecewise_jerk_problem", I would recommend having them both in compile and test and just use a flag to switch between which one you use. --- Either way works, but if we only remove the "piecewise_jerk_problem" from BUILD but not their source files as well. But we may accidentally leave dead code of fem_1d_qp_problem.cc/h without protection by compilation or test.

modules/planning/math/piecewise_jerk/BUILD
```diff
@@ -18,16 +18,28 @@ cc_library(
...
     ],
     deps = [
-        ":fem_1d_qp_problem",
+        ":piecewise_jerk_problem",
         "//cyber/common:log",
         "@gtest//:main",
     ],
 )
 
+cc_library(
```
Capri2014: Was the test of "fem_1d_qp_problem" removed?

xiaoxq:
Per discussion, do merge commit.

xiaoxq:
Per discussion, do merge-commit on this pull request to keep history.

PR #7792, opened 2019-04-11
Planning: add path_time_qp problem

HongyiSun:


panjiacheng:
There is a typo with the last commit: It should start with "Planning" rather than "Prediction".

HongyiSun:
> There is a typo with the first commit: It should start with "Planning" rather than "Prediction".

Fixed, thank you!

PR #7824, opened 2019-04-12
Planning: enable GenerateFallbackSpeed

HongyiSun:


modules/planning/common/speed_profile_generator.cc
```diff
@@ -133,8 +133,8 @@ SpeedData SpeedProfileGenerator::GenerateFallbackSpeed(
     common::VehicleConfigHelper::GetConfig().vehicle_param();
 
   std::array<double, 3> init_s = {0.0, init_v, init_a};
-  std::array<double, 3> end_s = {0.0, 0.0, 0.0};
-  std::array<double, 5> w = {1.0, 0.0, 10.0, 100.0, 0.0};
+  std::array<double, 3> end_s = {stop_distance, 0.0, 0.0};
+  std::array<double, 5> w = {10000.0, 0.0, 1.0, 0.01, 0.0};
```
Capri2014: Make this weight a config loaded from conf.  or at least mark a TODO here. Since this might need to be re-tuned when you encounter other cases.

HongyiSun: add a TODO here, thank you!

PR #7847, opened 2019-04-12
planning: more in stage_approach of BareIntersection scenario

jmtao:


modules/planning/scenarios/util/util.cc
```diff
@@ -60,6 +62,65 @@ hdmap::PathOverlap* GetOverlapOnReferenceLine(
...
+    Frame* const frame,
+    ReferenceLineInfo* const reference_line_info) {
+  CHECK_NOTNULL(frame);
+  CHECK_NOTNULL(reference_line_info);
+
+  // check
+  const auto& reference_line = reference_line_info->reference_line();
+  if (!WithinBound(0.0, reference_line.Length(), stop_line_s)) {
+    AERROR << "stop_line_s[" << stop_line_s << "] is not on reference line";
+    return 0;
```
xiaoxq: Return bool.

jmtao: good catch. thx. amended

PR #7861, opened 2019-04-15
Prediction: add lane sequence id into offline feature for feature & l…

kechxu:
…abel matching


modules/prediction/common/feature_output.h
```diff
@@ -59,7 +59,8 @@ class FeatureOutput {
    */
   static void InsertDataForLearning(const Feature& feature,
                                     const std::vector<double>& feature_values,
-                                    const std::string& category);
+                                    const std::string& category,
+                                    const LaneSequence* lane_sequence_ptr);
```
HongyiSun: Better to assign a default nullptr here?

PR #7871, opened 2019-04-15
Common: update vehicle param to include vehicle brand, Planning: update trajectory partition chart in open space planner

Capri2014:


modules/common/data/gem_config.pb.txt
```diff
@@ -1,4 +1,5 @@
 vehicle_param {
+  brand: GEM
```
xiaoxq: Would it be possible to move the config to `modules/calibration/data/<vehicle_id>/` as it looks like per-vehicle configuration instead of global common data?

PR #7892, opened 2019-04-16
Planning: fix and add longitudinal_jerk to gflag

HongyiSun:


modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
```diff
@@ -416,7 +416,7 @@ void PiecewiseJerkProblem::CalculateOffset(std::vector<c_float>* q) {
   q->resize(kNumParam);
   for (int i = 0; i < kNumParam; ++i) {
     if (i < N) {
-      q->at(i) = -1.0 * weight_.x_mid_line_w *
+      q->at(i) = -1.0 * weight_.x_ref_w *
                  (std::get<0>(x_bounds_[i]) + std::get<1>(x_bounds_[i]));
```
xiaoxq: How about using `std::tie` to extract the elements and shorten code?

PR #7895, opened 2019-04-16
planning: adjust max_valid_stop_distance conf for stop-sign scenario

jmtao:


modules/planning/scenarios/BUILD
```diff
@@ -24,9 +24,11 @@ cc_library(
     name = "stage",
     srcs = [
         "stage.cc",
+        "stage_intersection_cruise.cc",
     ],
     hdrs = [
         "stage.h",
+        "stage_intersection_cruise.h",
```
xiaoxq: Split to its own target?

modules/planning/scenarios/stage_intersection_cruise.h
```diff
@@ -0,0 +1,54 @@
...
+        right_of_way_status_(right_of_way_status) {}
+
+ private:
+  Stage::StageStatus Process(const common::TrajectoryPoint& planning_init_point,
+                             Frame* frame) override;
+
+ private:
+  Stage::StageStatus FinishStage();
+
+ private:
```
xiaoxq: Leaving one `private:` is OK.

modules/planning/scenarios/BUILD
jmtao: sure. will fix it in another commit.

modules/planning/scenarios/stage_intersection_cruise.h
jmtao: will be removed in my next commit

PR #7909, opened 2019-04-17
Planning: openspace: reduce using CHECK

JasonZhou404:


modules/planning/open_space/coarse_trajectory_generator/node3d.cc
```diff
@@ -63,27 +63,30 @@ Node3d::Node3d(const std::vector<double>& traversed_x,
                const PlannerOpenSpaceConfig& open_space_conf) {
   CHECK(XYbounds.size() == 4)
       << "XYbounds size is not 4, but" << XYbounds.size();
+  DCHECK(traversed_x.size() == traversed_y.size() &&
+         traversed_x.size() == traversed_phi.size())
+      << "input pose data not in same size";
```
xiaoxq: 1. Call CHECK separately, otherwise you don't know what's wrong when it's wrong.
2. Use CHECK_EQ.

PR #7913, opened 2019-04-17
Control: address the backward driving stuck issue due to rough calib …

kevin-y-wang:
…table


Yuntianyi-Chen:
What is the backward driving stuck issue here? Does it happen when backing a car?

PR #7929, opened 2019-04-18
Continuous integration: cat the detail log when lint meet error

mickeyouyou:


mickeyouyou:
> 1. Plz check the output. It's too noisy and hard to find the failure logs.
> 2. It's not good practice to find lint error in CI. It's developer's responsibility to run `apollo.sh lint` before submitting any piece of code. We should save CI resource because it's shared and limited.

thanks xiangquan. 

Ii's better if touch the error detail

jinghaomiao:
lint check is done locally. We'd avoid checking details on the remote server if CI is deployed on travis.

mickeyouyou:
image

PR #7933, opened 2019-04-18
Planning: Pass scala types by value per best practice.

xiaoxq:
https://github.com/xiaoxq/apollo/blob/master/docs/technical_tutorial/best_coding_practice.md
Point 5


xiaoxq:
@jmtao https://stackoverflow.com/questions/20982042/passing-scalar-types-by-value-or-reference-does-it-matter

PR #7936, opened 2019-04-18
Planning: various changes with side-pass

panjiacheng:
See every commit's name for details.


panjiacheng:
The new architecture for the new side-pass is still being implemented. Therefore, this one still uses the old side-pass for now.

PR #7939, opened 2019-04-18
Docs: links

natashadsouza:


docs/specs/qp_spline_path_optimizer.md
```diff
@@ -18,25 +18,25 @@ Split the path into **n** segments. each segment trajectory is defined by a poly
 
 Each segment ***i*** has accumulated distance $d_i$ along reference line. The trajectory for the segment is defined as a polynomial of degree five by default.
 
-<p>
+```
```
mickeyouyou: Formulas will parse failed.

`p` is html element, that contains the tex start and end with `$$`, that will not render again but can parse by TEX plugin.

natashadsouza: @mickeyouyou these are documents, and the format for writing code in docs is ```code```

PR #7940, opened 2019-04-18
Control: vehicle state and trajectory transformation for reverse driving

kevin-y-wang:
…iving

### DIFF
--- a/modules/common/vehicle_state/vehicle_state_provider.cc
+++ b/modules/common/vehicle_state/vehicle_state_provider.cc
@@ -234,10 +234,8 @@ math::Vec2d VehicleStateProvider::EstimateFuturePosition(const double t) const {
 math::Vec2d VehicleStateProvider::ComputeCOMPosition(
     const double rear_to_com_distance) const {
   // set length as distance between rear wheel and center of mass.
-  // TODO(Yu): the vehicle state transformation will be synconized with
-  // the planning reference transformation
   Eigen::Vector3d v;
-  if (FLAGS_reverse_localization_transform &&
+  if (!FLAGS_coordinate_transform_to_com &&
       vehicle_state_.gear() == canbus::Chassis::GEAR_REVERSE) {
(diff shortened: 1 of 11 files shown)


Capri2014:
@kevin-y-wang Just add TODO here to deprecate in future

kevin-y-wang:
Update the commit with "TODO" for explaining the strategy

kevin-y-wang:
Modifications finished per comment. Thanks.

kevin-y-wang:
Another new commit is added into the pull request

modules/common/vehicle_state/vehicle_state_provider.cc
```diff
@@ -234,7 +234,15 @@ math::Vec2d VehicleStateProvider::EstimateFuturePosition(const double t) const {
...
-  Eigen::Vector3d v(0.0, rear_to_com_distance, 0.0);
+  // TODO(Yu): the vehicle state transformation will be synconized with
+  // the planning reference transformation
+  Eigen::Vector3d v;
+  if (FLAGS_reverse_localization_transform &&
+      vehicle_state_.gear() == canbus::Chassis::GEAR_REVERSE) {
+    v << 0.0, 0.0, 0.0;
+  } else {
+    v << 0.0, rear_to_com_distance, 0.0;
+  }
```
JasonZhou404: What you mean by planning reference transformation. If transform planning reference, there is no need to change com in vehicle state provider

kevin-y-wang: If currently we transform the planning reference to the COM, then all the control gains for forward driving have to be re-tuned. It may be conflict with all the current tests. 
With respect to your suggestion, I may consider to only transform the planning reference when backward driving (currently no one but us try the backward driving), to see if it produces the better effect.

kevin-y-wang:
Synchronize the vehicle_state_provider and trajectory analyzer for location transformation between center of rear-axis and COM

PR #7944, opened 2019-04-19
framework: make SchedulerChoreography constructor private & modify un…

zhangchaoltt:
…it test


cyber/scheduler/scheduler_choreo_test.cc
```diff
@@ -0,0 +1,83 @@
+/******************************************************************************
+ * Copyright 2018 The Apollo Authors. All Rights Reserved.
+ *
```
jinghaomiao: 2019?

zhangchaoltt: done

PR #7958, opened 2019-04-20
modify dynamics_model

purewater0901:
###  Dynamics Model Calculation  

In Dynamics Model, it would use ref_heading_rate instead of heading_error_rate. I found apollo said reference [following book](https://www.springer.com/jp/book/9781461414322). This book also uses ref_heading_rate.


Capri2014:
@purewater0901 The dynamic model is building around the "nominal" dynamics and therefore utilizing the "error" state as state instead of  "actual" state.

purewater0901:
@Capri2014 Thank you for your replying.   

>The dynamic model used in Apollo is building around the "nominal" dynamics and therefore utilizing the "error" state as state instead of "actual" state.

I understood you use error state as state, but apollo's state equation looks like

  
  
 
 
is lateral and yaw error.   

Due to 
 
matrix A and Matrix C in the state equation should be incorporated.

Also, [the book](https://www.springer.com/jp/book/9781461414322) uses

PR #7983, opened 2019-04-22
Dreamview: add signal and stopsign for offlineview

kiboliu:


modules/dreamview/frontend/src/renderer/tileground.js
```diff
@@ -58,15 +80,44 @@ export default class TileGround {
...
+                    mesh.position.set(itemPos.x, itemPos.y, itemPos.z);
+                    if (item.id) {
+                        // store signal id as name in mesh
+                        mesh.name = item.id;
+                    }
+                    mesh.matrixAutoUpdate = false;
+                    mesh.updateMatrix();
+
+                    // a tile may have multiple signal or stopsign
+                    currentItems[`${key}_${item.x}_${item.y}`] = mesh;
```
vlin17: How come we want to include x,y? Key is not unique?

modules/dreamview/frontend/src/renderer/tileground.js
```diff
@@ -22,6 +40,8 @@ export default class TileGround {
 
         this.hash = -1;
         this.currentTiles = {};
+        this.currentSignal = {};
+        this.currentStopsign = {};
```
vlin17: currentStopSign

modules/dreamview/frontend/src/renderer/tileground.js
```diff
@@ -1,7 +1,25 @@
...
+    y: TRAFFIC_LIGHT_SCALE,
+    z: TRAFFIC_LIGHT_SCALE
+};
+
+const STOP_SIGN_SCALE = 0.01;
+const stopSignScales = {
+    x: STOP_SIGN_SCALE,
+    y: STOP_SIGN_SCALE,
+    z: STOP_SIGN_SCALE
+};
```
vlin17: These constants are used in map.js as well, is it possible to consolidate them?

modules/dreamview/frontend/src/renderer/tileground.js
kiboliu: For same row and col, multiple signals can exists, thus add x,y for key.

PR #7986, opened 2019-04-22
Perception: fill lane detection results into output channel: /perception/obstacles

yg13:
The fitting parameters c0, c1, c2, c3 lies in car coordinates: x forward, y left, z up

### DIFF
--- a/modules/perception/onboard/component/fusion_camera_detection_component.cc
+++ b/modules/perception/onboard/component/fusion_camera_detection_component.cc
@@ -35,6 +35,16 @@ namespace onboard {
 
 using apollo::cyber::common::GetAbsolutePath;
 
+static void fill_lane_msg(apollo::perception::LaneMarker *lane_marker,
+                          const base::LaneLineCubicCurve &curve_coord) {
+  lane_marker->set_c0_position(curve_coord.d);
+  lane_marker->set_c1_heading_angle(curve_coord.c);
+  lane_marker->set_c2_curvature(curve_coord.b);
+  lane_marker->set_c3_curvature_derivative(curve_coord.a);
+  lane_marker->set_longitude_start(curve_coord.x_start);
+  lane_marker->set_longitude_end(curve_coord.x_end);
+}
+
 static int GetGpuId(const camera::CameraPerceptionInitOptions &options) {
   camera::app::PerceptionParam perception_param;
   std::string work_root = "";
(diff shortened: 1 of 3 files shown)


Capri2014:
@yg13 Do we plan to use it in standard mode or highway mode? Just an FYI that it would be no problem in highway mode but hard to use by PnC in standard mode as the lane in Vehicle FLU coordinate system.

modules/perception/onboard/component/fusion_camera_detection_component.cc
```diff
@@ -798,6 +804,56 @@ int FusionCameraDetectionComponent::MakeProtobufMsg(
...
+
+  // write out lanes in ego coordinates
+  apollo::perception::LaneMarkers *lane_markers =
+      obstacles->mutable_lane_marker();
+  for (const auto &lane : lane_objects) {
+    base::LaneLineCubicCurve curve_coord = lane.curve_image_coord;
+    if (lane.curve_car_coord.a == NULL)
+      curve_coord = lane.curve_car_coord;
+
+    if (lane.pos_type == base::LaneLinePositionType::EGO_LEFT) {
```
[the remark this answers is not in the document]
yg13: I actually tried switch cases first. But because the pose_type includes 12 classes, and code will not compile without listing all the cases. So I think it might be more concise to use if/else here.

yg13:
@Capri2014 I see your point. FLU representation might not be used directly for PNC. I think c0, c1, and x_start, x_end can be converted to world coordinates directly and has definite meanings to use. However for c2, c3, I am not sure. If the whole curve shape is actually needed in world coordinate system, a polyline representation can be converted from ego to world. 

Maybe for this PR, I just keep current format. If any representation in world coordinates is required I can add these in a separate PR later.

modules/perception/onboard/component/fusion_camera_detection_component.cc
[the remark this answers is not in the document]
xiaoxq: You may need to add `default:` to cover types not listed.

modules/perception/onboard/component/fusion_camera_detection_component.cc
```diff
@@ -35,6 +35,16 @@ namespace onboard {
 
 using apollo::cyber::common::GetAbsolutePath;
 
+static void fill_lane_msg(apollo::perception::LaneMarker *lane_marker,
+                          const base::LaneLineCubicCurve &curve_coord) {
```
xiaoxq: Put input parameters before output parameters.

modules/perception/onboard/component/fusion_camera_detection_component.cc
```diff
@@ -798,6 +812,42 @@ int FusionCameraDetectionComponent::MakeProtobufMsg(
...
+  apollo::perception::LaneMarker *lane_marker_r0 =
+      lane_markers->mutable_right_lane_marker();
+  apollo::perception::LaneMarker *lane_marker_l1 =
+      lane_markers->add_next_left_lane_marker();
+  apollo::perception::LaneMarker *lane_marker_r1 =
+      lane_markers->add_next_right_lane_marker();
+
+  for (const auto &lane : lane_objects) {
+    base::LaneLineCubicCurve curve_coord = lane.curve_image_coord;
+    if (&lane.curve_car_coord.a != nullptr) {
```
xiaoxq: This seems true forever.

yg13: Can you suggest a way to check whether this struct has been assigned values?

yg13: Never mind, I will modify lane detection to make sure there are always valid outputs in curve_car_coord

PR #7987, opened 2019-04-22
Planning: refactor on speed fallback in lane follow stage

JasonZhou404:


JasonZhou404:
stable test set passes without increasing failure rate. 
no error cases in stable test as well.

PR #7993, opened 2019-04-23
tools: update canbus files when add new vehicle

mickeyouyou:


modules/tools/gen_vehicle_protocol/template/controller.cc.tpl
```diff
@@ -1,5 +1,6 @@
 /******************************************************************************
- * Copyright 2018 The Apollo Authors. All Rights Reserved.
+<<<<<<< HEAD
```
Capri2014: Please clean this up.

mickeyouyou: thanks, done.

mickeyouyou:
so many noisy commits will reopen a new one.

PR #8000, opened 2019-04-23
Canbus: add the new micro car Zhongyun

macDure:
Add the new micro car Zhongyun. Here are the Zhongyun canbus source code. The code had been checked, tested, and linted.


modules/canbus/proto/chassis_detail.proto
```diff
@@ -35,6 +36,7 @@ message ChassisDetail {
   optional Transit transit = 20;
   optional Ge3 ge3 = 21;
   optional Wey wey = 22;
```
Capri2014: this could cause problem if we merge coolhigh back to master right? Should we reserve a field for Coolhigh?

macDure: I found the r3.0 chassis_detail ch=19, https://github.com/ApolloAuto/apollo/blob/r3.0.0/modules/canbus/proto/chassis_detail.proto, but the master chassis_detail has increased number to 22, so coolhigh is behind of master number. As we discuss, we keep the master number in order.

PR #8022, opened 2019-04-24
Planning: wrap up the new side-pass implementation and enable it.

panjiacheng:


modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -535,6 +532,42 @@ bool PathBoundsDecider::GetBoundaryFromLanesAndADC(
...
+  // Sanity checks.
+  CHECK_NOTNULL(path_bound);
+  CHECK(!path_bound->empty());
+
+  // Go through every point on reference_line, and take care those that
+  // deviate from lane-center line considerably.
+  //  - Expand lane boundary on one side. (already done previously)
+  //  - Treat the other side as if there is an obstacle.
+  for (size_t i = 1; i < path_bound->size(); ++i) {
+  }
```
xiaoxq: Remove incomplete code.

modules/planning/tasks/optimizers/path_decider/path_decider.cc
```diff
@@ -173,15 +173,17 @@ bool PathDecider::MakeStaticObstacleDecision(
                                                obstacle->Id(), object_decision);
       }
     } else if (FLAGS_enable_nudge_decision) {
-      // nudge
-      if (curr_l > sl_boundary.end_l()) {
+      // 3. NUDGE if laterally very close.
+      if (sl_boundary.end_l() < curr_l - min_nudge_l) {  // &&
+          // sl_boundary.end_l() > curr_l - min_nudge_l - 0.3) {
```
xiaoxq: Refine the comments, tidy and serious.

modules/planning/tasks/deciders/path_lane_borrow_decider/path_lane_borrow_decider.cc
```diff
@@ -143,10 +143,10 @@ bool PathLaneBorrowDecider::IsBlockingObstacleFarFromIntersection(
       reference_line_info.FirstEncounteredOverlaps();
   for (const auto& overlap : first_encountered_overlaps) {
     ADEBUG << overlap.first << ", " << overlap.second.DebugString();
-    if (overlap.first != ReferenceLineInfo::CLEAR_AREA &&
-        overlap.first != ReferenceLineInfo::CROSSWALK &&
```
xiaoxq: ditto.

modules/planning/tasks/optimizers/path_decider/path_decider.cc
panjiacheng: Tomorrow we will do road test. This is for the purpose of easily switching between different configurations. I will clean things up when road test completes.

modules/planning/tasks/deciders/path_lane_borrow_decider/path_lane_borrow_decider.cc
panjiacheng: ditto

modules/planning/tasks/deciders/path_bounds_decider.cc
panjiacheng: fixed this.

modules/planning/tasks/optimizers/path_decider/path_decider.cc
xiaoxq: It's not a good practice though. A solution might be to switch with FLAGS, so that you don't need to rebuild.
And at least a TODO to clarify that it will be tested and cleaned up later.

PR #8062, opened 2019-04-26
ignore global flagfile

Fla3inH0tCheet0s:
A lot of dreamview interactions cause changes in `modules/common/data/global_flagfile.txt`. It's hard to keep track of manual software changes and GUI default changes. This PR is meant to reduce those.

This PR was tested with
```
docker/scripts/dev_start.sh
docker/scripts/dev_into.sh
./apollo.sh build_opt
rm -r data/log
bootstrap.sh
ls data/log
```
Everything builds and the log files are in the right location.

Note that running `docker/scripts/dev_start.sh` is necessary when running with this change.


Fla3inH0tCheet0s:
It's confusing to mix parameters with UI state. gflags already has default values. There is not need to duplicate defaults in this file. Specific parameters can be put in other files.

xiaoxq:
It makes a lot of sense. Thank you for the great suggestion!
As of now we still rely on it for some cross-module configuration and testing. Please leave it for a while and we'll try to resolve. Thanks!

.gitignore
```diff
@@ -84,3 +84,7 @@ docs/demo_guide/*.bag
 
 # Logs
 cyber/log/*
+
+# Local HMI configuration
+modules/common/data/global_flagfile.txt
```
xiaoxq: It's not only for HMI. It's for global cross-module configuration.

cyber/logger/logger_util.cc
```diff
@@ -53,7 +53,7 @@ const std::vector<std::string>& GetLoggingDirectories() {
     if (!FLAGS_log_dir.empty()) {
       logging_directories_list.emplace_back(FLAGS_log_dir.c_str());
     } else {
-      logging_directories_list.emplace_back("./");
+      logging_directories_list.emplace_back("/apollo/log");
```
xiaoxq: /apollo/data/log

Fla3inH0tCheet0s:
I thought `modules/common/data/global_flagfile.txt` could be replaced, by the default values in gflag's `DEFINE` macro.

If you want to keep it for cross-module configuration, you can do `modules/module_name/conf/module_name.conf` -> `modules/common/data/global_flagfile.txt` -> `modules/common/data/hmi.txt`. Each arrow represents setting the `--flagfile` gflag similar to how each module conf file does it currently. Then, `modules/common/data/global_flagfile.txt` could be tracked by git and `modules/common/data/hmi.txt` could be ignored.

xiaoxq:
After some investigation I think that we have to keep it for a while due to some daily workflow, cloud jobs as well as release mechanism. It will be on my schedule to do deep clean on this.
Thank you for the heads up :)

PR #8065, opened 2019-04-26
Planning: consider lon_speed for piecewise_path

HongyiSun:


panjiacheng:
I guess you meant “lateral” speed rather than “longitudinal” speed.

YajiaZhang:
> I guess you meant “lateral” speed rather than “longitudinal” speed.

It is indeed lon speed.

PR #8079, opened 2019-04-26
planning: open space enable constraint checking for final ipopt output

deidaraho:


modules/planning/open_space/trajectory_smoother/distance_approach_ipopt_interface.cc
```diff
@@ -1379,6 +1379,11 @@ void DistanceApproachIPOPTInterface::finalize_solution(
   free(cind_L);
   free(hessval);
 
+  // debug only
```
Capri2014: why not check constant before allocating state/time results?

deidaraho: > why not check constant before allocating state/time results?

it is fixed.

PR #8080, opened 2019-04-26
planning: retire task DECIDER_RULE_BASED_STOP and rename it to OPEN_S…

jmtao:
…PACE_PRE_STOP_DECIDER for open-space-parking only

### DIFF
--- a/modules/planning/tasks/deciders/decider.cc
+++ b/modules/planning/tasks/deciders/decider.cc
@@ -20,8 +20,12 @@
 
 #include "modules/planning/tasks/deciders/decider.h"
 
+#include "modules/common/util/util.h"
+
 namespace apollo {
 namespace planning {
+
+using common::util::WithinBound;
 
 Decider::Decider(const TaskConfig& config) : Task(config) {}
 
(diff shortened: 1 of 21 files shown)


modules/planning/tasks/deciders/decider.cc
```diff
@@ -36,5 +40,58 @@ apollo::common::Status Decider::Execute(Frame* frame) {
...
+    const std::vector<std::string>& wait_for_obstacles, Frame* const frame,
+    ReferenceLineInfo* const reference_line_info) {
+  CHECK_NOTNULL(frame);
+  CHECK_NOTNULL(reference_line_info);
+
+  // check
+  const auto& reference_line = reference_line_info->reference_line();
+  if (!WithinBound(0.0, reference_line.Length(), stop_line_s)) {
+    AERROR << "stop_line_s[" << stop_line_s << "] is not on reference line";
+    return 0;
```
YajiaZhang: Better to return true or false to avoid confusion. I blame the original author of the code :-)

jmtao: that would be me.  will fix it. thx

JasonZhou404:
Thx for the help!

PR #8082, opened 2019-04-26
Planning: revert breaking open space debug, and fix for missing open space decider conf

Capri2014:
Since we have decided to change designs to use Astar when smoother failed, there is no point of making the recording part more complicated than needed.

### DIFF
--- a/modules/planning/on_lane_planning.cc
+++ b/modules/planning/on_lane_planning.cc
@@ -666,9 +666,9 @@ void OnLanePlanning::AddOpenSpaceOptimizerResult(
   smoothed_line->set_label("Smooth");
   size_t adc_label = 0;
   for (const auto& point : smoothed_trajectory.vehicle_motion_point()) {
-    const auto& x = point.trajectory_point().path_point().x();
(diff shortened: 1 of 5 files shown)


modules/planning/on_lane_planning.cc
```diff
@@ -613,110 +613,112 @@ void OnLanePlanning::ExportOpenSpaceChart(
...
+  }
+
+  auto smoothed_trajectory = open_space_debug.smoothed_trajectory();
+  auto* smoothed_line = chart->add_line();
+  smoothed_line->set_label("Smooth");
+  size_t adc_label = 0;
+  for (const auto& point : smoothed_trajectory.vehicle_motion_point()) {
+    const auto& x = point.trajectory_point().path_point().x();
+    const auto& y = point.trajectory_point().path_point().y();
+    const auto& heading = point.trajectory_point().path_point().theta();
```
xiaoxq: For scalar types just store the value, which gives you best locality.

Capri2014: @xiaoxq Done

PR #8085, opened 2019-04-27
docker:add scripts to intall ad-rss-lib and plat-sw package

gengqx:
hope to push those two lib into docker.

### DIFF
--- a/docker/build/installers/install_intel_ad_rss_lib.sh
+++ b/docker/build/installers/install_intel_ad_rss_lib.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+
+###############################################################################
+# Copyright 2019 The Apollo Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


docker/build/installers/install_intel_ad_rss_lib.sh
```diff
@@ -0,0 +1,30 @@
...
+set -e
+
+REPO="https://github.com/intel/ad-rss-lib"
+TAG="v1.1.0"
+DIR="/home/tmp/ad-rss-lib"
+
+git clone "${REPO}" "${DIR}"
+pushd "${DIR}"
+git checkout "${TAG}"
+rm -fr .git
```
xiaoxq: How about downloading the release tar ball directly and then unzip?
https://github.com/intel/ad-rss-lib/archive/v1.1.0.zip

gengqx: It would be ok,i just reference the workspace.in file,see the rss is a git repo. I would close this request,I just hope those two lib could put into docker.

PR #8095, opened 2019-04-29
tools: introducing smart_recorder for recording selective messages

michael4338:


modules/data/tools/smart_recorder/trigger_base.h
```diff
@@ -0,0 +1,74 @@
...
+  virtual void Pull(const RecordMessage& msg) = 0;
+
+  // Decide if the current message needs to be restored
+  virtual bool ShouldRestore(const RecordMessage& msg) const = 0;
+
+  // Getters
+  const std::unordered_map<std::string, std::string>& GetChannelTypes() const {
+    return channels_types_map_;
+  }
+  const std::string GetTriggerName() const { return trigger_name_; }
```
xiaoxq: const &.

modules/data/tools/smart_recorder/small_topics_trigger.h
```diff
@@ -0,0 +1,43 @@
...
+ * @brief
+ * A specialized trigger that does not trigger anything but indicates
+ * what small topics need to be restored
+ */
+class SmallTopicsTrigger : public TriggerBase {
+ public:
+  SmallTopicsTrigger();
+
+  bool Init(const SmartRecordTrigger& trigger_conf) override;
+  void Pull(const RecordMessage& msg) override{};
```
xiaoxq: Space after override.

modules/data/tools/smart_recorder/record_processor.h
```diff
@@ -0,0 +1,67 @@
...
+                  std::string restored_output_dir);
+  bool Init(const SmartRecordTrigger& trigger_conf);
+  bool Process();
+  ~RecordProcessor() { writer_->Close(); }
+
+ private:
+  void LoadSourceRecords();
+  void CollectAllChannels(const SmartRecordTrigger& trigger_conf);
+  void RestoreMessage(const RecordMessage& msg);
+  const std::string GetDefaultOutputFile() const;
```
xiaoxq: const &.

modules/data/tools/smart_recorder/record_processor.cc
```diff
@@ -0,0 +1,184 @@
...
+  for (const auto& iter : channels_types_map_) {
+    all_channels_.insert(iter.first);
+  }
+}
+
+void RecordProcessor::RestoreMessage(const RecordMessage& msg) {
+  auto* raw_message = ProtobufFactory::Instance()->GenerateMessageByType(
+      channels_types_map_[msg.channel_name]);
+  CHECK(raw_message) << "Failed to do reflection from " << msg.channel_name;
+  std::string get_desc_str("");
```
xiaoxq: Same with `std::string get_desc_str;`.

modules/data/tools/smart_recorder/trigger_base.cc
```diff
@@ -0,0 +1,63 @@
...
+    return false;
+  }
+  for (const auto& iter : trigger_obj_->wanted_channel_types()) {
+    InsertChannelType(iter.first, iter.second);
+  }
+  return true;
+}
+
+void TriggerBase::LockTrigger(const SmartRecordTrigger& trigger_conf) {
+  for (const auto& iter : trigger_conf.triggers()) {
```
xiaoxq: Rename `iter` as `trigger` to better describe its type.

modules/data/tools/smart_recorder/record_processor.h
michael4338: it said returning a reference to local variable:)

xiaoxq: It's not a reference, it's a copy.
1. If user want a local copy and modify later: `std::string file = GetDefaultOutputFile()`
2. If user just want a read only reference: `const std::string& file = GetDefaultOutputFile()`

So it's still better to return a const reference which saves memory for case 2.

And there is no reason to return a CONST copy. As for case 1 it will create two objects, the first one is const and returns by the function, the second one is non-const which is copy-constructed from the const one.

michael4338: got it, the warning with & was "returning reference to temporary", sorry for misleading:), so removed both const and reference like #1 suggested.

PR #8099, opened 2019-04-30
 Planning: Costheta reference line smoother improvements

JasonZhou404:
Not enabled.

### DIFF
--- a/modules/planning/reference_line/cos_theta_problem_interface.cc
+++ b/modules/planning/reference_line/cos_theta_problem_interface.cc
@@ -72,35 +72,13 @@ bool CosThetaProbleminterface::get_bounds_info(int n, double* x_l, double* x_u,
   // a. for x, y
   for (size_t i = 0; i < num_of_points_; ++i) {
     size_t index = i << 1;
-    double x_lower = 0.0;
-    double x_upper = 0.0;
-    double y_lower = 0.0;
-    double y_upper = 0.0;
-    double bound = std::min(lateral_bounds_[i], default_max_point_deviation_);
-
-    if (i == 0 && has_fixed_start_point_) {
-      x_lower = start_x_ - relax_;
-      x_upper = start_x_ + relax_;
-      y_lower = start_y_ - relax_;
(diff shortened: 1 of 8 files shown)


modules/planning/reference_line/cos_theta_problem_interface.cc
```diff
@@ -555,5 +619,196 @@ void CosThetaProbleminterface::hessian_strcuture() {
...
+    g[index] = x[index];
+    g[index + 1] = x[index + 1];
+  }
+  return true;
+}
+
+/** Method to generate the required tapes */
+void CosThetaProbleminterface::generate_tapes(int n, int m, int* nnz_jac_g,
+                                              int* nnz_h_lag) {
+  Ipopt::Number* xp = new double[n];
```
xiaoxq: Avoid old-school memory management.
Use `vector<double> xp;`, and get the raw pointer as `&xp[0]` if the upstream API requires.

JasonZhou404: Part of them fixed. Some of them are not fixed successfully yet, will look into it afterwards

modules/planning/reference_line/cos_theta_reference_line_smoother.h
```diff
@@ -43,25 +44,31 @@ class CosThetaReferenceLineSmoother : public ReferenceLineSmoother {
   void SetAnchorPoints(const std::vector<AnchorPoint>&) override;
 
  private:
-  bool Smooth(const std::vector<Eigen::Vector2d>& point2d,
-              const std::vector<double>& lateral_bounds,
-              std::vector<common::PathPoint>* ptr_smoothed_point2d);
+  bool CosThetaSmooth(const std::vector<Eigen::Vector2d>& point2d,
```
YajiaZhang: Better rename it to smooth; the function is defined in CosThetaReferenceLineSmoother, no need to specifically state CosThetaSmooth.

xiaoxq:
CI seems stuck. I'll force-merge it per committer request.

PR #8107, opened 2019-04-30
Planning: deprecate old open space planning files. 

Capri2014:


Capri2014:
Pass rate in simulation kept the same.

PR #8113, opened 2019-05-01
Perception: Refined CIPV detection

techoe:


modules/perception/camera/app/cipv_camera.cc
```diff
@@ -83,29 +83,37 @@ bool Cipv::DistanceFromPointToLineSegment(const Point2Df &point,
...
       if (curve_image_point_size < min_laneline_length_for_cipv_) {
         *b_left_valid = false;
       } else {
         *b_left_valid = true;
 
+        for (size_t j = 0; j < curve_car_coord_point_set_size; ++j) {
+          // ground_point
+          x = lane_object.curve_car_coord_point_set[j].x;
+          y = lane_object.curve_car_coord_point_set[j].y;
+          egolane_ground->left_line.line_point.emplace_back(Point2Df(x, y));
```
xiaoxq: `emplace_back(x, y)` would work.

modules/perception/camera/app/cipv_camera.cc
```diff
@@ -142,10 +152,10 @@ bool Cipv::MakeVirtualLane(const LaneLineSimple &ref_lane_line,
   virtual_lane_line->line_point.clear();
   if (b_image_based_cipv_ == false) {
     for (uint32_t i = 0; i < ref_lane_line.line_point.size(); ++i) {
-      Eigen::Vector2f virtual_line_point(
+      Point2Df virtual_line_point(
           ref_lane_line.line_point[i](0),
           ref_lane_line.line_point[i](1) + offset_distance);
-      virtual_lane_line->line_point.push_back(virtual_line_point);
+      virtual_lane_line->line_point.emplace_back(virtual_line_point);
```
xiaoxq: 1. Use `push_back` when the object is already constructed.
2. Or use `emplace_back(x, y)` to construct in place which is more efficient.

modules/perception/camera/app/cipv_camera.cc
```diff
@@ -190,12 +202,12 @@ bool Cipv::MakeVirtualEgoLaneFromYawRate(const float yaw_rate,
   right_lane_line->line_point.clear();
 
   if (b_image_based_cipv_ == false) {
-    for (uint32_t i = 0; i < 120; i += 5) {
+    for (uint32_t i = 1; i < 25; i += 1) {
```
xiaoxq: Could it be a const global variable to better describe the magic number `25`?

modules/perception/camera/lib/lane/postprocessor/darkSCNN/darkSCNN_lane_postprocessor.cc
```diff
@@ -201,7 +201,7 @@ bool DarkSCNNLanePostprocessor::Process2D(
               std::fabs(xy_point(1) - xy_points[value].back()(1)) < 1.0f) {
             xy_points[value].push_back(xy_point);
             uv_points[value].push_back(uv_point);
-          }
+        }
```
xiaoxq: Need indents.

modules/perception/camera/app/cipv_camera.cc
techoe: Thank you for the review! I am replacing them now.

modules/perception/camera/app/cipv_camera.cc
techoe: I prefer emplace_back to push_back. Eventually I plan to remove all dynamic data allocations and replace them with fixed_vector or circular_vector

modules/perception/camera/app/cipv_camera.cc
techoe: 25 is changed to constant, kMaxNumVirtualLanePoint, which doesn't have to be configurable.

modules/perception/camera/lib/lane/postprocessor/darkSCNN/darkSCNN_lane_postprocessor.cc
techoe: Fixed.

PR #8133, opened 2019-05-01
planning: remove old pull-over implementation

jmtao:


Capri2014:
Add @panjiacheng as reviewer as well since he would be refactoring the old pull-over logic into PathBoundDecider per discussion.

PR #8142, opened 2019-05-02
planning: fix a bug in destination. when stop point is not on lane, s…

jmtao:
…till need create the virtual obstacle, but skip the STOP decision

### DIFF
--- a/modules/planning/common/util/common.cc
+++ b/modules/planning/common/util/common.cc
@@ -108,6 +108,12 @@ int BuildStopDecision(const std::string& stop_wall_id,
     return -1;
   }
 
+  const auto& stop_wall_box = stop_wall->PerceptionBoundingBox();
+  if (!reference_line.IsOnLane(stop_wall_box.center())) {
+    ADEBUG << "stop point is not on lane. SKIP STOP decision";
+    return 0;
+  }
+
   // build stop decision
   auto stop_point = reference_line.GetReferencePoint(
       stop_wall->PerceptionSLBoundary().start_s() - stop_distance);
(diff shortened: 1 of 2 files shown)


modules/planning/common/util/common.cc
```diff
@@ -108,6 +108,12 @@ int BuildStopDecision(const std::string& stop_wall_id,
     return -1;
   }
 
+  const auto stop_wall_box = stop_wall->PerceptionBoundingBox();
```
YajiaZhang: const auto& ?

modules/planning/traffic_rules/destination.cc
```diff
@@ -78,16 +54,32 @@ int Destination::MakeDecisions(
...
+  const auto& adc_sl = reference_line_info->AdcSlBoundary();
+  const auto& dest =
+      PlanningContext::Instance()->mutable_planning_status()->destination();
+  if (adc_sl.start_s() > dest_sl.s() && !dest.has_passed_destination()) {
+    ADEBUG << "Destination at back, but we have not reached destination yet";
+    return 0;
+  }
+
+  double dest_lane_s = std::max(
+      0.0, routing_end.s() - FLAGS_virtual_stop_wall_length -
```
YajiaZhang: Better to use fmax for floating numbers.

modules/planning/common/util/common.cc
jmtao: fixed. amended. thx~

modules/planning/traffic_rules/destination.cc
jmtao: fixed. amended. thx~

PR #8147, opened 2019-05-03
Planning: fix nan in kappa estimation

JasonZhou404:


modules/planning/reference_line/cos_theta_reference_line_smoother.cc
```diff
@@ -336,8 +336,9 @@ bool CosThetaReferenceLineSmoother::GenerateRefPointProfile(
     double yds = y_over_s_first_derivatives[i];
     double xdds = x_over_s_second_derivatives[i];
     double ydds = y_over_s_second_derivatives[i];
-    double kappa = (xds * ydds - yds * xdds) /
-                   (std::sqrt(xds * xds + yds * yds) * (xds * xds + yds * yds));
+    double kappa =
+        (xds * ydds - yds * xdds) /
+        (std::sqrt(xds * xds + yds * yds) * (xds * xds + yds * yds) + 1e-6);
```
YajiaZhang: In this case, the reference line will be a single point or multiple points with same position. I think there might be something wrong in the implementation.

xiaoxq:
Force merge due to travis issue.

PR #8159, opened 2019-05-04
[WIP]Planning: use last good path data to do speed fallback

JasonZhou404:
Dreamland public testing


HongyiSun:
fixed in #8189 , which passed all public simulation cases

PR #8168, opened 2019-05-06
Planning: pull-over and reference-line offset bug fix.

panjiacheng:


modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -274,6 +274,100 @@ std::string PathBoundsDecider::GenerateFallbackPathBound(
...
 
+bool PathBoundsDecider::SearchPullOverPosition(
+    const ReferenceLineInfo& reference_line_info,
+    const std::vector<std::tuple<double, double, double>>& path_bound,
+    std::tuple<double, double, double>* const pull_over_position) {
+  double adc_end_s = reference_line_info.AdcSlBoundary().end_s();
+  double destination_s =
+      reference_line_info.SDistanceToDestination() + adc_end_s;
+  // Check if destination is some distance away from ADC.
+  double destination_to_adc_buffer = 10.0;
```
Capri2014: make this to config

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -274,6 +274,100 @@ std::string PathBoundsDecider::GenerateFallbackPathBound(
...
+  if (destination_s - adc_end_s < destination_to_adc_buffer) {
+    AWARN << "Destination is too close to ADC.";
+    return false;
+  }
+  // Check if destination is within path-bounds searching scope.
+  double destination_to_pathend_buffer = 10.0;
```
Capri2014: same for this one

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -274,6 +274,100 @@ std::string PathBoundsDecider::GenerateFallbackPathBound(
...
+  if (destination_s + destination_to_pathend_buffer >=
+      std::get<0>(path_bound.back())) {
+    AWARN << "Destination is not within path-bounds searching scope.";
+    return false;
+  }
+
+  // Search for a feasible location for pull-over.
+  double road_edge_buffer = 0.15;
```
Capri2014: diddo

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -274,6 +274,100 @@ std::string PathBoundsDecider::GenerateFallbackPathBound(
...
+  double pull_over_space_length =
```
Capri2014: use const whenever possible.

modules/planning/common/frame.h
```diff
@@ -169,6 +170,13 @@ class Frame {
 
   perception::TrafficLight GetSignal(const std::string &traffic_light_id) const;
 
+  std::tuple<bool, double, double, double> pull_over_info() {
+    return pull_over_info_;
+  }
+
+  void set_pull_over_info(const std::tuple<bool, double, double, double>&
```
Capri2014: Where is this "set_pull_over_info" called? Also is the tuple defined as "is_feasible_widnow, pull_over_x, pull_over_y, pull_over_phi"? If so, could you add comments here so to avoid confusion?

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -274,6 +274,100 @@ std::string PathBoundsDecider::GenerateFallbackPathBound(
   return "";
 }
 
+bool PathBoundsDecider::SearchPullOverPosition(
+    const ReferenceLineInfo& reference_line_info,
+    const std::vector<std::tuple<double, double, double>>& path_bound,
+    std::tuple<double, double, double>* const pull_over_position) {
```
YajiaZhang: Better to name it pull_over_configuration.

modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -274,6 +274,100 @@ std::string PathBoundsDecider::GenerateFallbackPathBound(
...
+      VehicleConfigHelper::GetConfig().vehicle_param().length() * 2.0;
+  double adc_half_width =
+      VehicleConfigHelper::GetConfig().vehicle_param().width() / 2.0;
+  int i = static_cast<int>(path_bound.size()) - 1;
+  // 1. Locate the first point before destination.
+  while (i >= 0 && std::get<0>(path_bound[i]) > destination_s) { --i; }
+  // 2. Find a window that is close to road-edge.
+  bool has_a_feasible_window = false;
+  while (i >= 0 && std::get<0>(path_bound[i]) -
+         std::get<0>(path_bound.front()) > pull_over_space_length) {
```
YajiaZhang: Infinite loop? I didn't see where i changes at each iteration.

modules/planning/common/frame.h
panjiacheng: Thanks. This problem is solved. Please see the most recent commit.

modules/planning/common/frame.h
```diff
@@ -169,6 +170,13 @@ class Frame {
 
   perception::TrafficLight GetSignal(const std::string &traffic_light_id) const;
 
+  // tuple<exist_pull_over_position, x, y, theta>
+  std::tuple<bool, double, double, double> pull_over_info() {
```
xiaoxq: Return a const reference. You save a copy if user only wants an immutable reference.

panjiacheng: fixed this.

PR #8170, opened 2019-05-06
Planning: use ref_x in qp_offset

HongyiSun:


modules/planning/math/piecewise_jerk/piecewise_jerk_problem.h
```diff
@@ -80,6 +80,10 @@ class PiecewiseJerkProblem {
     x_end_ = x_end;
   }
 
+  void SetRefX(std::vector<double> x_ref);
```
YajiaZhang: Better rename to SetZeroOrderReference

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.h
```diff
@@ -80,6 +80,10 @@ class PiecewiseJerkProblem {
...
+
+  void SetDirivativePenalty(std::vector<double> penalty_dx);
```
YajiaZhang: Same here, SetFirstOrderPenalty

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
```diff
@@ -104,6 +104,19 @@ bool PiecewiseJerkProblem::OptimizeWithOsqp(
   return true;
 }
 
+void PiecewiseJerkProblem::SetZeroOrderReference(std::vector<double> x_ref) {
+  if (x_ref.size() == num_of_knots_) {
+    x_ref_ = std::move(x_ref);
```
xiaoxq: A copy is always created. Just pass `const std::vector<double>&` and assign it to `x_ref_`. You saved a copy when `x_ref.size() != num_of_knots_`.

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
```diff
@@ -416,17 +429,12 @@ void PiecewiseJerkProblem::CalculateOffset(std::vector<c_float>* q) {
...
   q->resize(kNumParam);
-  for (int i = 0; i < kNumParam; ++i) {
-    if (i < N) {
-      q->at(i) = -1.0 * weight_.x_ref_w *
-                 (std::get<0>(x_bounds_[i]) + std::get<1>(x_bounds_[i]));
-    } else {
-      q->at(i) = 0.0;
-    }
+  for (int i = 0; i < N; ++i) {
+    q->at(i) += -2.0 * weight_.x_ref_w * x_ref_[i];
```
xiaoxq: What are these `-2.0`s? Name it or comment it.

HongyiSun: It's the offset term from x^2 - 2*x*x_ref - x_ref^2 in QP problem

Capri2014:
You may want to update the interfaces in https://github.com/ApolloAuto/apollo/blob/78086b336525394b318da370c32e1aac0117d57c/modules/planning/open_space/coarse_trajectory_generator/hybrid_a_star.cc#L397 as well to avoid build break.

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
YajiaZhang: Copy is not necessary if calling the function this way: SetZeroOrderReference(std::move(some_vector)). The caller can decide if some_vector will not be used any longer. In this way, some_vector will be moved to x_ref_ without copying.

HongyiSun:
> You may want to update the interfaces in
> 
> https://github.com/ApolloAuto/apollo/blob/78086b336525394b318da370c32e1aac0117d57c/modules/planning/open_space/coarse_trajectory_generator/hybrid_a_star.cc#L397
> 
> as well to avoid build break.

Solved

PR #8191, opened 2019-05-07
DataCollection: added reverse config

vlin17:


Capri2014:
LGTM @sjiang2018 @JasonXu12 Please double check if the criteria set for reverse gear makes sense.

PR #8192, opened 2019-05-07
Planning: save pull_over_info output from PathBoundDecider into plann…

Capri2014:
…ing_context instead of into frame

### DIFF
--- a/modules/planning/tasks/deciders/path_bounds_decider.cc
+++ b/modules/planning/tasks/deciders/path_bounds_decider.cc
@@ -31,6 +31,7 @@
 #include "modules/common/util/util.h"
 #include "modules/map/hdmap/hdmap_util.h"
 #include "modules/planning/common/path_boundary.h"
+#include "modules/planning/common/planning_context.h"
 #include "modules/planning/common/planning_gflags.h"
 #include "modules/planning/tasks/deciders/path_decider_obstacle_utils.h"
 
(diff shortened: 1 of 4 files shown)


modules/planning/tasks/deciders/path_bounds_decider.cc
```diff
@@ -364,8 +380,8 @@ bool PathBoundsDecider::SearchPullOverPosition(
       has_a_feasible_window = true;
       const auto& reference_line = reference_line_info.reference_line();
 
-      double pull_over_s = std::get<0>(path_bound[(i+j)/2]);
-      double pull_over_l = std::get<1>(path_bound[(i+j)/2]);
+      const double pull_over_s = std::get<0>(path_bound[(i + j) / 2]);
+      const double pull_over_l = std::get<1>(path_bound[(i + j) / 2]);
```
xiaoxq: Save `path_bound[(i + j) / 2]` to a variable and reuse.

modules/planning/tasks/deciders/path_bounds_decider.h
```diff
@@ -114,7 +114,8 @@ class PathBoundsDecider : public Decider {
   bool SearchPullOverPosition(
       const ReferenceLineInfo& reference_line_info,
       const std::vector<std::tuple<double, double, double>>& path_bound,
-      std::tuple<double, double, double>* const pull_over_configuration);
+      std::tuple<double, double, double, double, double>* const
```
xiaoxq: Long tuples are hard to read and maintain. Use struct or proto instead.
But it's OK leave as it is for this PR.

Capri2014: Yes, that is why I left a TODO for me and jiacheng for a re-factor

modules/planning/tasks/deciders/path_bounds_decider.cc
Capri2014: Done

PR #8194, opened 2019-05-07
drivers:update sensor_canbus template to use cyber framework

gengqx:


modules/drivers/canbus/sensor_canbus.h
```diff
@@ -160,10 +159,10 @@ bool SensorCanbus<SensorType>::Start() {
...
   // no need for timer.
   if (FLAGS_sensor_freq > 0) {
-    const double duration_ms = 1000.0 / FLAGS_sensor_freq;
-    timer_ = cyber::Timer(duration_ms, &SensorCanbus<SensorType>::OnTimer, this,
-                          false);
-    timer_.Start();
+    double duration_ms = 1000.0 / FLAGS_sensor_freq;
+    timer_.reset(new cyber::Timer(static_cast<uint32_t>(duration_ms),
+                                  [this]() { this->OnTimer(); }, false));
+    timer_->Start();
```
fengqikai1414: you can use TimerComponent directly. [Here](https://github.com/ApolloAuto/apollo/tree/master/cyber/examples/timer_component_example) is the example.

gengqx: I think use timer is the best practice, this sensor_canbus is a template classs to support two kind of trigger method, one is timer,the other is data_trigger, I don't know if change to timecomponent, could it support data_trigger.

fengqikai1414: Yes, you are right, using timer here is better.

PR #8198, opened 2019-05-07
Perception: Changed minimum lidar filter range 0.9->1.6 meter to remo…

techoe:
…ve ghost object behind. 
We should watch if this filter causes less detection nearby.


techoe:
OK. I will update the apollo-internal instead.

techoe:
And will remove this PR.

PR #8248, opened 2019-05-09
Add a markdown file which is coolhigh's user manual

lishuanghua:
Add a markdown file which is coolhigh's user manual


docs/specs/apollo_3.0_development_kits_user_manual_cn.md
```diff
@@ -0,0 +1,1123 @@
+教学套件使用手册
+
+# 目录
+      
+ - 概览
+ - 工控机系统安装
+	 - 工控机硬件安装
```
xiaoxq: Use space instead of tab.

PR #8276, opened 2019-05-10
Dreamview: support PiecewiseJerkSpeedOptimizer monitoring

vlin17:
Also removed dpPolyPath graph


modules/dreamview/frontend/src/store/config/PlanningGraph.yml
```diff
@@ -218,13 +218,7 @@ kappaGraph:
...
                 fill: false
                 showLine: true
-            DpPolyPathOptimizer:
-                color: 'rgba(27, 249, 105, 0.5)' # green
-                borderWidth: 2
-                pointRadius: 0
-                fill: false
-                showLine: true
-            QpSplinePathOptimizer:
+            PiecewiseJerkSpeedOptimizer:
```
HongyiSun: The QpSplinePathOptimizer also need to be removed instead change into PiecewiseJerkSpeedOptimizer

vlin17: removed!

PR #8296, opened 2019-05-14
Planning: disable jacobian ad for testing

JasonZhou404:


modules/planning/proto/planner_open_space_config.proto
```diff
@@ -98,6 +98,7 @@ message DistanceApproachConfig {
   // True to enable derivative check inside open space planner
   optional bool enable_initial_final_check = 25 [default = false];
   optional DistanceApproachMode distance_approach_mode = 26;
+  optional bool enable_jacobian_ad = 27;
```
Capri2014: please give a default value of false here.

JasonZhou404: done

PR #8307, opened 2019-05-14
Planning: remove buffer in st_boundary_mapper

JasonZhou404:


modules/planning/conf/planning_config.pb.txt
```diff
@@ -75,7 +75,7 @@ default_task_config: {
     high_speed_threshold: 12.58
     low_speed_threshold: 7.5
     minimal_kappa: 0.00001
-    point_extension: 1.0
+    point_extension: 0.0
```
YajiaZhang: When the changes are finalized, remove the parameters in the protobuf.

JasonZhou404: sure

PR #8321, opened 2019-05-14
Perception: Fixed key handler for Num Lock; Added manual calibration …

techoe:
…mode

### DIFF
--- a/modules/perception/camera/tools/offline/keycode.h
+++ b/modules/perception/camera/tools/offline/keycode.h
@@ -1,0 +1,116 @@
+/******************************************************************************
+ * Copyright 2019 The Apollo Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the License);
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+#pragma once
+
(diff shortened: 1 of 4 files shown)


modules/perception/camera/tools/offline/visualizer.cc
```diff
@@ -662,48 +663,59 @@ bool Visualizer::key_handler(const std::string &camera_name, const int key) {
     case 86: case 118:  // 'V' 'v'
       show_velocity_ = (show_velocity_ + 1) % 2;
       break;
-    case 65362:  // Up_Arrow
-      if (pitch_adj_degree_ + 0.05 <= max_pitch_degree_) {
+    case 1113938: case 65362:  // Up_Arrow with Num Lock (on/off)
```
xiaoxq: Better to name these key code as const variables ahead and use them, like
```
enum KeyCode {
  KEY_UPPER_V = 86,
  KEY_LOWER_V = 118,
  KEY_UP_ARROW = 1113938,
  ...
}
```
I guess there are some third-party libraries providing these values, which would be even better.

techoe: Thank you for the review. I modified the code to have enum. I checked 3rd party libraries but I think it is better to keep our own to avoid any dependency or copyright issues.

modules/perception/camera/tools/offline/keycode.h
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an AS IS BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+#pragma once
+
+namespace apollo {
```
xiaoxq: Please also add the `namespace perception`. If it's used by more modules in the future, we can move it to modules/common.

techoe: Agree. I added "namespace perception".

PR #8323, opened 2019-05-15
Prediction: update flag in frame env script accordingly

kechxu:


modules/prediction/common/message_process.cc
```diff
@@ -168,8 +168,9 @@ void MessageProcess::OnPerception(
 
   // Make evaluations
   EvaluatorManager::Instance()->Run();
-
-  if (FLAGS_prediction_offline_mode == 2) {
+  // TODO(kechxu) reorder the offline mode indices
```
xiaoxq: Also add a TODO item to remove the magic numbers.

PR #8324, opened 2019-05-15
Prediction: update the lane-scanning feature extraction.

panjiacheng:


modules/prediction/common/feature_output.cc
```diff
@@ -88,13 +88,26 @@ void FeatureOutput::InsertFeatureProto(const Feature& feature) {
 void FeatureOutput::InsertDataForLearning(
     const Feature& feature, const std::vector<double>& feature_values,
     const std::string& category, const LaneSequence* lane_sequence_ptr) {
+  std::vector<std::string> dummy_string_feature_values;
```
xiaoxq: const, or simply pass `{}` to the function.

modules/prediction/common/feature_output.cc
```diff
@@ -88,13 +88,26 @@ void FeatureOutput::InsertFeatureProto(const Feature& feature) {
...
   DataForLearning* data_for_learning =
       list_data_for_learning_.add_data_for_learning();
   data_for_learning->set_id(feature.id());
   data_for_learning->set_timestamp(feature.timestamp());
   for (size_t i = 0; i < feature_values.size(); ++i) {
     data_for_learning->add_features_for_learning(feature_values[i]);
   }
+  for (size_t i = 0; i < string_feature_values.size(); ++i) {
+    data_for_learning->add_string_features_for_learning(
+        string_feature_values[i]);
```
xiaoxq: Clean and efficient:
```
*data_for_learning->mutable_string_features_for_learning() = {
    string_feature_values.begin(), string_feature_values.end()};
```

modules/prediction/container/obstacles/obstacle.cc
```diff
@@ -1202,7 +1202,7 @@ void Obstacle::BuildLaneGraphFromLeftToRight() {
   }
 
   // Build lane_points.
-  if (feature->has_lane() && feature->lane().has_lane_graph()) {
+  if (feature->has_lane() && feature->lane().has_lane_graph_ordered()) {
```
xiaoxq: You don't need to check `has_lane` because the second condition will cover it.

modules/prediction/evaluator/vehicle/lane_scanning_evaluator.cc
```diff
@@ -101,6 +105,20 @@ void LaneScanningEvaluator::Evaluate(Obstacle* obstacle_ptr,
...
 }
 
+bool LaneScanningEvaluator::ExtractStringFeatures(
+    const LaneGraph& lane_graph,
+    std::vector<std::string>* const string_feature_values) {
+  for (int i = 0; i < lane_graph.lane_sequence_size(); ++i) {
+    const LaneSequence& lane_sequence = lane_graph.lane_sequence(i);
+    string_feature_values->push_back("|");
+    for (int j = 0; j < lane_sequence.lane_segment_size(); ++j) {
+      const LaneSegment& lane_segment = lane_sequence.lane_segment(j);
```
xiaoxq: Learn the `for-each` style:
`for (const LaneSegment& lane_segment : lane_sequence.lane_segment())`

Use it whenever possible. It is simpler and could avoid off-one bug

modules/prediction/proto/offline_features.proto
```diff
@@ -21,6 +21,7 @@ message DataForLearning {
   optional double timestamp = 2;
   // The features for learning algorithms:
   repeated double features_for_learning = 3;
+  repeated string string_features_for_learning = 7;
```
xiaoxq: IMO the `string_` prefix is not needed.

panjiacheng: This is to differentiate from the existing "features_for_learning".

PR #8348, opened 2019-05-16
Common/Data/Canbus: use VehicleID to differentiate different vehicles. 

Capri2014:


modules/common/configs/proto/vehicle_config.proto
```diff
@@ -36,45 +36,52 @@ enum VehicleBrand {
...
+}
+
 message VehicleParam {
   optional VehicleBrand brand = 1;
   // Car center point is car reference point, i.e., center of rear axle.
-  optional double front_edge_to_center = 2 [default = nan];
-  optional double back_edge_to_center = 3 [default = nan];
-  optional double left_edge_to_center = 4 [default = nan];
-  optional double right_edge_to_center = 5 [default = nan];
+  optional VehicleID id = 2;
```
xiaoxq: Better not to change field ID. Just use the next available one, i.e., 21.

Capri2014: @xiaoxq This config only serves as configuration, not publish/recorded information. I thought we can freely change protos like these?

PR #8357, opened 2019-05-16
Planning: enable fem pos reference line

JasonZhou404:


modules/planning/reference_line/discrete_points_reference_line_smoother.cc
```diff
@@ -259,7 +258,7 @@ void DiscretePointsReferenceLineSmoother::DeNormalizePoints(
                   auto curr_y = p.second;
                   std::pair<double, double> xy(curr_x + zero_x_,
                                                curr_y + zero_y_);
-                  p = xy;
+                  p = std::move(xy);
```
xiaoxq: Could you rename this single-char variable to a longer and meaningful name?

JasonZhou404: fixed

PR #8362, opened 2019-05-16
planning: exit condition of approach stage in pull-over scenario

jmtao:


modules/planning/conf/scenario/pull_over_config.pb.txt
```diff
@@ -1,6 +1,7 @@
 scenario_type: PULL_OVER
 pull_over_config: {
   start_pull_over_scenario_distance: 50.0
+  max_adc_stop_speed: 0.3
```
Capri2014: Is it the minimum chassis speed resolution? If so, try to use it from vehicle_param https://github.com/ApolloAuto/apollo/blob/0dad7437c1b5a05c1cdbc1b7ff3e96fd89693716/modules/common/configs/proto/vehicle_config.proto#L79 , since this would be different for different vehicles.

jmtao: good to know.  we use this parameter in quite a few scenarios. 
will clean them up in a later commit.

PR #8366, opened 2019-05-16
planning: adjust destination fence based on pull-over position

jmtao:


modules/planning/traffic_rules/destination.cc
```diff
@@ -77,62 +78,35 @@ int Destination::MakeDecisions(Frame* frame,
...
-    } else {
-      PullOver(&dest_point);
-      ADEBUG << "destination: PULL OVER";
+  if (FLAGS_enable_scenario_pull_over) {
+    const auto& pull_over_status =
+        PlanningContext::Instance()->planning_status().pull_over();
+    if (pull_over_status.exist_pull_over_position() &&
+        pull_over_status.has_pull_over_s()) {
+      // build stop decision based on pull-over position
+      AERROR << "BuildStopDecision: pull-over position";
```
Capri2014: This should be ADEBUG right?

modules/planning/traffic_rules/destination.cc
```diff
@@ -77,62 +78,35 @@ int Destination::MakeDecisions(Frame* frame,
...
-                            frame, reference_line_info);
-    ADEBUG << "destination: STOP at current lane";
   }
-  */
 
   // build stop decision
-  ADEBUG << "BuildStopDecision: destination";
-  std::string stop_wall_id = FLAGS_destination_obstacle_id;
-  const std::vector<std::string> wait_for_obstacle_ids;
+  AERROR << "BuildStopDecision: destination";
```
Capri2014: same for this one.

modules/planning/traffic_rules/destination.cc
jmtao: fixed. amended. thx

PR #8372, opened 2019-05-17
Planning: rule_based_stop task and fix side pass stop

JasonZhou404:


modules/planning/conf/scenario/side_pass_config.pb.txt
```diff
@@ -0,0 +1,61 @@
+scenario_type: SIDE_PASS
```
YajiaZhang: This is deprecated. Remove the file.

PR #8386, opened 2019-05-17
Update the user manual of CoolHigh

lishuanghua:
Update the user manual of CoolHigh


docs/specs/apollo_3.0_development_kits_user_manual_cn.md
```diff
@@ -569,32 +570,56 @@ rostopic echo /apollo/sensor/camera/traffic/image_short
...
 
-**设备连接说明**：数据输出端口通过以太网连接到工控机以太网口。
+ - 数据口：通过以太网线缆将接口盒连接到IPC。
+ - 12V电源接口：可以使用雷达自带的12V适配器进行供电或者使用自己的电源线连接12V电源供电。
+ - GPS授时接口：需要将GPS的授时接口与激光雷达接口盒的授时接口进行连接。
+ 
+### 激光雷达的配置及启动
 
-控制盒接口如图所示：
+ - 设置激光雷达ip和IPC的ip处在相同号段：雷达出厂默认ip地址为192.168.1.201，将IPC的ip地址修改为静态ip且与雷达ip地址处于相同的号段，例如192.168.1.xxx，如下图所示：
```
Capri2014: I thought our practice is to reset the lidar ip to be the same range of IPC ip instead of doing it opposite way? Correct me if I ma wrong.

lishuanghua: I have modified it already, please review it again. think you !

PR #8390, opened 2019-05-17
Planning: add SolverDefaultSettings to piecewise_jerk_problem

HongyiSun:


modules/planning/math/piecewise_jerk/piecewise_jerk_problem.h
```diff
@@ -98,6 +98,8 @@ class PiecewiseJerkProblem {
                                          std::vector<c_float>* lower_bounds,
                                          std::vector<c_float>* upper_bounds);
 
+  virtual void SolverDefaultSettings(OSQPSettings* settings);
```
xiaoxq: Best practice for this API: `OSQPSettings SolverDefaultSettings() const;`.
It leverages RVO well: https://en.wikipedia.org/wiki/Copy_elision#Return_value_optimization

xiaoxq: https://abseil.io/tips/11

PR #8395, opened 2019-05-18
record file flushes

Fla3inH0tCheet0s:
## Description
If cyber_recorder records a lot of data a plain Linux system, the whole system may freeze. It helps to flush periodically.

According to https://ext4.wiki.kernel.org/index.php/Ext3_Data%3DOrdered_vs_Data%3DWriteback_mode
> a careful application or library should always use fsync() at points where the application is at a stable commit point.

Without `fsync(fd_)`, record_file_integration_test will fail. record_file_integration_test is not run normally because it may wear out the disk and it assumes that the system is not running anything else.

## Test
```
sudo apt install valgrind
./apollo.sh citest_extended
ls bazel-bin/cyber/record/record_file_integration_test
# Assert that the binary is missing
valgrind --leak-check=full --show-leak-kinds=all bazel-bin/cyber/record/record_file_test
# Only errors from fastrtps
bazel test -c opt //cyber/record:record_file_integration_test
# Assert passes
valgrind --leak-check=full --show-leak-kinds=all bazel-bin/cyber/record/record_file_integration_test
# test fails, but only valgrind errors are from fastrtps
```


baowanyu:
I'm analyzing this case — small message and High frequency. Will update later...

baowanyu:
“fsync” is a synchronized operation, it is blocking other operations in record flush thread. 

If “fsync” operation is added, only 1 to 2 million messages can be written in 1,000 sample times in your test case. The original code can write 6 to 8 million messages in 1,000 sample times in your test case.

So far, recorder program will not call fsync operation directly, OS will perform this operation by itself.

Fla3inH0tCheet0s:
> OS will perform this operation by itself

The operating system seems to block other real time threads without the `fsync`. If you take a look at `cyber/record/file/record_file_integration_test.cc`, there is a thread that needs to run in real time. You can imagine this thread is the chassis send a CAN message at a 10ms period. This thread does not use the disk. When the disk is being written heavily, the system does not schedule the thread. The thread can take 100ms or more to run.

Please run
```
bazel test -c opt //cyber/record:record_file_integration_test
```
with and without this line https://github.com/ApolloAuto/apollo/pull/8395/files#diff-7a8b066d4d339903a00e855910bad719R159.

freeHackOfJeff:
Please don't use fsync(2) and fdatasync(2) if we need to process data in parallel, no mater in ordered or writeback mode for ext3/4.  both system call only fire data flushing operation to disk(fdatasysnc(2) also update timestamp), but not really ensure file data/metadata is really written into the physical drive. some data may still be lost in case of system/filesystem crash to keep the consistent of filesystem.

Fla3inH0tCheet0s:
Thanks @freeHackOfJeff for the additional context.

The `fsync` command is a workaround for not being able to schedule threads properly when writing large topics to disk. For example, writing camera and lidar data could cause timeouts in other parts of the system like the chassis module. Has anyone seen this behavior?

Fla3inH0tCheet0s:
I removed the `fsync` call. It sounds like it is not right to do this in the user space.

The root cause to the problem that I'm experiencing may be system related. When I ran my benchmark on another Linux system, it worked fine. According to https://bugzilla.kernel.org/show_bug.cgi?id=111181, it may be related to Samusung SSD's and Nvidia GPUs.

If I find a better solution, I will post it here.

jinghaomiao:
@baowanyu  Please review this ticket.

Fla3inH0tCheet0s:
The original issue may have been due to a system issue. Upgrading to Ubuntu 18.04 helped out the scheduler, when writing to disk. Flush does help the system, but the record file sometimes becomes sluggish leading to gaps in the record file.

Since the focus is different, I created a new PR with a different name here: https://github.com/ApolloAuto/apollo/pull/9108.

PR #8398, opened 2019-05-19
Tools: smart recorder v2 - realtime processor

michael4338:


modules/common/adapters/adapter_gflags.cc
```diff
@@ -208,3 +208,5 @@ DEFINE_string(rtcm_data_topic, "/apollo/sensor/gnss/rtcm_data",
 DEFINE_string(data_topic, "/apollo/data", "Data topic.");
 DEFINE_string(tf_topic, "/tf", "Transform topic.");
 DEFINE_string(tf_static_topic, "/tf_static", "Transform static topic.");
+DEFINE_string(recorder_status_topic, "/apollo/recorder/status",
```
xiaoxq: Let's set it as `/apollo/data/recorder/status`

modules/data/tools/smart_recorder/interval_pool.cc
```diff
@@ -25,6 +25,17 @@ namespace data {
 
 IntervalPool::IntervalPool() {}
 
+void IntervalPool::AddInterval(const Interval interval) {
```
xiaoxq: `const Interval&`

modules/data/tools/smart_recorder/proto/smart_recorder_status.proto
```diff
@@ -0,0 +1,17 @@
...
+  RECORDING = 0;
+  TERMINATING = 1;
+  STOPPED = 2;
+}
+
+message SmartRecorderStatus {
+  optional apollo.common.Header header = 1;
+  optional RecordingState recording_state = 2;
+  optional string state_message = 3;
+}
```
xiaoxq: One blank row at file end.

modules/data/tools/smart_recorder/proto/smart_recorder_status.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto2";
+
+package apollo.data;
+
+import "modules/common/proto/header.proto";
+
+enum RecordingState {
+  RECORDING = 0;
+  TERMINATING = 1;
+  STOPPED = 2;
```
xiaoxq: In proto enum, the first value will be the default if the field is not set.
Maybe better to put `STOPPED` as the default or something like `UNKNOWN`.

modules/data/tools/smart_recorder/realtime_record_processor.cc
```diff
@@ -0,0 +1,269 @@
...
+using cyber::common::EnsureDirectory;
+using cyber::common::GetFileName;
+using cyber::common::PathExists;
+using cyber::CreateNode;
+using cyber::record::HeaderBuilder;
+using cyber::record::Recorder;
+using cyber::record::RecordMessage;
+using cyber::record::RecordFileReader;
+using cyber::record::RecordReader;
+using cyber::record::RecordViewer;
```
xiaoxq: Put aliases into the anonymous namespace.

modules/data/tools/smart_recorder/realtime_record_processor.cc
```diff
@@ -0,0 +1,269 @@
...
+  Header* status_headerpb = status.mutable_header();
+  status_headerpb->set_timestamp_sec(cyber::Time::Now().ToSecond());
+  status.set_recording_state(state);
+  status.set_state_message(message);
+  AINFO << "send message with state " << state << ", " << message;
+  recorder_status_writer_->Write(status);
+}
+
+bool RealtimeRecordProcessor::GetNextValidRecord(
+    std::string* const record_path) const {
```
xiaoxq: The first `const` is not necessary, as we forbid pointer redirect anywhere.

PR #8399, opened 2019-05-19
Control: refactor state/trajectory Coord Transform and design Look-Ah…

kevin-y-wang:
…ead controller

### DIFF
--- a/modules/control/conf/control.conf
+++ b/modules/control/conf/control.conf
@@ -8,6 +8,9 @@
 --use_cyber_time=false
 --enable_slope_offset=false
 
---coordinate_transform_to_com=false
+--state_transform_to_com_reverse=false
+--state_transform_to_com_drive=true
+--trajectory_transform_to_com_reverse=false
+--trajectory_transform_to_com_drive=true
 # --reverse_heading_vehicle_state=false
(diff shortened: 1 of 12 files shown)


modules/control/conf/control.conf
```diff
@@ -8,6 +8,9 @@
 --use_cyber_time=false
 --enable_slope_offset=false
 
---coordinate_transform_to_com=false
+--state_transform_to_com_reverse=false
+--state_transform_to_com_drive=true
```
Capri2014: This may cause potential problem, since VehicleState is shared by quite a few modules, and whenever this is on, the planning algorithms would be affected as well.

Capri2014: @kevin-y-wang have you tried turn this on and count if any failure cases for planning test with perfect control?

kevin-y-wang: So far the Public 206 Cases (100% passed) and Open-Space 128 Cases (~45 passed) by echo_lincoln model show the current state/trajectory coordination works fine in this small scope.
I will compare the Enable 1762 cases on both perfect_control and echo_lincoln model to verify the current coordination

PR #8410, opened 2019-05-20
planning: open space fix dual warm up fail

deidaraho:


Capri2014:
@deidaraho Better to move the handling of dual variable warm up failure out to https://github.com/ApolloAuto/apollo/blob/master/modules/planning/open_space/trajectory_smoother/dual_variable_warm_start_problem.cc , otherwise you may need to handle this failure twice (In both OSQP/IPOPT interfaces, and potentially more).

PR #8411, opened 2019-05-20
Planning: update pull-over config.

panjiacheng:


modules/planning/tasks/deciders/path_assessment_decider/path_assessment_decider.cc
```diff
@@ -60,12 +60,13 @@ Status PathAssessmentDecider::Process(
   } else {
     ADEBUG << "There are " << candidate_path_data.size() << " candidate paths";
   }
+  auto end_time0 = std::chrono::system_clock::now();
```
Capri2014: use const reference here.

modules/planning/tasks/deciders/path_assessment_decider/path_assessment_decider.cc
```diff
@@ -60,12 +60,13 @@ Status PathAssessmentDecider::Process(
...
 
   // 1. Remove invalid path.
   std::vector<PathData> valid_path_data;
   for (const auto& curr_path_data : candidate_path_data) {
-    RecordDebugInfo(curr_path_data, curr_path_data.path_label(),
-                    reference_line_info);
+    // RecordDebugInfo(curr_path_data, curr_path_data.path_label(),
```
Capri2014: Why comment this out?

modules/planning/tasks/deciders/path_assessment_decider/path_assessment_decider.cc
```diff
@@ -76,6 +77,10 @@ Status PathAssessmentDecider::Process(
       }
     }
   }
+  auto end_time1 = std::chrono::system_clock::now();
```
Capri2014: same here, use const reference.

modules/planning/tasks/deciders/path_assessment_decider/path_assessment_decider.cc
panjiacheng: Only need to turn this on when debugging. Otherwise, too many paths will be displayed.

Capri2014: @panjiacheng There is a gflag to turn on/off debug in case you want it. https://github.com/ApolloAuto/apollo/blob/d8182ee0758d8fb5ef7e5acd75aa9538b87e3316/modules/planning/common/planning_gflags.cc#L506

panjiacheng: Got it. Thx. Since Jinyun needs this pr. I will merge first then update later.

PR #8413, opened 2019-05-20
Planning: open space roi interface for pull over (code complete)

JasonZhou404:


modules/planning/proto/open_space_roi_decider_config.proto
```diff
@@ -3,23 +3,30 @@ syntax = "proto2";
...
+    PULL_OVER = 2;
+    NARROW_STREET_U_TURN = 3;
+  }
+  optional RoiType roi_type = 1;
   // longitudinal range of parking roi
-  optional double roi_longitudinal_range = 1 [default = 10.0];
+  optional double roi_longitudinal_range = 2 [default = 10.0];
   // parking spot range detection threshold
-  optional double parking_start_range = 2 [default = 7.0];
+  optional double parking_start_range = 3 [default = 7.0];
```
Capri2014: @JasonZhou404 Just add a TODO here to remind ourselves that this might need to either merge or differentiate with the ones from senarios. (I.e, PullOver), as of now we set similar things twice in Scenarios level and in Core algorithm level.

modules/planning/tasks/deciders/open_space_decider/open_space_roi_decider.cc
```diff
@@ -53,29 +55,53 @@ Status OpenSpaceRoiDecider::Process(Frame *frame) {
...
+  std::array<Vec2d, 4> spot_vertices;
+  Path nearby_path;
+
+  const auto &roi_type = config_.open_space_roi_decider_config().roi_type();
+  if (roi_type == OpenSpaceRoiDeciderConfig::PARKING) {
+    const auto &routing_request =
+        frame->local_view().routing->routing_request();
+
+    if (routing_request.has_parking_space() &&
+        routing_request.parking_space().has_id()) {
```
xiaoxq: Not good practice to check `has_field`. `routing_request.parking_space().id().id().empty()` is better.

PR #8427, opened 2019-05-21
Planning: used the State design pattern to switch  scenes of the PublicRoadPlanner.

davidhopper2003:
Add a virtual function `UpdateScenarioType` to the `Scenario` class and use the State design pattern to switch  scenes to avoid tedious `switch-case` statements. If a scene is switched to the `LANE_FOLLOW` scene directly after it has been finished, you don't need to override the virtual function `UpdateScenarioType`, otherwise override it and add your scene switching logic.


davidhopper2003:
@jmtao 
Apollo currently uses the Strategy design pattern switching scenario, and my implementation is the State design pattern. Honestly, the two patterns are pretty similar in practice, except that the scenes are switched in different ways. I don't think my implementation is better. It's just my idea for your reference. Thanks for your review.

PR #8438, opened 2019-05-21
planning: (1) no PULL_OVER while around junctions; (2) create destination fence based on pull_over x/y instead of s/l

jmtao:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -165,12 +165,31 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectPullOverScenario(
          << "] destination_s[" << dest_sl.s() << "] adc_front_edge_s["
          << adc_front_edge_s << "]";
 
-  const bool pull_over_scenario =
+  bool pull_over_scenario =
```
YajiaZhang: Maybe naming to is_pull_over_scenario.

modules/planning/scenarios/scenario_manager.cc
```diff
@@ -165,12 +165,31 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectPullOverScenario(
...
+  if (pull_over_scenario) {
+    constexpr double kDisanceToAvoidJunction = 8.0;  // meter
+    for (const auto& overlap : first_encountered_overlap_map_) {
+      if (overlap.first == ReferenceLineInfo::PNC_JUNCTION ||
+          overlap.first == ReferenceLineInfo::SIGNAL ||
+          overlap.first == ReferenceLineInfo::STOP_SIGN ||
+          overlap.first == ReferenceLineInfo::YIELD_SIGN) {
+        const double distance_to = overlap.second.start_s - dest_sl.s();
+        const double distance_passed = dest_sl.s() - overlap.second.end_s;
+        if (distance_to > 0 && distance_to < kDisanceToAvoidJunction ||
```
YajiaZhang: Minor thing, if distance_to is a double, 0 will do an implicit type conversion from int to double. To avoid this, 0.0 will be preferred.

PR #8441, opened 2019-05-22
planning: add time matching buffer to accout for numerical errors

YajiaZhang:


modules/planning/common/planning_gflags.cc
```diff
@@ -325,7 +325,7 @@ DEFINE_bool(enable_follow_accel_constraint, true,
 DEFINE_bool(enable_sqp_solver, true, "True to enable SQP solver.");
 
 /// thread pool
-DEFINE_uint32(max_planning_thread_pool_size, 15,
+DEFINE_uint64(max_planning_thread_pool_size, 15,
```
YajiaZhang: For better IDE support ...

PR #8443, opened 2019-05-22
Planning: add pull over stop in open space pre stop decider

JasonZhou404:


modules/planning/proto/open_space_pre_stop_decider_config.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto2";
+
+package apollo.planning;
+
+message OpenSpacePreStopDeciderConfig {
+  // roi scenario definitions
+  enum StopType {
```
Capri2014: Looks like some of the enums can be shared or merged?

PR #8445, opened 2019-05-22
Tools: add script for calling smart_recorder

michael4338:


scripts/record_message.py
```diff
@@ -0,0 +1,154 @@
...
+        """
+        reuse_pool_dir = os.path.join(disk, 'data/bag', 'ReusedRecordsPool')
+        task_id = datetime.datetime.now().strftime('%Y-%m-%d-%H-%M-%S')
+        task_dir = os.path.join(disk, 'data/bag', task_id)
+        print('Recording bag to {}'.format(task_dir))
+        log_file = '/apollo/data/log/smart_recorder.out'
+        cmd = '''
+            source /apollo/scripts/apollo_base.sh
+            source /apollo/framework/install/setup.bash
+            nohup smart_recorder --source_records_dir={} --restored_output_dir={} > {} 2>&1 &
```
xiaoxq: Please call it with absolute path, instead of changing the PATH env.

PR #8459, opened 2019-05-22
control: publish target point position

sjiang2018:


modules/control/proto/control_cmd.proto
```diff
@@ -147,6 +147,9 @@ message SimpleLateralDebug {
   // term for control usage
   optional double lateral_error_feedback = 28;
   optional double heading_error_feedback = 29;
+
+  optional double target_point_x = 30;
+  optional double target_point_y = 31;
```
xiaoxq: Better to use single message field like `PathPoint target_point = 30;`

modules/control/proto/control_cmd.proto
```diff
@@ -147,6 +153,8 @@ message SimpleLateralDebug {
   // term for control usage
   optional double lateral_error_feedback = 28;
   optional double heading_error_feedback = 29;
+
+  optional CurrentTargetPoint current_target_point = 30;
```
xiaoxq: I think you can just reuse the type defined in modules/common.

sjiang2018: Fixed. Thank you.

PR #8480, opened 2019-05-23
Planning: add loading for steer in open space planner hybrid Astar

Capri2014:


modules/planning/open_space/coarse_trajectory_generator/hybrid_a_star.cc
```diff
@@ -449,22 +449,32 @@ bool HybridAStar::CombinePathAndSpeedProfile(
...
-    AWARN << "path data is empty";
+    AERROR << "path data is empty";
     return false;
   }
 
-  for (double cur_rel_time = 0.0; cur_rel_time < speed_data.TotalTime();
+  const double kEpsilon = 1e-6;
+
+  for (double cur_rel_time = 0.0;
+       cur_rel_time < speed_data.TotalTime() + kEpsilon;
```
xiaoxq: Calculate the upper bound once and keep the result in variable, to make it shorter and more efficient.

modules/planning/common/path/discretized_path.cc
```diff
@@ -52,10 +52,10 @@ PathPoint DiscretizedPath::Evaluate(const double path_s) const {
                                                            *it_lower, path_s);
 }
 
-std::vector<PathPoint>::const_iterator DiscretizedPath::QueryLowerBound(
-    const double path_s) const {
+std::vector<PathPoint>::const_iterator
+DiscretizedPath::QueryLowerBoundbyAbsoluteS(const double path_s) const {
   auto func = [](const PathPoint &tp, const double path_s) {
-    return tp.s() < path_s;
+    return std::fabs(tp.s()) < std::fabs(path_s);
```
YajiaZhang: This will not work when both positive and negative s exist. std::lower_bound requires a sorted container using the comparator.

PR #8489, opened 2019-05-24
planning: moved consts in dp_st_cost to gflags or protos

sjiang2018:


modules/planning/common/speed/st_boundary.cc
```diff
@@ -280,7 +280,7 @@ bool STBoundary::GetUnblockSRange(const double curr_time, double* s_upper,
   CHECK_NOTNULL(s_upper);
   CHECK_NOTNULL(s_lower);
 
-  *s_upper = FLAGS_decision_horizon;
+  *s_upper = FLAGS_speed_lon_decision_horizon;
```
HongyiSun: s_upper is decision_horizon instead of speed_lon

modules/planning/proto/dp_st_speed_config.proto
```diff
@@ -37,4 +37,8 @@ message DpStSpeedConfig {
   // other constraint
   optional double max_acceleration = 60 [default = 4.5];
   optional double max_deceleration = 61 [default = -4.5];
+
+  // buffer
+  optional double safe_time_buffer = 62 [default = 3.0];
```
HongyiSun: should start from 70

sjiang2018: Fixed. Thank you.

PR #8497, opened 2019-05-24
Planning&Common: update evaluation function and update combine path/speed in hybrid A*

Capri2014:


Capri2014:
@YajiaZhang Sorry I close the previous PR accidentally, to reply to your comments "This will not work when both positive and negative s exist. std::lower_bound requires a sorted container using the comparator." --- Sorry I forgot the stitching part would have both negative/positive `s`. And one thing is that the original implementation can not handle this case either.  -- But I changed my implementation now and add new interface so this can be used without  affecting the original implementation, and later we can probably talk about how to unify and simply the interfaces.

PR #8545, opened 2019-05-29
planning: keep pull-over stop fence until new routing/destination is set

jmtao:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -910,10 +910,36 @@ void ScenarioManager::UpdatePlanningContextTrafficLightScenario(
...
+      const auto& routing_end =
+          *(routing->routing_request().waypoint().rbegin());
+      reference_line.XYToSL({routing_end.pose().x(), routing_end.pose().y()},
+                            &dest_sl);
+
+      common::SLPoint pull_over_sl;
+      reference_line.XYToSL({pull_over_status.x(), pull_over_status.y()},
+                            &pull_over_sl);
+
+      constexpr double kDestMaxDelta = 30.0;  // meter
```
Capri2014: @jmtao In PathBoundDecider there is also a `pull_over_destination_to_pathend_buffer` in us as in https://github.com/ApolloAuto/apollo/blob/abe92a3111dda9c7e8fd903997424d1161115e31/modules/planning/tasks/deciders/path_bounds_decider/path_bounds_decider.cc#L439 , will these be of conflict or is it better if we can merge these two?

PR #8555, opened 2019-05-30
Prediction: process and normalize the feature_map tensor in junction_map_evaluator

HongyiSun:


modules/prediction/common/semantic_map.cc
```diff
@@ -68,11 +68,14 @@ void SemanticMap::RunCurrFrame(
   }
 
   // Crop ego_vehicle for demo
-  cv::Mat output_img =
-      CropByHistory(obstacle_id_history_map_.at(-1), cv::Scalar(0, 0, 255));
-  cv::namedWindow("Demo window", cv::WINDOW_NORMAL);
-  cv::imshow("Demo window", output_img);
-  cv::waitKey();
+  if (false) {
```
kechxu: use a gflag or add a comment to this disabling

PR #8560, opened 2019-05-30
cyber: cyber_node variable typo

mickeyouyou:


cyber/tools/cyber_node/cyber_node
```diff
@@ -15,27 +15,27 @@
 # limitations under the License.
 # ****************************************************************************
 
+from cyber_py import cyber
```
xiaoxq: Put standard libs first, per PEP8 requirement.

PR #8578, opened 2019-05-31
Prediction: update feature extraction.

panjiacheng:


modules/prediction/proto/lane_graph.proto
```diff
@@ -36,6 +36,10 @@ message LaneSegment {
   optional double end_s = 3 [default = 0.0];
   optional uint32 lane_turn_type = 4 [default = 0];
   repeated LanePoint lane_point = 5;
+  // If this lane-segment is the one that's closest to ADC,
+  // then adc_s gives the s of the ADC.
+  optional double adc_s = 7 [default = 0.0];
+  optional int32 adc_lane_point_idx = 8 [default = 0];
```
xiaoxq: You don't need to provide default value for all fields. Leverage its own default:
https://developers.google.com/protocol-buffers/docs/proto#optional

PR #8583, opened 2019-05-31
Planning: optimized the GetFirstOverlap function of the ReferenceLineInfo class to eliminate unnecessary object copy operations.

davidhopper2003:


modules/planning/common/reference_line_info.cc
```diff
@@ -219,15 +219,22 @@ bool ReferenceLineInfo::GetFirstOverlap(
...
+      overlap_min_s_iter = iter;
+      overlap_min_s = iter->start_s;
     }
   }
+
+  // Ensure that the path_overlaps is not empty.
+  if (overlap_min_s_iter != path_overlaps.end()) {
+    *path_overlap = *overlap_min_s_iter;
+  }
+
```
jmtao: If the purpose of this change is just to avoid multiple "copy" action in line 227, you can just create a local pointer (hdmap::PathOverlap*), and set it to &overlap at line 227, and then at the end, copy it to *path_overlap if not nullptr.

no need to change so many lines of code.

davidhopper2003: @jmtao Thanks. First, I prefer to use iterators so that the C++ STL style is more obvious. Second,

davidhopper2003: Thanks once more for your comment.

PR #8604, opened 2019-06-01
Bridge: add head classes for udp frames

muleisheng:


modules/bridge/common/bridge_header.h
```diff
@@ -0,0 +1,107 @@
...
+  void SetMsgSize(size_t msg_size) { msg_size_ = msg_size; }
+
+ private:
+  template <typename T, size_t S>
+  char *SerializeBasicType(const T *value, char *buf, size_t size) {
+    if (!buf || size < S) {
+      return nullptr;
+    }
+    header_size_ += S + 1;
+    char *p = buf;
```
xiaoxq: Try to avoid single-char variable name. Make it meaningful.

PR #8634, opened 2019-06-04
autopep8 on all python files

Fla3inH0tCheet0s:
It might help other contributors to use `autopep8`. This changeset was created in `bash` with
```
find . -name '*.py' -exec autopep8 --in-place --aggressive --aggressive '{}' \;
```
It would probably be best to add this to the robot actions.


yaoweifeng01:
@Fla3inH0tCheet0s  Please use autopep8 --ignore=E266 --in-place --aggressive --aggressive. ignore E266 at least, because ## is a flag of doxygen which we will use. And can this format be added in Docker?

modules/tools/vehicle_calibration/plot_results.py
```diff
@@ -34,23 +34,23 @@
...
     for line in f:
-        items = line.split(',')
-        cmd = round(float(items[0]))
-        speed = float(items[1])
-        acc = float(items[2])
-        if cmd in cmd_table:
-            speed_table = cmd_table[cmd]
-            if speed in speed_table:
-                speed_table[speed].append(acc)
+            items = line.split(',')
```
xiaoxq: This indent is wrong.

Fla3inH0tCheet0s: It's resolved. autopep8 may have gotten confused, but I think python would run it the same: https://github.com/hhatto/autopep8/issues/486.

Fla3inH0tCheet0s:
@yaoweifeng01 That's fine.

It's looks like documentation strings are currently used, but `##` allow for special commands in doxygen: http://www.doxygen.nl/manual/docblocks.html. `--ignore=E266` makes no difference for now, but in the future it may help doxygen.

modules/tools/prediction/data_pipelines/data_preprocessing/generate_junction_labels.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 ###############################################################################
 
+from online_to_offline import LabelGenerator
```
xiaoxq: Ditto

PR #8638, opened 2019-06-04
Control: refactor the usage of the standstill acceleration for lon_co…

kevin-y-wang:
…ntroller

### DIFF
--- a/modules/control/proto/control_conf.proto
+++ b/modules/control/proto/control_conf.proto
@@ -61,4 +61,6 @@ message ControlConf {
 
   optional double query_relative_time = 35;
   optional double minimum_speed_protection = 36;
+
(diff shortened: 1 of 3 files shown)


modules/control/proto/control_conf.proto
```diff
@@ -61,4 +61,6 @@ message ControlConf {
 
   optional double query_relative_time = 35;
   optional double minimum_speed_protection = 36;
+
+  optional double max_path_remain_when_stopped = 37 [default = 0.3];
```
Capri2014: Did you miss a proto filed here?

Capri2014: Hoops, my bad

PR #8643, opened 2019-06-04
Localization: Update local-map & Add tutorial for ndt localization module

tthhee:


modules/localization/conf/localization.conf
```diff
@@ -211,3 +211,7 @@
 # need cpu to support avx intel intrinsics
 # default: false
 --if_use_avx=true
+
+--lidar_extrinsics_file=/apollo/modules/localization/msf/params/velodyne_params/velodyne64_novatel_extrinsics_example.yaml
+--map_dir=/apollo/Bag/Test_Data/Sunnyvale_Big_Loop
```
Capri2014: This `map_dir` should be overwritten or mismatch with the ones we selected from dreamview frontend.

modules/localization/msf/local_map/base_map/base_map_matrix.cc
```diff
@@ -1,5 +1,5 @@
 /******************************************************************************
- * Copyright 2017 The Apollo Authors. All Rights Reserved.
+ * Copyright 2018 The Apollo Authors. All Rights Reserved.
```
Capri2014: 2019

modules/localization/conf/localization.conf
jinghaomiao: Please use the map directory gflag globally across all modules to minimize confusion

tthhee: oh， sorry, it's my mistake.

PR #8650, opened 2019-06-04
Planning: pull-over bug fixes.

panjiacheng:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -195,6 +195,7 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectPullOverScenario(
     }
   }
 
+<<<<<<< HEAD
```
jmtao: fix this

panjiacheng: fixed. thx.

PR #8656, opened 2019-06-05
Control: adjust the re-init stitching conditions with road-test data

kevin-y-wang:


modules/planning/common/trajectory_stitcher.cc
```diff
@@ -57,9 +57,10 @@ std::vector<TrajectoryPoint>
 TrajectoryStitcher::ComputeReinitStitchingTrajectory(
     const double planning_cycle_time, const VehicleState& vehicle_state) {
   TrajectoryPoint reinit_point;
-  constexpr double kEpsilon = 0.2;
```
Capri2014: @kevin-y-wang Please add a TODO here, if localization publish corrected IMU acceleration data in both modes, we should decrease this to a smaller value.

kevin-y-wang: Fixed. Thanks

PR #8671, opened 2019-06-05
Planning: fixes on pull over open space boundary

JasonZhou404:


modules/planning/tasks/deciders/open_space_decider/open_space_roi_decider.cc
```diff
@@ -296,35 +282,57 @@ bool OpenSpaceRoiDecider::GetBoundary(
...
     check_point_s =
+        start_s +
         index *
-        config_.open_space_roi_decider_config().roi_linesegment_length();
+            config_.open_space_roi_decider_config().roi_linesegment_length();
     check_point_s = check_point_s >= end_s ? end_s : check_point_s;
     last_check_point_heading = check_point_heading;
   }
 
+  DCHECK_EQ(right_lane_boundary->size(), left_lane_boundary->size());
```
Capri2014: Do not do CHECK here, return false if not match.

PR #8694, opened 2019-06-07
commit ncut for lidar segmentation

weidezhang:


modules/perception/base/point_cloud_util.cc
```diff
@@ -0,0 +1,112 @@
+/******************************************************************************
+ * Copyright 2018 The Apollo Authors. All Rights Reserved.
```
techoe: 2019

modules/perception/lidar/lib/segmentation/ncut/common/flood_fill.cc
```diff
@@ -0,0 +1,159 @@
+/******************************************************************************
+ * Copyright 2018 The Apollo Authors. All Rights Reserved.
```
techoe: 2019

PR #8701, opened 2019-06-07
Planning: slightly adjust curvature-realted bounds with physical limits

kevin-y-wang:


Capri2014:
I think in future we should set this one from one time calculation based on vehicle parameters. So this parameter can de-couple from the vehicle and automatically adjust when we select different vehicles. @YajiaZhang @yifeijiang

PR #8720, opened 2019-06-08
count of objects "alive" in pool

storypku:


storypku:
Anybody have idea on how to pass the Travis CI check?

storypku:
It seems that Both node and count play a role in Head when atomic_compare. Meanwhile, Counts of Objects alive in pool have little practical use other than debugging. So, close this pr.

PR #8723, opened 2019-06-09
Bridge: sender and receiver implementation

muleisheng:


modules/bridge/common/bridge_proto_buf.h
```diff
@@ -0,0 +1,160 @@
...
+#include <string>
+#include <memory>
+#include "modules/bridge/common/macro.h"
+#include "modules/bridge/common/bridge_header.h"
+
+namespace apollo {
+namespace bridge {
+
+const size_t FRAME_SIZE = 1024;
+
```
jinghaomiao: 用constexpr来定义

muleisheng: 收到

PR #8729, opened 2019-06-10
Add CoolHigh canbus code to branch master

lishuanghua:
Add CoolHigh canbus code to branch master


modules/canbus/proto/BUILD
```diff
@@ -19,6 +19,7 @@ proto_library(
         "vehicle_parameter.proto",
         "wey.proto",
         "zhongyun.proto",
+	"ch.proto",
```
Capri2014: indent

PR #8731, opened 2019-06-10
All: replaced DCHECK_NOTNULL with CHECK_NOTNULL.

davidhopper2003:
DCHECK_NOTNULL detects whether a pointer is empty in Debug mode but doesn't work in Release mode.


davidhopper2003:
> use the return

Thank you. But I don't know what you mean.

PR #8750, opened 2019-06-11
Localization:  Add tutorial for map generation && fix some scripts.

zhouyao4321:


JackFu123:
Please check: 1)In preparation part, localization data from [Apollo Data Open Platform] should be specified. Because there are so many datasets on the page. 2)The "Multi-Sensor Fusion Localization Data" can't be downloaded successfully either from US or China. 3)Why is the dataset "US only"? Even in the Chinese tutorial?

zhouyao4321:
> Please check: 1)In preparation part, localization data from [Apollo Data Open Platform] should be specified. Because there are so many datasets on the page. 2)The "Multi-Sensor Fusion Localization Data" can't be downloaded successfully either from US or China. 3)Why is the dataset "US only"? Even in the Chinese tutorial?

1) The name of localization data has been specified.
2) The data is stored on a server in Hong Kong from Baidu Cloud. Maybe it will be slow if downloaded from the US. 
3) The demo data is collected from US, it contains high-precision map data. There may be some legal risks if we do not limit the open area. The Chinese document is for native Chinese speakers living in the US.

PR #8757, opened 2019-06-11
Velodyne: conf path is incorrect

techoe:
The path was incorrect. velodyne128_compensator_conf.pb.txt does not exist.


techoe:
> velodyne128_fusion_compensator.pb.txt is consistent

@fengliu00 , Thank you for the review! I agree with you. I changed it to velodyne128_fusion_compensator.pb.txt

PR #8762, opened 2019-06-12
Control: set trajectory v=a=0 only when v< min_resolution and a< max_…

kevin-y-wang:
…acc_when_stop


modules/control/control_component.cc
```diff
@@ -371,7 +371,9 @@ Status ControlComponent::CheckInput(LocalView *local_view) {
   for (auto &trajectory_point :
        *local_view->trajectory.mutable_trajectory_point()) {
     if (std::abs(trajectory_point.v()) <
-        control_conf_.minimum_speed_resolution()) {
+            control_conf_.minimum_speed_resolution() &&
```
sjiang2018: vehicle_param.max_abs_when_stopped

PR #8768, opened 2019-06-12
Control: add localization_pose message into record for Control Channel

kevin-y-wang:


modules/control/controller/lat_controller.cc
```diff
@@ -86,7 +86,9 @@ LatController::LatController() : name_("LQR-based Lateral Controller") {
   AINFO << "Using " << name_;
 }
 
-LatController::~LatController() { CloseLogFile(); }
+LatController::~LatController() {
+  CloseLogFile();
+}
```
xiaoxq: These are auto-adjusted by clang-format tool. No need to change.

kevin-y-wang: Fixed. Thanks.

kevin-y-wang:
Move the localization pose into input debug per comment. Thanks

modules/control/control_component.cc
```diff
@@ -219,6 +219,10 @@ Status ControlComponent::ProduceControlCommand(
     auto debug = control_command->mutable_debug()->mutable_input_debug();
     debug->mutable_localization_header()->CopyFrom(
         local_view_.localization.header());
+    debug->mutable_localization_pose()->mutable_position()->CopyFrom(
```
Capri2014: why copy them separately ?

PR #8774, opened 2019-06-12
Cyber: optimization record interface and add dox

baowanyu:


cyber/record/record_writer.h
```diff
@@ -35,48 +35,134 @@ namespace apollo {
...
+  * @brief Get channel list.
+  *
+  * @return List container with all channel name string.
+  */
+  std::set<std::string> GetChannelList() const override;
 
  private:
   bool WriteMessage(const proto::SingleMessage& single_msg);
   bool SplitOutfile();
+  bool IsNewChannel(const std::string& channel_name) const;
```
michael4338: why bring back to be private?

cyber/record/record_viewer.h
```diff
@@ -31,25 +31,68 @@ namespace apollo {
...
+  /**
+  * @brief Is this record reader is valid.
+  *
+  * @return True for valid, false for not.
+  */
   bool IsValid() const;
-  bool Update(RecordMessage* message);
-  uint64_t begin_time() const;
-  uint64_t end_time() const;
-  std::set<std::string> GetChannelList() const;
```
michael4338: the change can break existing callers.  have we checked the existing references and modify them accordingly

modules/data/tools/smart_recorder/realtime_record_processor.cc
```diff
@@ -155,7 +155,6 @@ bool RealtimeRecordProcessor::Process() {
       break;
     }
     auto reader = std::make_shared<RecordReader>(record_path);
-    reader->SetFlushMode(true);
     RecordViewer viewer(reader, 0, UINT64_MAX,
                         ChannelPool::Instance()->GetAllChannels());
     AINFO << "checking " << record_path << ": " << viewer.begin_time() << " - "
```
michael4338: this will be broken by the API change

baowanyu: OK , I will check it

cyber/record/record_writer.h
baowanyu: Since it has already checked whether is new channel in WriteChannel().
Anyway, I will reconfirm the change and push a new one.

michael4338: yea but it will print a warning log if existing inside of WriteChannel, which is really noisy if write large number of messages

baowanyu: OK，I move this method to public.

cyber/record/record_viewer.h
baowanyu: Fixed in new request.

PR #8777, opened 2019-06-12
planning: relax open space state ending constraints into cost function

deidaraho:


Capri2014:
Great! Also another thing worth thinking is since we now try the dynamic end constraint relaxation. Maybe we could further decouple and reformulate this problem into nominal trajectory + error state as well? Of course this is dangerous since our trajectory might be very highly non-linear and we do not do this properly. But we may have chance to significantly further reduce the calculation cost along this path.

PR #8791, opened 2019-06-13
tools: added plot of speed dsteering.

yifeijiang:


modules/tools/plot_planning/speed_dsteering_data.py
```diff
@@ -0,0 +1,89 @@
...
+        speed_mps = chassis.speed_mps
+        timestamp_sec = chassis.header.timestamp_sec        
+
+        if self.last_timestamp_sec is None:
+            self.last_steering_percentage = steering_percentage
+            self.last_speed_mps = speed_mps
+            self.last_timestamp_sec = timestamp_sec
+            return
+
+        if (timestamp_sec - self.last_timestamp_sec) > 0.02:
```
Capri2014: @yifeijiang I thought the chassis update rate is 100HZ? (0.01s)

PR #8793, opened 2019-06-13
Planning: add the option to use the qp_spline_reference_line generation algorithm.

panjiacheng:


JasonZhou404:
Why add this option? If for road testing, remember to turn on reference stitching as well

JasonZhou404:
Line 17. In non stitching mode qp smoother may be shaky in narrow roads

panjiacheng:
This is for road testing. The new reference line algorithm is buggy and fails for roundabouts. We may need to use the old one for road tests.

PR #8867, opened 2019-06-18
tools: added record reader

yifeijiang:


modules/tools/plot_planning/record_reader.py
```diff
@@ -0,0 +1,40 @@
...
+from modules.canbus.proto import chassis_pb2
+from modules.localization.proto import localization_pb2
+
+class RecordItemReader:
+    def __init__(self, record_file):
+        self.record_file = record_file
+    def read(self, topics):
+        reader = RecordReader(self.record_file)
+        for msg in reader.read_messages():
+            if msg.topic in topics and msg.topic == "/apollo/canbus/chassis":
```
xiaoxq: Check at beginning:
```
if msg.topic not in topics:
    continue
```

modules/tools/plot_planning/record_reader.py
```diff
@@ -0,0 +1,40 @@
...
+                chassis.ParseFromString(msg.message)
+                data = {"chassis": chassis}
+                yield data
+
+            if msg.topic in topics and msg.topic == "/apollo/localization/pose":
+                location_est = localization_pb2.LocalizationEstimate()
+                location_est.ParseFromString(msg.message)
+                data = {"pose": location_est}
+                yield data
+        return
```
xiaoxq: No need to return. I guess this is not allowed when there is `yield`.

PR #8884, opened 2019-06-19
planning: bug fixed in openspace fallback trajectory

sjiang2018:


modules/planning/tasks/deciders/open_space_decider/open_space_fallback_decider.cc
```diff
@@ -277,6 +280,8 @@ bool OpenSpaceFallbackDecider::IsCollisionFreeTrajectory(
   double ego_width = vehicle_config.vehicle_param().width();
   auto trajectory_pb = trajectory_gear_pair.first;
   const size_t point_size = trajectory_pb.NumOfPoints();
+  // TODO(SHU): find a proper buffer time
+  const double kCollisionTimeBuffer = 100 * FLAGS_trajectory_time_resolution;
```
Capri2014: please move this to configs in proto

PR #8897, opened 2019-06-20
planning: modified fallback logics

sjiang2018:


Capri2014:
Will hold until further debugged

PR #8899, opened 2019-06-20
 Planning: add the speed limit during switching Open-Space partitions

kevin-y-wang:


Capri2014:
Also add it into configs

kevin-y-wang:
Refactor finished. Thanks.

PR #8922, opened 2019-06-21
Common: add separate video channels

michael4338:


michael4338:
> Should we change configs in https://github.com/ApolloAuto/apollo/tree/master/modules/drivers/video/conf ?

those should be able to be overwritten when choose the vehicle

PR #8923, opened 2019-06-21
Dreamview: add parking space id

kiboliu:


modules/dreamview/frontend/src/renderer/map.js
```diff
@@ -540,11 +563,12 @@ export default class Map {
...
     shouldDrawTextOfThisElementKind(kind) {
-        // Ex: mapping 'lane' to 'showMapLaneId' option
-        const optionName = `showMap${kind[0].toUpperCase()}${kind.slice(1)}Id`;
-
-        // NOTE: return false if the option is not found
-        return STORE.options[optionName] === true;
+        // showMapLaneId option controls both laneId and parkingSpaceId
+        if (STORE.options['showMapLaneId']
+            && ['parkingSpace', 'lane'].includes(kind)) {
+                return true;
```
siyangy: indentation

modules/dreamview/frontend/src/renderer/map.js
```diff
@@ -213,6 +213,29 @@ export default class Map {
...
+        const text = this.textRender.drawText(parkingSpace.id.id, scene, colorMapping.WHITE);
+        const points = _.get(parkingSpace, 'polygon.point');
+        if (points && points.length >= 3 && text) {
+            const point1 = points[0];
+            const point2 = points[1];
+            const point3 = points[2];
+            let textPosition = {
+                x: (point1.x + point3.x) / 2,
+                y: (point1.y + point3.y) / 2,
+                z: 0.04
```
vlin17: is there a reason we don't want to use Z_OFFSET?

modules/dreamview/frontend/src/renderer/map.js
vlin17: how about "return STORE.options['showMapLaneId'] && ['parkingSpace', 'lane'].includes(kind)"?

PR #8942, opened 2019-06-24
add extra util for perception

weidezhang:


modules/perception/base/file_util.cc
```diff
@@ -0,0 +1,305 @@
...
+  return ifs.good()
+             ? static_cast<int>(count(istreambuf_iterator<char>(ifs),
+                                      istreambuf_iterator<char>(), '\n')) +
+                   1
+             : -1;
+}
+
+}  // namespace base
+}  // namespace perception
+}  // namespace apollo
```
xiaoxq: Remove redundant functions which are already in https://github.com/ApolloAuto/apollo/blob/master/cyber/common/file.h

modules/perception/base/noncopyable.h
```diff
@@ -0,0 +1,44 @@
...
+  void operator=(const TypeName&);
+// A macro to disallow all the implicit constructors, namely the
+// default constructor, copy constructor and operator= functions.
+//
+// This should be used in the private: declarations for a class
+// that wants to prevent anyone from instantiating it. This is
+// especially useful for classes containing only static methods.
+#define DISALLOW_IMPLICIT_CONSTRUCTORS(TypeName) \
+  TypeName();                                    \
+  DISALLOW_COPY_AND_ASSIGN(TypeName);
```
xiaoxq: Ditto. https://github.com/ApolloAuto/apollo/blob/master/cyber/common/macros.h

modules/perception/base/string_util.cc
```diff
@@ -0,0 +1,125 @@
...
+    return true;
+  } catch (...) {
+    AERROR << "std::stoi() failed.str:" << str;
+    return false;
+  }
+}
+
+}  // namespace base
+}  // namespace perception
+}  // namespace apollo
```
xiaoxq: Ditto. https://github.com/ApolloAuto/apollo/blob/master/modules/common/util/string_util.h

modules/perception/base/file_util.cc
weidezhang: yea, but i think we should move this out of cyber since it is not cybertron specific. shall we consolidate them together later ?

modules/perception/base/noncopyable.h
weidezhang: shall we move this out of cyber ?

modules/perception/base/string_util.cc
weidezhang: same as here

modules/perception/base/file_util.cc
weidezhang: i guess i will do the modification at cyber first and then try moving it out.

xiaoxq: It was ever in Apollo modules/common. But to follow the rules of:
1. No repeated code.
2. It is OK for Apollo to depend on Cyber, but Cyber should not depend on Apollo. So it's migrated into Cyber.
Sorry for the trouble, but as of now we are most likely to stick to this. If there are new utils, it's OK to add in perception module utils, but redundant functions are not allowed.

weidezhang: agreed. will modify based on cyber first

cyber/common/file.cc
```diff
@@ -322,6 +331,78 @@ std::string GetCurrentPath() {
...
+  }
+  // 文件，直接删除
+  if (type == TYPE_FILE) {
+    if (remove(filename.c_str()) != 0) {
+      AERROR << "failed to remove file: " << filename;
+      return false;
+    }
+    return true;
+  }
+  // 目录，需要迭代删除
```
jinghaomiao: Do not use Chinese in the comment

cyber/common/file.cc
```diff
@@ -322,6 +331,78 @@ std::string GetCurrentPath() {
...
+    AWARN << "failed to opendir: " << filename;
+    return false;
+  }
+  dirent* dir_info = NULL;
+  while ((dir_info = readdir(dir)) != NULL) {
+    if (strcmp(dir_info->d_name, ".") == 0 ||
+        strcmp(dir_info->d_name, "..") == 0) {
+      continue;
+    }
+    // 拼接目录路径
```
jinghaomiao: Same

cyber/common/file.cc
weidezhang: will resolve it in a separate commit

cyber/common/file.cc
weidezhang: will resolve it in a separate commit

cyber/common/file.cc
weidezhang: done fix

cyber/common/file.cc
weidezhang: done fix

PR #8943, opened 2019-06-24
Planning: update pull-over status info into the debug planning_internal.

panjiacheng:


modules/planning/proto/planning_internal.proto
```diff
@@ -198,6 +199,8 @@ message PlanningData {
   optional ScenarioDebug scenario = 26;
   optional OpenSpaceDebug open_space = 27;
   optional SmootherDebug smoother = 28;
+
+  optional apollo.planning.PullOverStatus pull_over_status = 29;
```
Capri2014: @panjiacheng @yifeijiang Should this be better under one of the scenario specific debug proto?

panjiacheng: Let's put it here. The pull-over-status is set by path-bounds decider, agnostic to scenarios. If in the future we have more scenarios, they might also ask path-bounds decider to find a pull-over position.

Capri2014: OK

PR #8737, opened 2019-06-10
Autopep8 prelim

Fla3inH0tCheet0s:
Let's break up https://github.com/ApolloAuto/apollo/pull/8634 into a manual piece and an automated piece. These changes are manual so that `autopep8` parses the files correctly. Syntax errors errors and the order of imports without `  # noqa` need to be taken into account before `autopep8` runs.

PR #8953, opened 2019-06-25
Cyber: Autopep8 prelim

Fla3inH0tCheet0s:
Continuing from https://github.com/ApolloAuto/apollo/pull/8737. This branch is rebased and the sys paths are relative.

### DIFF
--- a/cyber/python/examples/listener.py
+++ b/cyber/python/examples/listener.py
@@ -18,9 +18,9 @@
 
 import sys
 
-sys.path.append("../")
-from cyber_py import cyber
-from cyber.proto.unit_test_pb2 import ChatterBenchmark
+sys.path.append("/apollo/cyber/python")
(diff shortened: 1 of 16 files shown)


yaoweifeng01:
The cyber *.py has been formatted by command "autopep8 --ignore=E266,E265 --in-place --aggressive --aggressive".Please check it.

Fla3inH0tCheet0s:
@yaoweifeng01 

That works. I'm not sure if the changes in modules/tools/localization/evaluate_compare.py are still valid. For newer versions of autopep8, you will want to add `# noqa` to some of the imports so that `sys.path.append("../")` doesn't get reordered. autopep8 1.4.4 (pycodestyle: 2.5.0) reorders import paths. I guess a better solution would be to avoid the need of `sys.path.append` in the container.

PR #8955, opened 2019-06-25
Cyber: Make sure the accquired lock in segment will be release

KitFung:
I found a issue while my development base on the segment component.
The segment will not release it owned lock if a exception happen before manually release the lock.

For example
1. Some one accquired a write lock
2. A exception happen
3. The program shutdown
4. The accquired block have been locked forever

So I suggest adding a bool array to make sure all the accquired lock must be release in destructor.

Since the accquire and release to the same lock will only do by the same process serially, so we can use a non thread safe vector to impl it. So the cost of this change is minior.

### DIFF
--- a/cyber/transport/shm/segment.cc
+++ b/cyber/transport/shm/segment.cc
@@ -67,6 +67,7 @@ bool Segment::AcquireBlockToWrite(std::size_t msg_size,
   writable_block->index = index;
   writable_block->block = &blocks_[index];
   writable_block->buf = block_buf_addrs_[index];
+  acquired_write_lock_[index] = true;
   return true;
 }
 
@@ -76,6 +77,7 @@ void Segment::ReleaseWrittenBlock(const WritableBlock& writable_block) {
     return;
   }
   blocks_[index].ReleaseWriteLock();
+  acquired_write_lock_[index] = false;
 }
 
 bool Segment::AcquireBlockToRead(ReadableBlock* readable_block) {
@@ -106,6 +108,7 @@ bool Segment::AcquireBlockToRead(ReadableBlock* readable_block) {
   }
   readable_block->block = blocks_ + index;
   readable_block->buf = block_buf_addrs_[index];
+  acquired_read_lock_[index] = true;
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


ctwfei:
Hi KitFung,

Thank you very much for your suggestion and submission, but usually the program does not call the destructor when the program is terminated abnormally. So I don't think this submission can fix the issue.

KitFung:
> Hi KitFung,
> 
> Thank you very much for your suggestion and submission, but usually the program does not call the destructor when the program is terminated abnormally. So I don't think this submission can fix the issue.

Is there any example? Since I haven't meet that case yet, I may investigate that.

ctwfei:
> > Hi KitFung,
> > Thank you very much for your suggestion and submission, but usually the program does not call the destructor when the program is terminated abnormally. So I don't think this submission can fix the issue.
> 
> Is there any example? Since I haven't meet that case yet, I may investigate that.

Actually, I mean the case you mentioned above. If [the code](https://github.com/ApolloAuto/apollo/blob/master/cyber/transport/dispatcher/shm_dispatcher.cc/#L65-L82) throw  an exception and does not catch it or otherwise terminates without executing its destructors (e.g. the code crashes), so unlocking the lock in a destructor is not useful.

Fortunately, we haven't had a deadlock like this case so far, but we will try to fix the snake in the grass  later.

Thanks again for the case.

KitFung:
> > > Hi KitFung,
> > > Thank you very much for your suggestion and submission, but usually the program does not call the destructor when the program is terminated abnormally. So I don't think this submission can fix the issue.
> > 
> > 
> > Is there any example? Since I haven't meet that case yet, I may investigate that.
> 
> Actually, I mean the case you mentioned above. If [the code](https://github.com/ApolloAuto/apollo/blob/master/cyber/transport/dispatcher/shm_dispatcher.cc/#L65-L82) throw an exception and does not catch it or otherwise terminates without executing its destructors (e.g. the code crashes), so unlocking the lock in a destructor is not useful.
> 
> Fortunately, we haven't had a deadlock like this case so far, but we will try to fix the snake in the grass later.
> 
> Thanks again for pointing out the issue.

Owh, I found out my code change is handling interrupt instead of exception.
The interrupt will be catch by the signal handler, therefore the object will be destroy correctly.
While the std::abort or terminate will not do that.

The code work for me since most of the issue is caused by the interrupt while my testing.

I will close this mr, thanks for your notice.

PR #8956, opened 2019-06-25
 Planning: add s_curve config into iterative anchoring smoother

Capri2014:


modules/planning/conf/planner_open_space_config.pb.txt
```diff
@@ -9,7 +9,7 @@ warm_start_config {
   next_node_num: 10
   step_size: 0.5
   traj_forward_penalty: 1.0
-  traj_back_penalty: 1.0
+  traj_back_penalty: 1.5
```
JasonZhou404: Change it on purpose?

Capri2014: Yes, but I will put it into a separate commit later with other changes to help stabilize Hybrid Astar in manual mode.

modules/planning/proto/planner_open_space_config.proto
```diff
@@ -145,6 +145,7 @@ message IterativeAnchoringConfig {
   optional double max_reverse_acc = 13 [default = 2.0];
   optional double max_acc_jerk = 14 [default = 4.0];
   optional double delta_t = 15 [default = 0.2];
+  optional PiecewiseJerkSpeedConfig s_curve_config = 16;
```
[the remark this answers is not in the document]
Capri2014: Yes, I agree with you. A better solution would be to separate qp setting  and put it in a separate config proto. But that requires refactor from bottom up. Will do that in later iteration.

PR #8959, opened 2019-06-25
Planning: disable curve constraint and estimate bound in open space p…

Capri2014:
…lanner


Capri2014:
Yes it looks strange, I run it again with high pass rate with control in loop. Previous run might be affected by the simulation publishing new release. Will close this as of now.

PR #8975, opened 2019-06-25
Prediction: dump obstacle conf info to prediction results

kechxu:


modules/prediction/predictor/predictor_manager.cc
```diff
@@ -322,7 +322,8 @@ void PredictorManager::PredictObstacle(
       obstacle->latest_feature().priority());
   prediction_obstacle->set_is_static(obstacle->IsStill());
   if (FLAGS_prediction_offline_mode == 3) {
-    FeatureOutput::InsertPredictionResult(obstacle->id(), *prediction_obstacle);
+    FeatureOutput::InsertPredictionResult(obstacle->id(), *prediction_obstacle,
+        obstacle->obstacle_conf());
```
xiaoxq: Align with first param.

modules/prediction/predictor/predictor_manager.cc
```diff
@@ -322,7 +322,8 @@ void PredictorManager::PredictObstacle(
       obstacle->latest_feature().priority());
   prediction_obstacle->set_is_static(obstacle->IsStill());
   if (FLAGS_prediction_offline_mode == 3) {
```
xiaoxq: Name the magic number:
```
static constexpr int XXX_MODE = 3;
if (FLAGS_prediction_offline_mode == XXX_MODE) {
}
```

PR #8978, opened 2019-06-26
Dreamview: show pull over status

kiboliu:
Screenshot from 2019-06-25 17-27-01


modules/dreamview/backend/simulation_world/simulation_world_service.cc
```diff
@@ -942,6 +942,12 @@ void SimulationWorldService::UpdatePlanningData(const PlanningData &data) {
   for (auto &path : data.path()) {
     DownsamplePath(path, planning_data->add_path());
   }
+
+  // Update pull over status
+  if (data.has_pull_over_status()) {
+    planning_data->clear_pull_over_status();
```
vlin17: please move clear_pull_over_status() out of the if statement in case the previous frame has data but current one doesn't.

modules/dreamview/frontend/src/renderer/trajectory.js
```diff
@@ -36,9 +37,42 @@ function normalizePlanningTrajectory(trajectory, coordinates) {
     return result;
 }
 
+function getPullOverStatus({ lengthFront, lengthBack, widthLeft, widthRight }) {
```
vlin17: Let's create a class PlanningStatus in a new file to draw pull over status.

modules/dreamview/frontend/src/renderer/trajectory.js
```diff
@@ -97,5 +131,39 @@ export default class PlanningTrajectory {
...
+                }
+            } else {
+                if (this.pullOverStatus) {
+                    this.pullOverStatus.traverse((child) => {
+                        if (child.geometry !== undefined) {
+                            child.geometry.dispose();
+                            child.material.dispose();
+                        }
+                        scene.remove(this.pullOverStatus);
+                    });
```
vlin17: reset this.pullOverStatus = null here?

PR #9008, opened 2019-06-26
can message data length could be less than 8

liangyd:
The canbus message data length could be an integer between 0 and 8. The current if-statement will filter out those can messages whose data length is not equal to 8.


jinghaomiao:
I am closing this ticket as the issue has been addressed.

PR #9015, opened 2019-06-27
map tools: data verification tool, second ten files

ooooooops:
map tools: data verification tool, second ten files


xiaoxq:
LGTM on coding.

PR #9016, opened 2019-06-27
map tools: data verification tool, third ten files

ooooooops:
map tools: data verification tool, third ten files


xiaoxq:
LGTM on coding.

PR #9020, opened 2019-06-27
Localization: Use corrected imu from SINS

zhouyao4321:
Corresponding issue: 
http://newicafe.baidu.com/issue/IDG-Apollo-3670/show?from=page

注意: 需要等定位的docker volume更新之后才能通过CI .


modules/localization/msf/local_integ/localization_integ_process.cc
```diff
@@ -241,6 +245,21 @@ void LocalizationIntegProcess::GetResult(IntegState *state, InsPva *sins_pva,
   return;
 }
 
+void LocalizationIntegProcess::GetRemoveBiasImu(ImuData *imu_data) {
```
xiaoxq: This name is a little confusing with the two verbs.

PR #9021, opened 2019-06-27
Planning: refine boundary formation by delta-theta between curb-segments

JasonXu12:


modules/planning/tasks/deciders/open_space_decider/open_space_roi_decider.h
```diff
@@ -71,6 +71,7 @@ class OpenSpaceRoiDecider : public Decider {
 
   void SetPullOverSpotEndPose(Frame *const frame);
 
+  // @brief Get road boundaires of both sides
```
xiaoxq: boundaries

PR #9088, opened 2019-07-03
adding efficient graph cut testing

weidezhang:


modules/perception/lidar/lib/segmentation/ncut/common/graph_felzenszwalb/BUILD
```diff
@@ -0,0 +1,22 @@
+load("//tools:cpplint.bzl", "cpplint")
+
+load("//tools:cpplint.bzl", "cpplint")
```
mickeyouyou: It may be repeated here.

PR #9108, opened 2019-07-09
record file cleanup

Fla3inH0tCheet0s:
Continued from https://github.com/ApolloAuto/apollo/pull/8395.


barangoru:
@Fla3inH0tCheet0s : can you please take action on this PR?

natashadsouza:
@Fla3inH0tCheet0s please review the build failure messages and make another pull request. Thank you.

cyber/record/file/record_file_writer.h
```diff
@@ -123,7 +123,10 @@ bool RecordFileWriter::WriteSection(const T& message) {
     AERROR << "Do not support this template typename.";
     return false;
   }
-  Section section = {type, message.ByteSize()};
+  Section section;
+  /// zero out whole struct even if padded
+  memset(&section, 0, sizeof(section));
+  section = {type, static_cast<int64_t>(message.ByteSize())};
```
baowanyu: Not need static cast to int64_t here.

cyber/record/file/record_file_writer.h
```diff
@@ -123,7 +123,10 @@ bool RecordFileWriter::WriteSection(const T& message) {
     AERROR << "Do not support this template typename.";
     return false;
   }
-  Section section = {type, message.ByteSize()};
+  Section section;
+  /// zero out whole struct even if padded
+  memset(&section, 0, sizeof(section));
```
baowanyu: Use the default initialization "s = {x, y}" directly is better, you don't need "memset".

cyber/record/file/record_file_writer.h
```diff
@@ -135,9 +138,10 @@ bool RecordFileWriter::WriteSection(const T& message) {
            << ", actual count: " << count;
     return false;
   }
-  ZeroCopyOutputStream* raw_output = new FileOutputStream(fd_);
-  message.SerializeToZeroCopyStream(raw_output);
-  delete raw_output;
+  {
+    FileOutputStream raw_output(fd_);
+    message.SerializeToZeroCopyStream(&raw_output);
+  }
```
baowanyu: It's a good way.

cyber/record/file/record_file_writer.h
Fla3inH0tCheet0s: According to https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.message_lite#MessageLite.ByteSize, `ByteSize()` returns `int`. `int` doesn't specify a size. On my system `int` is 4 bytes. I think we still want the cast.

cyber/record/file/record_file_writer.h
Fla3inH0tCheet0s: Later in the code
```
write(fd_, &section, sizeof(section));
```
writes the memory footprint of the object not just the object itself. There may be memory in the object between the members. Therefore, the program is reading from uninitialized memory. `memset` is an attempt at setting these bytes. `valgrind` reports an error, if this is not set. It's better to set this memory so that the record files are more regular.

cyber/record/file/record_file_writer.h
baowanyu: Converting low precision int to high precision int do not need explicit type casting. 
I think it's better to use explicit casting where it is necessary. 
The compiler will issue an error message unless an explicit cast is made.

cyber/record/file/record_file_writer.h
baowanyu: I think constructor construct "struct {int int}" with a block of contiguous memory, each field is located at a certain fixed offset from the start. 

But it may have its own reason for “valgrind” to report an error, it may have considered different compilers or otherwise.

I agree with this change.

cyber/record/file/record_file_writer.h
Fla3inH0tCheet0s: That works for me. I made the change.

PR #9115, opened 2019-07-09
Dreamview: display all detected traffic light status

kiboliu:
Also extract traffic light & stop sign codes to traffic_control and let dreamview&offlineview use the same methods to render them.


modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,270 @@
...
+};
+
+const EPSILON = 1e-9;
+const TRAFFIC_LIGHT = 'signal';
+const STOP_SIGN = 'stopSign';
+
+export default class TrafficControl {
+    constructor() {
+        this.hash = -1;
+        this.trafficControl = {
```
siyangy: It's a little overgeneralized to call everything TrafficControl (this includes the class name, class data member and function name). Why do we want to process & store trafficLight & stopSign together? There seems to be enough discrepancies that we need to have some confusing and long switch cases to handle them separately later.

modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,270 @@
...
+                    scene.add(mesh);
+
+                    const id = OFFLINE_PLAYBACK ? item.id : item.id.id;
+                    this.trafficControl[item.type][id] = mesh;
+                });
+            }
+        });
+    }
+
+    removeTrafficControl(currentItemIds, scene) {
```
siyangy: What's currentItemIds? Are they ids to remove or to keep?

modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,270 @@
...
+                this.trafficControl[type] = currentItems;
+            });
+    }
+
+    clearTrafficLightStatus() {
+        if (this.trafficControl[TRAFFIC_LIGHT]) {
+            Object.values(this.trafficControl[TRAFFIC_LIGHT]).forEach((mesh) => {
+                if (mesh) {
+                    // clear all subsignals' status
+                    [4, 5, 6].forEach((index) => {
```
siyangy: Can we get rid of the magic numbers?

modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,270 @@
...
+            trafficSignal.forEach(signal => signalId2Color[signal.id] = signal.currentSignal);
+
+            Object.keys(this.trafficControl[TRAFFIC_LIGHT])
+                .filter(id => id in signalId2Color)
+                .forEach((id) => {
+                    const mesh = this.trafficControl[TRAFFIC_LIGHT][id];
+                    if (mesh) {
+                        switch (signalId2Color[id]) {
+                            case 'GREEN':
+                                mesh.children[4].material.emissive.set(0x17f470);
```
siyangy: Same, getting rid of the magic numbers will help remove the switch case here.

modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,270 @@
...
+            });
+        }
+    }
+
+    updateTrafficLightStatus(trafficSignal) {
+        if (!Array.isArray(trafficSignal)) {
+            return;
+        }
+
+        this.clearTrafficLightStatus();
```
siyangy: So this is to go through all the traffic lights in the map?

modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,270 @@
...
+                    const id = OFFLINE_PLAYBACK ? item.id : item.id.id;
+                    this.trafficControl[item.type][id] = mesh;
+                });
+            }
+        });
+    }
+
+    removeTrafficControl(currentItemIds, scene) {
+        Object.keys(currentItemIds)
+            .filter(type => [TRAFFIC_LIGHT, STOP_SIGN].includes(type))
```
siyangy: This is pretty confusing to iterate through the currentItemIds which looks like an array but actually a map keyed by type to extract out ids with specific types. Seems better to take the needed ids out prior to calling the function.

modules/dreamview/frontend/src/renderer/traffic_control.js
kiboliu: Traffic light and stop sign have similar processing flow, now I separate their different configs and pass type as function parameters to get relative configs. Hope that looks much cleaner.

modules/dreamview/frontend/src/renderer/traffic_control.js
kiboliu: to keep

modules/dreamview/frontend/src/utils/models.js
```diff
@@ -51,3 +51,7 @@ export function loadObject(materialFile, objectFile, scale, callback) {
 export function loadTexture(textureFile, onLoadCallback, onErrorCallback) {
     textureLoader.load(textureFile, onLoadCallback, undefined, onErrorCallback);
 }
+
+export function loadMaterial(materialFile, onLoadCallback, onErrorCallback) {
+    mtlLoader.load(materialFile, onLoadCallback, undefined, onErrorCallback);
```
siyangy: Looks like we never pass in onErrorCallback? Is just `mtlLoader.load(materialFile, onLoadCallback);` okay?

modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,328 @@
...
+                .forEach((id) => {
+                    const mesh = this.mesh[TRAFFIC_LIGHT][id];
+                    if (mesh) {
+                        const subsignal = signalId2Color[id];
+                        const index = SUBSIGNAL_TO_INDEX[subsignal];
+                        if (index) {
+                            const lightMaterial = _.get(
+                                this.SUBSIGNAL_MATERIALS, `${subsignal}.LIGHT`);
+
+                            mesh.children[index].material = lightMaterial;
```
siyangy: Have a variable for mesh.children[index] for clarity.

modules/dreamview/frontend/src/renderer/tileground.js
```diff
@@ -82,28 +77,15 @@ export default class TileGround {
...
+            newItems.push(item);
 
-                    // a tile may have multiple signal or stopSign
-                    currentItems[`${key}_${item.x}_${item.y}`] = mesh;
-                    scene.add(mesh);
-                });
+            if (!this.currentItems[type]) {
+                this.currentItems[type] = [];
+            }
+            this.currentItems[type].push(`${key}_${item.id}`);
```
siyangy: Seems better to push an object with fields like key & id as opposed doing string concat and split.

modules/dreamview/frontend/src/renderer/tileground.js
```diff
@@ -82,28 +77,15 @@ export default class TileGround {
...
-                    mesh.rotation.x = Math.PI / 2;
-                    mesh.rotation.y = item.heading;
-                    mesh.position.set(itemPos.x, itemPos.y, itemPos.z);
-                    if (item.id) {
-                        // store signal id as name in mesh
-                        mesh.name = item.id;
-                    }
-                    mesh.matrixAutoUpdate = false;
-                    mesh.updateMatrix();
+            item.type = type;
```
siyangy: Why store a type here given we are passing a type to addTrafficControl? Overall not a fan of passing this `type` around, would rather have function like addTrafficSignal and addStopSign.

modules/dreamview/frontend/src/renderer/tileground.js
```diff
@@ -82,28 +77,15 @@ export default class TileGround {
...
+            newItems.push(item);
 
-                    // a tile may have multiple signal or stopSign
-                    currentItems[`${key}_${item.x}_${item.y}`] = mesh;
-                    scene.add(mesh);
-                });
+            if (!this.currentItems[type]) {
+                this.currentItems[type] = [];
```
siyangy: A function called 'getItems' shouldn't mutate internal structure like this.currentItems

modules/dreamview/frontend/src/renderer/index.js
```diff
@@ -493,6 +498,14 @@ class Renderer {
         const names = intersects.map(intersect => intersect.object.name);
         return names;
     }
+
+    addTrafficControl(type, elements) {
```
vlin17: TrafficControl is still part of the map and its add/remove is controlled by map. What if we move this TrafficControl object to map/tileground?

modules/dreamview/frontend/src/renderer/tileground.js
kiboliu: Separate traffic light and stop sign now

modules/dreamview/frontend/src/renderer/traffic_control.js
```diff
@@ -0,0 +1,361 @@
...
+        });
+    }
+
+    addTrafficLight(items, coordinates, scene) {
+        if (!items || items.length === 0) {
+            return;
+        }
+
+        items.forEach((item) => {
+            const config = {
```
siyangy: Move this to a class member

glebkiselev:
Is there any method to change traffic lights in the hand-mode? For example, the car in simcontrol is moving straight and meets the traffic light, then stopped. How to change the light to continue the movement?

PR #9117, opened 2019-07-10
Deleted vscode hidden files

jinghaomiao:


xiaoxq:
It's contributed by developers. But, yes, I vote to avoid hosting any IDE-specific configurations.

PR #9127, opened 2019-07-11
planning: initialized hailing scenario

sjiang2018:


YajiaZhang:
Maybe think about a better name for the scenario in the future.

modules/planning/scenarios/summon/summon_scenario_test.cc
```diff
@@ -0,0 +1,53 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+
+/**
+ * @file
+ **/
+
+#define protected public
```
Capri2014: Try avoid this practice

Capri2014:
@sjiang2018 @YajiaZhang @yifeijiang @jmtao Do we plan to make it a scenario or an/a few stage for in-lane scenario?

modules/planning/scenarios/summon/summon_scenario_test.cc
sjiang2018: Fixed. Thank you.

YajiaZhang:
> @sjiang2018 @YajiaZhang @yifeijiang @jmtao Do we plan to make it a scenario or an/a few stage for in-lane scenario?

Let's make it a scenario first. From the functionality aspect, this should be the similar to pull-over scenario.

PR #9134, opened 2019-07-12
Dreamview: update traffic signal

kiboliu:
* Update displaying planning traffic light. Now the rule is that if all traffic light overlapped are green, show green; else if there is one or more traffic light is not unknown, show that color; else show unknown. The rule is mainly for worldsim scenarios that sometimes required traffic lights are not configured. In real road test, perception module ideally ensure multiple traffic lights on the same place will return same colors.
* Update traffic light obj
* Fix the removal logic


modules/dreamview/backend/simulation_world/simulation_world_service.cc
```diff
@@ -955,13 +956,33 @@ void SimulationWorldService::UpdatePlanningData(const PlanningData &data) {
...
 
   // Update planning signal
-  // TODO(zhangqian): check with planning to decide a reasonable signal
   world_.clear_traffic_signal();
   if (data.has_signal_light() &&
       data.signal_light().signal_size() > 0) {
+    TrafficLight::Color current_signal = TrafficLight::UNKNOWN;
+    int green_light_count = 0;
+
+    for (auto &signal : data.signal_light().signal()) {
```
unacao: If planning is just looking at perception signals, should we implement this logic upon receiving perception traffic light?

kiboliu: Planning is looking at traffic light overlapped with its reference line, and then get these traffic lights' color from perception.

PR #9136, opened 2019-07-14
add bridge receiver test

xmyqsh:


modules/bridge/test/bridge_sender_test.cc
```diff
@@ -86,9 +86,12 @@ bool send(const std::string &remote_ip, uint16_t remote_port, uint32_t count) {
         ADEBUG << "sent msg failed ";
         break;
       }
-      ADEBUG << "sent " << nbytes << "bytes to server";
+      AINFO << "sent " << nbytes << " bytes to server with sequence num " << i;
```
jinghaomiao: use ADEBUG

xmyqsh: Done

PR #9145, opened 2019-07-16
cyber TimerComponent GetInterval

Fla3inH0tCheet0s:
It is convenient to get the interval directly from TimerComponent.


Fla3inH0tCheet0s:
@quning18 

> Any use case for this interfaces?

I'm thinking about using this interface for general timing analysis. May check for overruns.

PR #9160, opened 2019-07-17
Localization: replace old local_map with pyramid_map

tthhee:


deidaraho:
It is possible to chop this huge chunk of file changes into serval independent requests?

PR #9180, opened 2019-07-19
fix log file soft link

xmyqsh:
After:
`routing.INFO -> routing.in_dev_docker.`
`planning.INFO -> planning.in_dev_docker.`

Before:
`mainborad.INFO -> mainborad.in_dev_docker.`


jinghaomiao:
this is not necessary as we are trying to keep the log names succinct and straightforward

PR #9201, opened 2019-07-23
Localization: RTK localization refactor.

xiaoxq:


modules/localization/rtk/rtk_localization.cc
```diff
@@ -413,8 +402,8 @@ bool RTKLocalization::InterpolateIMU(const CorrectedImu &imu1,
                                      const double timestamp_sec,
                                      CorrectedImu *imu_msg) {
   DCHECK_NOTNULL(imu_msg);
-  if (!(imu1.has_header() && imu1.header().has_timestamp_sec() &&
-        imu2.has_header() && imu2.header().has_timestamp_sec())) {
+  if (!(imu1.header().has_timestamp_sec() &&
```
KaWaiTsoiBaidu: Don't need to check it `has_header()` first?

xiaoxq: No. Proto will give an empty instance if it doesn't exist. So it's safe to test `imu1.header().has_timestamp_sec()` only.

PR #9224, opened 2019-07-24
Perception: update time_diff

xmyqsh:


weidezhang:
sorry for the delay. i am double checking the logic. will approve once it's confirmed.

PR #9225, opened 2019-07-24
Perception: update location distance

xmyqsh:
I'm confused with this equation.
location_dist = static_cast<float>(sqrt(dx * dx * 0.5 + dy * dy * 2));


xmyqsh:
@techoe 
Got it!
Thanks for your comprehensive explain.

PR #9227, opened 2019-07-24
Perception: fix RemoveGroundPointsInForegroundCluster and set points_in_rois

xmyqsh:
fix RemoveGroundPointsInForegroundCluster
     Note: the non_ground_indices is the indices of the cloud,
               not the indices of roi_indices
     (This mismatch may be caused by API update)

set points_in_roi

### DIFF
--- a/modules/perception/lidar/lib/segmentation/cnnseg/spp_engine/spp_engine.cc
+++ b/modules/perception/lidar/lib/segmentation/cnnseg/spp_engine/spp_engine.cc
@@ -132,14 +132,14 @@ size_t SppEngine::ProcessForegroundSegmentation(
 size_t SppEngine::RemoveGroundPointsInForegroundCluster(
     const base::PointFCloudConstPtr full_point_cloud,
     const base::PointIndices& roi_indices,
-    const base::PointIndices& roi_non_ground_indices) {
+    const base::PointIndices& non_ground_indices) {
   mask_.Set(full_point_cloud->size(), 0);
-  mask_.AddIndices(roi_indices);
-  mask_.RemoveIndicesOfIndices(roi_indices, roi_non_ground_indices);
-  mask_.Flip();
-  // at this time, all ground points has mask value 0
+  mask_.AddIndices(roi_indices, non_ground_indices);
+  // at this time, all non roi points or ground points has mask value 0
   for (size_t i = 0; i < clusters_.size(); ++i) {
     clusters_[static_cast<int>(i)]->RemovePoints(mask_);
+    clusters_[static_cast<int>(i)]->points_in_roi =
+      clusters_[static_cast<int>(i)]->points.size();
   }
   clusters_.RemoveEmptyClusters();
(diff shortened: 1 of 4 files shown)


modules/perception/lidar/lib/segmentation/cnnseg/spp_engine/spp_engine.cc
```diff
@@ -132,14 +132,14 @@ size_t SppEngine::ProcessForegroundSegmentation(
 size_t SppEngine::RemoveGroundPointsInForegroundCluster(
     const base::PointFCloudConstPtr full_point_cloud,
     const base::PointIndices& roi_indices,
-    const base::PointIndices& roi_non_ground_indices) {
```
weidezhang: the input does reflect the non_ground_indices in roi cloud instead of original cloud. so the original logic was correct. thx.

xmyqsh: Please double check [point_indices_temp_](https://github.com/ApolloAuto/apollo/blob/master/modules/perception/lidar/lib/ground_detector/spatio_temporal_ground_detector/spatio_temporal_ground_detector.cc#L175) when [use_roi_](https://github.com/ApolloAuto/apollo/blob/master/modules/perception/lidar/lib/ground_detector/spatio_temporal_ground_detector/spatio_temporal_ground_detector.cc#L136-L138) and [not use_roi_](https://github.com/ApolloAuto/apollo/blob/master/modules/perception/lidar/lib/ground_detector/spatio_temporal_ground_detector/spatio_temporal_ground_detector.cc#L145-L146)
in spatio_temporal_ground_detector.cc which is the default ground detector.
```
  // copy point data, filtering lower points under ground
  if (use_roi_) {
    for (i = 0; i < num_points; ++i) {
      index = frame->roi_indices.indices[i];
      const auto& pt = frame->world_cloud->at(index);
      point_indices_temp_[valid_point_num++] = index;
      data_[data_id++] = static_cast<float>(pt.x - cloud_center_(0));
      data_[data_id++] = static_cast<float>(pt.y - cloud_center_(1));
      data_[data_id++] = static_cast<float>(pt.z - cloud_center_(2));
    }
  } else {
    for (i = 0; i < num_points; ++i) {
      const auto& pt = frame->world_cloud->at(i);
      point_indices_temp_[valid_point_num++] = static_cast<int>(i);
      data_[data_id++] = static_cast<float>(pt.x - cloud_center_(0));
      data_[data_id++] = static_cast<float>(pt.y - cloud_center_(1));
      data_[data_id++] = static_cast<float>(pt.z - cloud_center_(2));
    }
  }
```

```
non_ground_indices.indices.push_back(point_indices_temp_[i]);
```

weidezhang: you actually right ! i get what you mean. looks like it is a bug. I will approve it.

PR #9231, opened 2019-07-25
modified the socketcan driver , ignore when length of can frame data …

wfgsy101:
…not equal to 8

modified modules/drivers/canbus/can_client/socket/socket_can_client_raw.cc
skip the data length checking of canbus frame


jinghaomiao:
We have to be very strict about the length of the can messages. If the length is not expected, and error message should be captured right away.

PR #9248, opened 2019-07-29
Check GlobalData Instance after google logging

telnet2:
Since `GlobalData::Instance()` access google logging before initialization, following warning printed to STDERR

image

To prevent this, it is desirable to move the CHECK_NULL after glog initialization unless the warning is intentional.


telnet2:
Dup with https://github.com/ApolloAuto/apollo/pull/9258

PR #9258, opened 2019-07-29
Cyber: moved common globaldata instance check

Fla3inH0tCheet0s:
There's a warning whenever any cyber program is run. To reproduce
```
bazel build //cyber/examples:talker
bazel-bin/cyber/examples/talker
```
The warning is
```
WARNING: Logging before InitGoogleLogging() is written to STDERR
```

After this change, the warning will go away. All glog calls should occur after the `google::InitGoogleLogging(binary_name);`. I assume this check is for async logger. Must all singletons be checked for not null?

### DIFF
--- a/cyber/init.cc
+++ b/cyber/init.cc
@@ -53,7 +53,6 @@ void InitLogger(const char* binary_name) {
   } else {
     ::apollo::cyber::Binary::SetName(binary_name);
   }
-  CHECK_NOTNULL(common::GlobalData::Instance());
 
   // Init glog
   google::InitGoogleLogging(binary_name);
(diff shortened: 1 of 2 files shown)


jinghaomiao:
@Fla3inH0tCheet0s Please retag your pull requests with "Module: description". Thanks!

Fla3inH0tCheet0s:
If you assume all singletons are used after `google::InitGoogleLogging(binary_name);`, you can add add a check to the callback at https://github.com/ApolloAuto/apollo/blob/51651b9105e55c14e65cc2bd349b479e55fffa36/cyber/common/macros.h#L59.

PR #9265, opened 2019-07-30
Docker: Add entrypoint for container one-time setup.

xiaoxq:


xiaoxq:
Per request by Shuangcheng.

PR #9271, opened 2019-07-31
Cyber: intra supports raw message pub proto sub

Fla3inH0tCheet0s:
This functionality should allow for cyber records to be published and subscribed by unmodified modules in the same process. A proto message can subscribe and read messages from a raw message publisher. It's implemented as overloads so there shouldn't be runtime overhead. A separate class was used because partial template specialization works at the class level and not the function level.

It may be slightly better to have the publisher do the type conversion so that multiple subscribers don't have to parse the RawMessage. For my use cases there are few subscribers on the same channel name so the optimization would do little.

There's a controversial part where the type checks are moved to fatal. I think this change is correct, because that is a programming error that we can only catch at runtime. Also, sometimes the error level messages are hard to see, but the crashes are more obvious. Generally, the subscribers are setup at boot so most modules would crash at boot. On the other hand, this means that a publisher can crash a node by publishing message with a different type. In practice, I think trading off correctness over preventing crashes in unusual situations is worth it.

### DIFF
--- a/cyber/transport/dispatcher/dispatcher.h
+++ b/cyber/transport/dispatcher/dispatcher.h
@@ -97,7 +97,7 @@ void Dispatcher::AddListener(const RoleAttributes& self_attr,
     handler =
         std::dynamic_pointer_cast<ListenerHandler<MessageT>>(*handler_base);
     if (handler == nullptr) {
-      AERROR << "please ensure that readers with the same channel["
+      AFATAL << "please ensure that readers with the same channel["
              << self_attr.channel_name()
              << "] in the same process have the same message type";
       return;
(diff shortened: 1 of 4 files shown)


fengqikai1414:
@Fla3inH0tCheet0s Thank you very much for your submission. This feature is really useful in some scenarios, so we have been working on development and testing within our team. After the test is passed, we will sync the code to github, please stay tuned for our updates. Thank you again for your interest in Cyber RT.

natashadsouza:
@Fla3inH0tCheet0s thank you so much for your contribution. It seems like your code did not pass our Build test. It could either be your code style or certain testing scenarios that failed. Please verify the details and resubmit your request.

fengqikai1414:
@Fla3inH0tCheet0s Pls see #9374 .

PR #9294, opened 2019-08-02
Localization: update pyramid_map module for local_pyramid_map

tthhee:


modules/localization/msf/local_pyramid_map/pyramid_map/pyramid_map_matrix_handler_test.cc
```diff
@@ -66,7 +67,7 @@ TEST_F(PyramidMapMatrixHandlerTestSuite, LossyMapFullAltMatrixHandler) {
   expect_size += pm_matrix->GetRowsSafe() * pm_matrix->GetColsSafe() *
                  (sizeof(unsigned char) + sizeof(unsigned char) +
                   sizeof(uint16_t) + sizeof(uint16_t) + sizeof(uint16_t));
-  std::unique_ptr<unsigned char> buf(new unsigned char[expect_size]);
+  unsigned char* buf = new unsigned char[expect_size];
```
jinghaomiao: The memory is not released. You need to be really careful about requesting new memory space.

tthhee: fixed

PR #9309, opened 2019-08-06
Perception: load undistort camera model by lane detector camera name

xmyqsh:


yg13:
The code itself looks OK, I am trying to confirm your motivation here. Are you trying to consider the case that lane detection uses a different camera compared to what is used for obstacle detection?

PR #9312, opened 2019-08-06
Perception: update camera param and config file path

xmyqsh:
Solve the miss matching of the config file.


KaWaiTsoiBaidu:
The extrinsic here `apollo/modules/perception/data/params/front_6mm_extrinsics.yaml` includes `euler_angles_degree` that this file `apollo/modules/perception/production/data/perception/camera/params/front_6mm_extrinsics.yaml` does not have, which is from the CIPV pull commit.   @techoe could you confirm it is safe to delete `apollo/modules/perception/data/params/front_6mm_extrinsics.yaml`?

techoe:
I think we are still using /apollo/modules/perception/data/params and alternatives are not provided in this PR.

xmyqsh:
@techoe 
What about other changes? If it is available, I will roll back the folder `/apollo/modules/perception/data/params`.
Otherwise, I will close this PR.

PR #9315, opened 2019-08-06
Perception: fix motion matrix generation, and set name of default motion as motion_3d

xmyqsh:


modules/perception/camera/lib/motion/plane_motion.cc
```diff
@@ -56,13 +56,10 @@ void PlaneMotion::generate_motion_matrix(base::VehicleStatus *vehicledata) {
         sqrt(vehicledata->velocity_x * vehicledata->velocity_x +
              vehicledata->velocity_y * vehicledata->velocity_y));
     float displacement = time_d * velocity;
-    trans(0) = static_cast<float>(displacement * cos(theta));
-    trans(1) = static_cast<float>(displacement * sin(theta));
-    // trans(0) = time_d * vehicledata->velocity_x;
-    // trans(1) = time_d * vehicledata->velocity_y;
+    trans << displacement, 0;
```
KaWaiTsoiBaidu: seems keeping the original code is more generic.

modules/perception/camera/lib/motion/plane_motion.cc
```diff
@@ -56,13 +56,10 @@ void PlaneMotion::generate_motion_matrix(base::VehicleStatus *vehicledata) {
...
 
     motion_2d.block(0, 0, 2, 2) = rot2d.toRotationMatrix().transpose();
-    motion_2d.block(0, 2, 2, 1) = -rot2d.toRotationMatrix().transpose() * trans;
+    motion_2d.block(0, 2, 2, 1) = -trans;
```
KaWaiTsoiBaidu: same here, seems keeping the original code is more generic.

modules/perception/camera/lib/motion/plane_motion.cc
xmyqsh: Here, it is equivalent.

modules/perception/camera/lib/motion/plane_motion.cc
xmyqsh: But, they are not equivalent here. More importantly, it is the `motion_3d`, not the `motion_2d` will be used later. That's the main purpose of this PR. Please double-check around here.

modules/perception/camera/lib/motion/plane_motion.cc
xmyqsh: Just want to keep consistent with the code style of `motion_3d`.
Maybe, we could comment it, not delete.
Or, code the `motion_3d` part in a complex way like `motion_2d` of the original. But I have not figured out have to calculate it.
In a word, this is a simple and easy solution I can figure out.

modules/perception/camera/lib/motion/plane_motion.cc
xmyqsh: same as above.

modules/perception/camera/lib/motion/plane_motion.cc
techoe: The modified code is incorrect.

modules/perception/camera/lib/motion/plane_motion.cc
```diff
@@ -76,20 +73,16 @@ void PlaneMotion::generate_motion_matrix(base::VehicleStatus *vehicledata) {
...
-        static_cast<float>(sqrt(displacement * displacement /
-                                (tan(yaw_delta) * tan(yaw_delta) +
-                                 tan(pitch_delta) * tan(pitch_delta) + 1)));
-    trans(1) = static_cast<float>(tan(yaw_delta) * trans(0));
-    trans(2) = static_cast<float>(tan(pitch_delta) * trans(0));
+    trans << displacement, 0, 0;
 
     motion_3d.block(0, 0, 3, 3) = rot3d.transpose();
-    motion_3d.block(0, 3, 3, 1) = -rot3d.transpose() * trans;
+    motion_3d.block(0, 3, 3, 1) = -trans;
```
techoe: Totally wrong change. motion_3D = [-RT | T]

xmyqsh: @techoe 
You mean motion_3D = [R^-1|-(R^-1)T], right?
And,
how about we calculate the trans as 
```
trans(0) = time_d * vehicledata->velocity_x;
trans(1) = time_d * vehicledata->velocity_y;
trans(2) = time_d * vehicledata->velocity_z;
```
I should be more accurate than the original one.
If so, I will create a new PR.

modules/perception/camera/lib/motion/plane_motion.cc
xmyqsh: @techoe 
Absolutely, you are right! It is my fault:)
Please help clarify the doubt. 
The code snip 
```
    trans(0) = static_cast<float>(displacement * cos(theta));
    trans(1) = static_cast<float>(displacement * sin(theta));
```
and 
```
    // trans(0) = time_d * vehicledata->velocity_x;
    // trans(1) = time_d * vehicledata->velocity_y;
```

is equivalent, right?

If so, we code use the latter one to format the motion 3D, which should be more accurate than the original one
```
    trans(0) =
        static_cast<float>(sqrt(displacement * displacement /
                                (tan(yaw_delta) * tan(yaw_delta) +
                                 tan(pitch_delta) * tan(pitch_delta) + 1)));
    trans(1) = static_cast<float>(tan(yaw_delta) * trans(0));
    trans(2) = static_cast<float>(tan(pitch_delta) * trans(0));
```

modules/perception/camera/lib/motion/plane_motion.cc
```diff
@@ -76,20 +73,16 @@ void PlaneMotion::generate_motion_matrix(base::VehicleStatus *vehicledata) {
     Eigen::AngleAxisf pitch_angle(pitch_delta, Eigen::Vector3f::UnitY());
     Eigen::AngleAxisf yaw_angle(yaw_delta, Eigen::Vector3f::UnitZ());
 
-    Eigen::Quaternion<float> q = roll_angle * pitch_angle * yaw_angle;
+    // here the intrinsic sequeeze is z-y-x(converted from z-x-y in location previously)
+    Eigen::Quaternion<float> q = yaw_angle * pitch_angle * roll_angle;
```
xmyqsh: @techoe 
Please help check the calculation of q, the rotation matrix.

xmyqsh:
@techoe 
It is a problem of integral the pose from velocity and angular velocity.
I have found the standard method here https://arxiv.org/pdf/1711.02508.pdf
Will create a new PR latter.

xmyqsh:
Not that. There is no acceleration.

xmyqsh:
@techoe 
It is a problem of integral the pose from velocity and angular velocity, and there is no acceleration.
So, it is not possible to update the velocity with angular velocity and acceleration who is the standard motion model with IMU.

But, we could update velocity with angular velocity and position like what **radar** does [rotation_novatel << 0, -angular_speed(2), angular_speed(1), angular_speed(2),0, -angular_speed(0), -angular_speed(1), angular_speed(0), 0;](https://github.com/ApolloAuto/apollo/blob/master/modules/perception/radar/lib/detector/conti_ars_detector/conti_ars_detector.cc#L44-L45) and [Eigen::Vector3d angular_trans_speed = rotation_radar * local_loc.topLeftCorner(3, 1);](https://github.com/ApolloAuto/apollo/blob/master/modules/perception/radar/lib/detector/conti_ars_detector/conti_ars_detector.cc#L70-L71)

In a word, just related to the equation of v = w * r.

It should be a WMR(Wheeled Mobile Robot) problem. 
The integration process is located around (5) of http://www.roboticsproceedings.org/rss10/p20.pdf 

Do you think so?

zhanghming:
@xmyqsh I am confused about the conversion process about transforming the raw radar object **ralative** speed to **world** speed.

What the calculation [Eigen::Vector3d angular_trans_speed = rotation_radar * local_loc.topLeftCorner(3, 1);](https://github.com/ApolloAuto/apollo/blob/master/modules/perception/radar/lib/detector/conti_ars_detector/conti_ars_detector.cc#L70-L71) do and why we should do following calculation to get the radar objects' world velocity:
```cpp
Eigen::Vector3d world_vel =
        static_cast<Eigen::Matrix<double, 3, 1, 0, 3, 1>>(
            radar2world_rotate * (local_vel + angular_trans_speed));
    Eigen::Vector3d vel_temp =
        world_vel + options.car_linear_speed.cast<double>();
    radar_object->velocity = vel_temp.cast<float>();
```

PR #9320, opened 2019-08-06
Common: fixed bugs in MPC_OSQP; added tests to compare result of MPC_…

sjiang2018:
…qpOASES and MPC_OSQP

### DIFF
--- a/modules/common/math/mpc_osqp.cc
+++ b/modules/common/math/mpc_osqp.cc
@@ -23,17 +23,23 @@ MpcOsqp::MpcOsqp(const Eigen::MatrixXd &matrix_a,
                  const Eigen::MatrixXd &matrix_b,
                  const Eigen::MatrixXd &matrix_q,
                  const Eigen::MatrixXd &matrix_r,
-                 const Eigen::MatrixXd &matrix_lower,
-                 const Eigen::MatrixXd &matrix_upper,
-                 const Eigen::MatrixXd &matrix_initial_state,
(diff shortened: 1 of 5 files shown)


modules/common/math/mpc_osqp.cc
```diff
@@ -19,21 +19,28 @@
 namespace apollo {
 namespace common {
 namespace math {
+// #define ADEBUG AINFO
```
Capri2014: remove this line please

modules/control/controller/mpc_controller.cc
```diff
@@ -365,6 +366,14 @@ Status MPCController::ComputeControlCommand(
   Matrix upper_bound(controls_, 1);
   upper_bound << wheel_single_direction_max_degree_, max_acceleration_;
 
+  double max = std::numeric_limits<double>::max();
```
Capri2014: use const here

modules/control/controller/mpc_controller.cc
```diff
@@ -365,6 +366,14 @@ Status MPCController::ComputeControlCommand(
...
+  Matrix lower_state_bound(basic_state_size_, 1);
+  lower_state_bound << -1.0 * max, -1.0 * max, -1.0 * max, -1.0 * max,
+      -1.0 * max, -1.0 * max;
+
+  Matrix upper_state_bound(basic_state_size_, 1);
```
Capri2014: We may need to load the state confs from configs instead of just setting it to max, please add TODO here

PR #9335, opened 2019-08-07
Localization: update pyramid_map module in local_pyrami_map

tthhee:


MikeZhu92:
May I know where this module is being using by apollo?
It seems like msf and ndt are still using the older version of the base_map and it's derived classes.

PR #9342, opened 2019-08-07
Planning: path_bounds_decider handles NO_BORROW

Fla3inH0tCheet0s:
There is UB on reading unitialized `lane_boundary_type`, when `lane_borrow_info` is set to `NO_BORROW`.

### DIFF
--- a/modules/planning/tasks/deciders/path_bounds_decider/path_bounds_decider.cc
+++ b/modules/planning/tasks/deciders/path_bounds_decider/path_bounds_decider.cc
@@ -1325,6 +1325,10 @@ void PathBoundsDecider::PathBoundsDebugString(
 bool PathBoundsDecider::CheckLaneBoundaryType(
     const ReferenceLineInfo& reference_line_info, const double check_s,
     const LaneBorrowInfo& lane_borrow_info) {
+  if (lane_borrow_info == LaneBorrowInfo::NO_BORROW) {
+    return false;
+  }
+
   const ReferenceLine& reference_line = reference_line_info.reference_line();
   auto ref_point = reference_line.GetNearestReferencePoint(check_s);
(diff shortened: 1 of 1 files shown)


modules/planning/tasks/deciders/path_bounds_decider/path_bounds_decider.cc
```diff
@@ -1339,6 +1339,8 @@ bool PathBoundsDecider::CheckLaneBoundaryType(
     lane_boundary_type = hdmap::LeftBoundaryType(waypoint);
   } else if (lane_borrow_info == LaneBorrowInfo::RIGHT_BORROW) {
     lane_boundary_type = hdmap::RightBoundaryType(waypoint);
+  } else {
```
jmtao: if you want to check NO_BORROW, can u please move this to the beginning of this function, for better performance? thx

Fla3inH0tCheet0s: Is this function in a hot loop? It seems like these are mostly quick checks.

If we want to have minimal latency for the `lane_borrow_info==LaneBorrowInfo::NO_BORROW` case, another option is doing the check before the `CheckLaneBoundaryType` call in https://github.com/ApolloAuto/apollo/blob/15f375a8ff1464bdab9eb65cac091c72ab7a5613/modules/planning/tasks/deciders/path_bounds_decider/path_bounds_decider.cc#L785. It's probably good to have the `CheckLaneBoundaryType` method be valid for all values of `lane_borrow_info`.

Both?

jmtao: yes. you can add this check at both places.

PR #9350, opened 2019-08-08
 Planning: added ADC initial status to ParkAndGo planning context  

sjiang2018:


modules/planning/scenarios/park_and_go/stage_check.cc
```diff
@@ -55,6 +47,20 @@ Stage::StageStatus ParkAndGoStageCheck::FinishStage(const bool success) {
   return Stage::FINISHED;
 }
 
+void ADCInitStatus(Frame* frame) {
```
jmtao: If this is just a state shared among stages in this scenario, use ParkAndGoContext defined in the scenario class. 
PlanningContext is a static class for state shared among and beyond scenario/stage.

PR #9367, opened 2019-08-11
Map: fix stop_for_destination_ assigning

xmyqsh:
`GetWaypointIndex(routing_waypoint_index_.back().waypoint)`
must be always equal to 
`routing_waypoint_index_.back().index`

### DIFF
--- a/modules/map/pnc_map/pnc_map.cc
+++ b/modules/map/pnc_map/pnc_map.cc
@@ -191,10 +191,7 @@ bool PncMap::UpdateVehicleState(const VehicleState &vehicle_state) {
     return false;
   }
 
-  int last_index = GetWaypointIndex(routing_waypoint_index_.back().waypoint);
-  if (next_routing_waypoint_index_ == routing_waypoint_index_.size() - 1 ||
-      (!stop_for_destination_ &&
-       last_index == routing_waypoint_index_.back().index)) {
+  if (next_routing_waypoint_index_ == routing_waypoint_index_.size() - 1) {
     stop_for_destination_ = true;
   }
   return true;


modules/map/pnc_map/pnc_map.cc
```diff
@@ -191,10 +191,7 @@ bool PncMap::UpdateVehicleState(const VehicleState &vehicle_state) {
     return false;
   }
 
-  int last_index = GetWaypointIndex(routing_waypoint_index_.back().waypoint);
-  if (next_routing_waypoint_index_ == routing_waypoint_index_.size() - 1 ||
-      (!stop_for_destination_ &&
-       last_index == routing_waypoint_index_.back().index)) {
```
jmtao: GetWaypointIndex(routing_waypoint_index_.back().waypoint) may NOT be equal to routing_waypoint_index_.back().index.
for instance, when ADC passes the routing end. 

But it looks like (next_routing_waypoint_index_ == routing_waypoint_index_.size()) is always TRUE here when we need to stop for destination. 

Need another pair of eyes to confirm.

xmyqsh: @jmtao 
Please check twice.
`routing_waypoint_index_ ` is only assigned [here](https://github.com/ApolloAuto/apollo/blob/97651b1b4840ddc7c97c27c52c1cba0edb31ad7c/modules/map/pnc_map/pnc_map.cc#L249-L260):
```
  int i = 0;
  for (size_t j = 0; j < route_indices_.size(); ++j) {
    while (i < request_waypoints.size() &&
           RouteSegments::WithinLaneSegment(route_indices_[j].segment,
                                            request_waypoints.Get(i))) {
      routing_waypoint_index_.emplace_back(
          LaneWaypoint(route_indices_[j].segment.lane,
                       request_waypoints.Get(i).s()),
          j);
      ++i;
    }
  }
```
`GetWaypointIndex(WaypointIndex.waypoint)` is definitely equal to `WaypointIndex.index`

Second, right, `next_routing_waypoint_index_ == routing_waypoint_index_.size()` is just a rough check. And the checking result `stop_for_destination_ ` is only for debug info currently.

The worked `stop for destination` is located related in `traffic_rule/destination.cc`.

xmyqsh:
@jmtao 
Please check twice.
`routing_waypoint_index_ ` is only assigned [here](https://github.com/ApolloAuto/apollo/blob/97651b1b4840ddc7c97c27c52c1cba0edb31ad7c/modules/map/pnc_map/pnc_map.cc#L249-L260):
```
  int i = 0;
  for (size_t j = 0; j < route_indices_.size(); ++j) {
    while (i < request_waypoints.size() &&
           RouteSegments::WithinLaneSegment(route_indices_[j].segment,
                                            request_waypoints.Get(i))) {
      routing_waypoint_index_.emplace_back(
          LaneWaypoint(route_indices_[j].segment.lane,
                       request_waypoints.Get(i).s()),
          j);
      ++i;
    }
  }
```
`GetWaypointIndex(WaypointIndex.waypoint)` is definitely equal to `WaypointIndex.index`

Second, right, `next_routing_waypoint_index_ == routing_waypoint_index_.size()` is just a rough check. And the checking result `stop_for_destination_ ` is only for debug info currently.

The worked `stop for destination` is located related in `traffic_rule/destination.cc`.

PR #9375, opened 2019-08-13
prediction:fix coredump when get map by LaneById return nullptr

gengqx:
LaneById may return nullptr,and push into vector.


kechxu:
Hi @gengqx , Thanks a lot for your debugging.

PR #9380, opened 2019-08-13
Merge Ubuntu-18.04 support

ycool:


xmyqsh:
It is good for moving boost and OpenCV into third_party.
The compatibility should have been improved.

PR #9385, opened 2019-08-14
Planning: remove useless continue

xmyqsh:


modules/planning/common/frame.cc
```diff
@@ -193,7 +193,6 @@ bool Frame::CreateReferenceLineInfo(
   for (auto &ref_info : reference_line_info_) {
     if (!ref_info.Init(obstacles())) {
       AERROR << "Failed to init reference line";
-      continue;
     } else {
```
jmtao: actually it's better to keep "continue" here.  
otherwise it may be bug prone for potential future code changes.

PR #9387, opened 2019-08-14
Planning: remove duplicate assignment

xmyqsh:


modules/planning/common/obstacle.cc
```diff
@@ -100,7 +100,6 @@ Obstacle::Obstacle(const std::string& id,
                    const ObstaclePriority::Priority& obstacle_priority,
                    const bool is_static)
     : Obstacle(id, perception_obstacle, obstacle_priority, is_static) {
-  is_caution_level_obstacle_ = (obstacle_priority == ObstaclePriority::CAUTION);
```
jmtao: This is to initialize a member variable from different constructors. 
not duplicated.

xmyqsh: cannot got it.
L74 has done the same thing(`is_caution_level_obstacle_ = (obstacle_priority == ObstaclePriority::CAUTION);`) as L103 does from L102

jmtao: I see. sorry I didn't notice it calls the base constructor first. 
thx!

PR #9403, opened 2019-08-15
Planning: added vehicle speed check for ParkAndGo scenario

sjiang2018:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -530,17 +531,25 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectParkAndGoScenario(
...
+          .vehicle_param()
+          .max_abs_speed_when_stopped();
+
   hdmap::LaneInfoConstPtr lane;
-  if (HDMapUtil::BaseMap().GetNearestLaneWithHeading(
-          adc_point, 2.0, vehicle_state.heading(), M_PI / 3.0, &lane, &s, &l) !=
-      0) {
+  // if vehicle is static and (off-lane or not on city_driving lane)
+  if (std::fabs(adc_speed) < max_abs_speed_when_stopped &&
+      (HDMapUtil::BaseMap().GetNearestLaneWithHeading(
```
jmtao: add another if here, since lane_type check should be inside when speed check is true.

PR #9423, opened 2019-08-16
Planning: discrete point smoothing in sqp formulation

JasonZhou404:


modules/planning/math/discretized_points_smoothing/fem_pos_deviation_smoother.cc
```diff
@@ -77,7 +83,54 @@ bool FemPosDeviationSmoother::SolveWithOsqp(
...
 }
 
-bool FemPosDeviationSmoother::SolveWithIpopt(
+bool FemPosDeviationSmoother::SqpWithOsqp(
+    const std::vector<std::pair<double, double>>& raw_point2d,
+    const std::vector<double>& bounds, std::vector<double>* opt_x,
+    std::vector<double>* opt_y) {
+  CHECK_NOTNULL(opt_x);
+  CHECK_NOTNULL(opt_y);
+
```
jinghaomiao: Do not use CHECK, but return false if they are null pointers.

JasonZhou404: Changes are made. Thanks

PR #9429, opened 2019-08-18
Planning: enhance intersection scenarios selection logic

xmyqsh:
One thing is exchange the order of `FLAGS_enable_scenario` checking to prevent the unenabled scenario affects other overlap scenario.

The other is to add the missing `break` after `pnc_junction_overlap` scenario is selected. I'm not sure this `break` is necesary or not. Please double check here.
NOTE: `first_encountered_overlaps` has already been sorted with `start_s`.

### DIFF
--- a/modules/planning/scenarios/scenario_manager.cc
+++ b/modules/planning/scenarios/scenario_manager.cc
@@ -648,7 +648,6 @@ void ScenarioManager::ScenarioDispatch(const common::TrajectoryPoint& ego_point,
       } else if (FLAGS_enable_scenario_bare_intersection &&
                  overlap.first == ReferenceLineInfo::PNC_JUNCTION) {
         pnc_junction_overlap = const_cast<hdmap::PathOverlap*>(&overlap.second);
-        break;
       }
     }
 
--- a/modules/planning/scenarios/scenario_manager.cc
+++ b/modules/planning/scenarios/scenario_manager.cc
@@ -560,6 +560,7 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectParkAndGoScenario(
 }
 
 void ScenarioManager::Observe(const Frame& frame) {
+  CHECK(!frame.reference_line_info().empty());
   // init first_encountered_overlap_map_
   first_encountered_overlap_map_.clear();
   const auto& reference_line_info = frame.reference_line_info().front();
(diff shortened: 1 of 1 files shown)


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -636,30 +637,30 @@ void ScenarioManager::ScenarioDispatch(const common::TrajectoryPoint& ego_point,
         reference_line_info.FirstEncounteredOverlaps();
     // note: first_encountered_overlaps already sorted
     for (const auto& overlap : first_encountered_overlaps) {
-      if (overlap.first == ReferenceLineInfo::SIGNAL ||
-          overlap.first == ReferenceLineInfo::STOP_SIGN ||
+      if ((FLAGS_enable_scenario_traffic_light &&
```
jmtao: this is really unnecessary. 
(1) overlap map is gather "fact" regardless of the feature enable flag. 
(2) keep feature enable flag as close as possible to the feature implementation code.  Don't spread it around which can be bug prone.

modules/planning/scenarios/scenario_manager.cc
```diff
@@ -636,30 +637,30 @@ void ScenarioManager::ScenarioDispatch(const common::TrajectoryPoint& ego_point,
...
+           overlap.first == ReferenceLineInfo::SIGNAL) ||
+          (FLAGS_enable_scenario_stop_sign &&
```
jmtao: same as above

modules/planning/scenarios/scenario_manager.cc
```diff
@@ -636,30 +637,30 @@ void ScenarioManager::ScenarioDispatch(const common::TrajectoryPoint& ego_point,
...
     }
 
     if (traffic_sign_overlap) {
       switch (overlap_type) {
         case ReferenceLineInfo::STOP_SIGN:
-          if (FLAGS_enable_scenario_stop_sign) {
-            scenario_type =
-                SelectStopSignScenario(frame, *traffic_sign_overlap);
-          }
+          scenario_type =
```
jmtao: please keep the feature enable gflag check here. 
reason stated as above

modules/planning/scenarios/scenario_manager.cc
xmyqsh: @jmtao Please double check the here.
Imagine the following scenario: The closest overlap is the `STOP_SIGN`, and then the `YIELD_SIGN`. And the `FLAGS_enable_scenario_stop_sign` is false.
What do we except here? The `YIELD_SIGN` overlap should be executed. Right?

But the original logic will skip the `YIELD_SIGN`, because of the unenabled `FLAGS_enable_scenario_stop_sign` closest `STOP_SIGN` overlap.

xmyqsh:
@jmtao Please double check the here.
Imagine the following scenario: The closest overlap is the `STOP_SIGN`, and then the `YIELD_SIGN`. And the `FLAGS_enable_scenario_stop_sign` is false.
What do we except here? The `YIELD_SIGN` overlap should be executed. Right?

But the original logic will skip the `YIELD_SIGN`, because of the unenabled `FLAGS_enable_scenario_stop_sign` closest `STOP_SIGN` overlap.

modules/planning/scenarios/scenario_manager.cc
jmtao: Not really. 
We enter intersection scenario (stop_sign/ traffic_light/yield_sign) only when we are very close to them.  So, here, "first encountered overlap" still makes sense no matter  if the feature is enabled or not. 
for instance, "first encountered overlap" is a STOP_SIGN when stop-sign scenario is disabled.  That also imply YIELD_SIGN, which is not captured here, is a next one, and still not close enough. we don't even have to think about it as a scenario yet. 

remember, traffic-rule will draw stop-fence for all these traffic overlaps based on traffic rules though.

xmyqsh: Hum, imply...
Alright, STOP_SIGN will not stack with YIELD_SIGN. That's true.

jmtao: if "imply" is not the right word, I can use a stronger one. 
"STOP_SIGN will not stack with YIELD_SIGN. That's true." we don't want to stack them here at all. here we only want to get "first encountered" sign, and process it. 
In reality, there won't be two different signs sit next to each other (at the same s) along the reference line.

PR #9430, opened 2019-08-18
Planning: make Stage::tasks_ static

xmyqsh:
Because the task config may be different between different stages.
The type of tasks_ has been redefined.
tasks_[task_type]   --->  tasks_[stage_type][task_type]

Make tasks_ static to make it as a class member, not Stage object member, to guarantee that each (stage,  task) pair only initializes once, no need to initialize them every stage creation.

### DIFF
--- a/modules/planning/scenarios/stage.cc
+++ b/modules/planning/scenarios/stage.cc
@@ -42,6 +42,12 @@ constexpr double kSpeedOptimizationFallbackCost = 2e4;
 // constexpr double kStraightForwardLineCost = 10.0;
 }  // namespace
 
+std::unordered_map<ScenarioConfig::StageType,
+                   std::unordered_map<TaskConfig::TaskType,
+                                      std::unique_ptr<Task>,
+                                      std::hash<int>>,
+                   std::hash<int>> Stage::tasks_;
+
 Stage::Stage(const ScenarioConfig::StageConfig& config) : config_(config) {
   // set stage_type in PlanningContext
   PlanningContext::Instance()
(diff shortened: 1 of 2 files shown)


modules/planning/scenarios/stage.h
```diff
@@ -89,9 +89,14 @@ class Stage {
   virtual Stage::StageStatus FinishScenario();
 
  protected:
-  std::map<TaskConfig::TaskType, std::unique_ptr<Task>> tasks_;
+  static std::unordered_map<ScenarioConfig::StageType,
```
jmtao: I don't think you can change it to static here.
see reason above at line 75

modules/planning/scenarios/stage.cc
```diff
@@ -61,11 +68,11 @@ Stage::Stage(const ScenarioConfig::StageConfig& config) : config_(config) {
...
         << "Task: " << TaskConfig::TaskType_Name(task_type)
         << " used but not configured";
-    auto iter = tasks_.find(task_type);
-    if (iter == tasks_.end()) {
+    auto iter = tasks_[cur_stage_].find(task_type);
+    if (iter == tasks_[cur_stage_].end()) {
       auto ptr = TaskFactory::CreateTask(*config_map[task_type]);
       task_list_.push_back(ptr.get());
-      tasks_[task_type] = std::move(ptr);
+      tasks_[cur_stage_][task_type] = std::move(ptr);
```
jmtao: When you use a "static" task_map, a stage type will be put into this map and stay.  
line 73 - 75 will not be executed.  A second run of "the same stage_type" may not have the same config file, and therefore this stage won't be constructed properpy with the corresponding conf file. 

For instance, CREEP stage in scenario A may have different set of conf from CREEP stage in scenario B.

xmyqsh:
@jmtao 
Got it.
But the [L73-L75](https://github.com/ApolloAuto/apollo/blob/1981d7329e02a51d786dd928e7e46081c57beb23/modules/planning/scenarios/stage.cc#L73-L75) will almost always run in every loop in the original version.
I will find a way to solve this in another PR.

xmyqsh:
@jmtao 
Got it.
`Scenario::RegisterStages()` have done what I want to do here.
Alright, everything is well now.

PR #9438, opened 2019-08-19
 Planning: moved ParkAndGo front obstacle buffer to config

sjiang2018:


modules/planning/scenarios/util/util.cc
```diff
@@ -284,8 +284,8 @@ bool CheckADCReadyToCruise(Frame* frame,
 bool CheckADCSurroundObstacles(const common::math::Vec2d adc_position,
                                const double adc_heading, Frame* frame,
                                const ScenarioParkAndGoConfig& scenario_config) {
-  // TODO(SHU): move to config
-  const double kFrontBuffer = 2.0;  // safe cruise distance
+  // safe cruise distance
+  const double kFrontBuffer = scenario_config.front_obstacle_buffer();
```
JasonZhou404: Directly pass buffer value into the function. As it is in util/, it would be better to make it scenario independent.

PR #9441, opened 2019-08-19
 planning: enable traffic-light-left-turn scenario

jmtao:


modules/planning/scenarios/traffic_light/unprotected_left_turn/stage_creep.cc
```diff
@@ -59,12 +60,67 @@ Stage::StageStatus TrafficLightUnprotectedLeftTurnStageCreep::Process(
...
+
+  if (traffic_light_all_green) {
+    // creep
+    // TODO(all): find proper creeping area for left turn
+    const double wait_time =
+        Clock::NowInSeconds() - GetContext()->creep_start_time;
+    const double timeout_sec = scenario_config_.creep_timeout_sec();
+    auto* task =
+        dynamic_cast<CreepDecider*>(FindTask(TaskConfig::CREEP_DECIDER));
+    if (task &&
```
YajiaZhang: "if (task != nullptr ...)" is the preferred style in c++11. "if (task)" does implicit type casting and might not produce the same result if the system changes.

PR #9452, opened 2019-08-20
Update header.proto

anchans:
Remove ROS reference


xiaoxq:
Nice catch! Could you sign the CLA so we can merge your contribution?

PR #9456, opened 2019-08-21
Planning: add move constructor for FrenetFramePath and DiscretizedPath and use it wherever it is possible

xmyqsh:
Motivation of this PR:
There is some piece of code intended to use [FrenetFramePath move constructor](https://github.com/ApolloAuto/apollo/blob/master/modules/planning/common/path/path_data.cc#L262) or [DiscretizedPath move constructor](https://github.com/ApolloAuto/apollo/blob/master/modules/planning/scenarios/lane_follow/lane_follow_stage.cc#L336), while there is no move constructor for FrenetFramePath and DiscretizedPath. Pieces of code like `SetFrenetPath(FrenetFramePath(std::move(frenet_frame_points)));` calls extra copy constructor additionally which is not intended for.

### DIFF
--- a/modules/planning/common/path/discretized_path.cc
+++ b/modules/planning/common/path/discretized_path.cc
@@ -29,8 +29,8 @@ namespace planning {
 
 using apollo::common::PathPoint;
 
-DiscretizedPath::DiscretizedPath(const std::vector<PathPoint> &path_points)
-    : std::vector<PathPoint>(path_points) {}
+DiscretizedPath::DiscretizedPath(std::vector<PathPoint> path_points)
+    : std::vector<PathPoint>(std::move(path_points)) {}
 
 double DiscretizedPath::Length() const {
   if (empty()) {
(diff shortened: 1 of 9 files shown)


modules/planning/common/path/discretized_path.cc
```diff
@@ -29,9 +29,12 @@ namespace planning {
 
 using apollo::common::PathPoint;
 
-DiscretizedPath::DiscretizedPath(const std::vector<PathPoint> &path_points)
+DiscretizedPath::DiscretizedPath(const std::vector<PathPoint>& path_points)
```
YajiaZhang: You can combine these two constructors by using a single pass by value constructor like this:
DiscretizedPath::DiscretizedPath(std::vector<PathPoint> path_points) : std::vector<PathPoint>(std::move(path_points)) {}

xmyqsh: @YajiaZhang 
I just follow the format of apollo::map::Path way.
Right. The pass by value way is more concise with little extra one move construction for pass-by-right-value or move assignment for pass-by-left-value.
I will update it to the pass by value way.

xmyqsh: @YajiaZhang 
Done.

PR #9458, opened 2019-08-21
Planning: retire and adjust remove empty path_data, as well as some other refines

xmyqsh:
1. retire and adjust remove empty path_data as @panjiacheng required.
2. some other refine

### DIFF
--- a/modules/planning/tasks/deciders/path_assessment_decider/path_assessment_decider.cc
+++ b/modules/planning/tasks/deciders/path_assessment_decider/path_assessment_decider.cc
@@ -84,9 +84,18 @@ Status PathAssessmentDecider::Process(
          << " msec.";
 
   // 2. Analyze and add important info for speed decider to use
+  size_t cnt = 0;
   const Obstacle* blocking_obstacle_on_selflane = nullptr;
-  for (auto& curr_path_data : valid_path_data) {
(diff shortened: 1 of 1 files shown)


modules/planning/tasks/deciders/path_assessment_decider/path_assessment_decider.cc
```diff
@@ -84,8 +84,10 @@ Status PathAssessmentDecider::Process(
          << " msec.";
 
   // 2. Analyze and add important info for speed decider to use
+  size_t cnt = 0;
   const Obstacle* blocking_obstacle_on_selflane = nullptr;
-  for (auto& curr_path_data : valid_path_data) {
+  for (size_t i = 0; i != valid_path_data.size(); ++i) {
+    auto& curr_path_data = valid_path_data[i];
     if (curr_path_data.path_label().find("fallback") != std::string::npos) {
```
yifeijiang: the fallback path data may not be skipped for empty check added below.

xmyqsh: @yifeijiang  Got it. Will resolve it.

xmyqsh: Done.

PR #9491, opened 2019-08-23
 Tools: smart recorder add events logging and one scenario

michael4338:


modules/data/tools/smart_recorder/regular_interval_trigger.cc
```diff
@@ -0,0 +1,44 @@
...
+
+void RegularIntervalTrigger::Pull(const RecordMessage& msg) {
+  if (!trigger_obj_->enabled()) {
+    return;
+  }
+  if (current_recording_time_ == 0) {
+    current_recording_time_ = msg.time;
+    return;
+  }
+  if (msg.time - current_recording_time_ > recording_interval_ * 1000000000UL) {
```
xiaoxq: Avoid magic number. Use a const variable like `S_TO_NS_FACTOR`

PR #9518, opened 2019-08-26
tools: added plot acc jerk.

yifeijiang:


modules/tools/plot_planning/plot_acc_jerk.py
```diff
@@ -0,0 +1,71 @@
...
+###############################################################################
+
+import math
+from record_reader import RecordItemReader
+from imu_speed_jerk import ImuSpeedJerk
+from imu_speed_acc import ImuSpeedAcc
+
+
+if __name__ == "__main__":
+    import sys
```
KaWaiTsoiBaidu: How about putting all `import` on the top of the file?

PR #9575, opened 2019-09-02
Dreamview: remove rwlock on lidar_height_

xmyqsh:


xmyqsh:
Cannot remove it.
We may change vehicle and its lidar height during simulation.

PR #9578, opened 2019-09-02
Dreamview: replace boost rwlock by std rwlock

xmyqsh:


unacao:
[23:55:31]./modules/dreamview/backend/data_collection_monitor/data_collection_monitor.h:23:24: fatal error: shared_mutex: No such file or directory
[23:55:31] #include <shared_mutex>

We use C++11 in our build environment, and rwlock is not supported in this version.

PR #9592, opened 2019-09-04
Dreamview: camera view backend

kiboliu:
To provide projecting HDmap to camera view, the frontend adds a 'Camera View' under the point of view, and by default this option is not displayed to users, thus users won't feel any difference; the backend adds a websocket for sending camera image & related transformation matrix (transformation matrix is for calculating camera real-world position & rotation, based on static transforms and localization) to frontend.

### DIFF
--- a/modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
+++ b/modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
@@ -17,7 +17,6 @@
 #include "modules/dreamview/backend/perception_camera_updater/perception_camera_updater.h"
 
 #include <float.h>
-#include <map>
 #include <string>
 #include "cyber/common/file.h"
 #include "modules/common/adapters/adapter_gflags.h"
@@ -30,8 +29,7 @@ namespace dreamview {
 using apollo::drivers::CompressedImage;
 using apollo::localization::LocalizationEstimate;
 using apollo::localization::Pose;
-using apollo::transform::TransformStampeds;
-using apollo::transform::Transform;
+using apollo::transform::TransformStamped;
 using apollo::common::Quaternion;
 
 namespace {
@@ -45,18 +43,18 @@ namespace {
 
   template <typename Point>
   void ConstructTransformationMatrix(const Quaternion &quaternion,
-      const Point &translation, Eigen::Matrix4d &matrix) {
-    matrix.setConstant(0);
+      const Point &translation, Eigen::Matrix4d *matrix) {
+    matrix->setConstant(0);
     Eigen::Quaterniond q;
     q.x() = quaternion.qx();
     q.y() = quaternion.qy();
     q.z() = quaternion.qz();
     q.w() = quaternion.qw();
-    matrix.block<3, 3>(0, 0) = q.normalized().toRotationMatrix();
-    matrix(0, 3) = translation.x();
-    matrix(1, 3) = translation.y();
-    matrix(2, 3) = translation.z();
-    matrix(3, 3) = 1;
+    matrix->block<3, 3>(0, 0) = q.normalized().toRotationMatrix();
+    (*matrix)(0, 3) = translation.x();
+    (*matrix)(1, 3) = translation.y();
+    (*matrix)(2, 3) = translation.z();
+    (*matrix)(3, 3) = 1;
   }
 }  // namespace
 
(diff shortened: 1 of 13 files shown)


modules/dreamview/proto/camera_update.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto2";
+
+package apollo.dreamview;
+
+message CameraUpdate {
+    repeated double localization = 1;
```
unacao: add some comments about the order of doubles and its meaning?

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
```diff
@@ -0,0 +1,221 @@
...
+    const std::shared_ptr<LocalizationEstimate> &localization) {
+  if (!enabled_) {
+    return;
+  }
+
+  std::lock_guard<std::mutex> lock(localization_mutex_);
+  localization_queue_.push_back(localization);
+}
+
+void PerceptionCameraUpdater::OnStaticTransform(
```
unacao: is there any code source for handling the static transform? if yes, maybe we can state in the comment for reference

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
```diff
@@ -0,0 +1,221 @@
...
+}
+
+void PerceptionCameraUpdater::OnImage(
+    const std::shared_ptr<CompressedImage> &compressed_image) {
+  if (!enabled_
+      || compressed_image->format() == "h265" /* skip video format */) {
+    return;
+  }
+
+  std::lock_guard<std::mutex> lock(image_mutex_);
```
unacao: is lock still required after line 117?

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
```diff
@@ -0,0 +1,221 @@
...
+  auto iter = localization_queue_.rbegin();
+  Pose image_pos;
+  for (; iter != localization_queue_.rend(); ++iter) {
+    const double tmp_diff = (*iter)->measurement_time()
+        - current_image_timestamp_;
+    if (tmp_diff > 0) {
+      timestamp_diff = tmp_diff;
+      image_pos = (*iter)->pose();
+    } else {
+      if (std::abs(tmp_diff) < timestamp_diff) {
```
unacao: std::fabs

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
```diff
@@ -0,0 +1,221 @@
...
+      break;
+    }
+  }
+  // clean useless localization
+  auto iter_begin = localization_queue_.begin();
+  while ((*iter_begin) != (*iter)) {
+    iter_begin = localization_queue_.erase(iter_begin);
+  }
+
+  Eigen::Matrix4d localization_matrix;
```
unacao: Extract this part as a utility function since static transform has similar codes

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
kiboliu: Lock is required to save image timestamp and image itself, I moved the lock behind to exclude the image compression process.

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
```diff
@@ -0,0 +1,221 @@
...
+    const std::shared_ptr<LocalizationEstimate> &localization) {
+  if (!enabled_) {
+    return;
+  }
+
+  std::lock_guard<std::mutex> lock(localization_mutex_);
+  localization_queue_.push_back(localization);
+}
+
+void PerceptionCameraUpdater::OnStaticTransform(
```
vlin17: Consider using buffer.cc to get static_transform info

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
```diff
@@ -82,33 +80,71 @@ void PerceptionCameraUpdater::Stop() {
...
-  }
 
   Eigen::Matrix4d localization_matrix;
   ConstructTransformationMatrix(image_pos.orientation(), image_pos.position(),
-      localization_matrix);
+      &localization_matrix);
   ConvertMatrixToArray(localization_matrix, localization);
 }
 
+bool PerceptionCameraUpdater::QueryTF(const std::string &frame_id,
```
vlin17: QueryStaticTF?

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.cc
```diff
@@ -212,9 +208,12 @@ void PerceptionCameraUpdater::GetUpdate(std::string *camera_update) {
     GetImageLocalization(&localization);
     *camera_update_.mutable_localization() = {localization.begin(),
                                               localization.end()};
-
-    camera_update_.SerializeToString(camera_update);
   }
+  std::vector<double> tf_static;
+  GetStaticTF(&tf_static);
```
vlin17: Do we need to get new static TF for every update?

modules/dreamview/backend/perception_camera_updater/perception_camera_updater.h
```diff
@@ -70,6 +68,12 @@ class PerceptionCameraUpdater {
    */
   void GetImageLocalization(std::vector<double> *localization);
 
+  apollo::transform::Buffer *tf_buffer_ =
+      apollo::transform::Buffer::Instance();
+  bool QueryTF(const std::string &frame_id, const std::string &child_frame_id,
+      Eigen::Matrix4d *matrix);
+  void GetStaticTF(std::vector<double> *tf_static);
```
vlin17: Might want to be more specific about the name

modules/transform/buffer.cc
```diff
@@ -111,6 +111,19 @@ void Buffer::SubscriptionCallbackImpl(
   }
 }
 
+bool Buffer::GetLatestStaticTF(const std::string &frame_id,
+    const std::string &child_frame_id, TransformStamped *tf) {
+  for (auto &reverse_iter = static_msgs_.rbegin();
```
vlin17: compilable?

kiboliu: Ah it's not, I'll just keep the auto

PR #9595, opened 2019-09-05
Planning: added unit tests for path_reuse_decider

sjiang2018:


modules/planning/tasks/deciders/path_reuse_decider/path_reuse_decider.cc
```diff
@@ -52,6 +52,7 @@ Status PathReuseDecider::Process(Frame* const frame,
 
 bool PathReuseDecider::CheckPathReusable(Frame* const frame) {
   if (history_->GetLastFrame() == nullptr) return false;
+  constexpr double kEpsilon = 1e-6;
```
KaWaiTsoiBaidu: Can consider to define this `constexpr` in the header file? But should be also fine defining it here.

PR #9635, opened 2019-09-10
Canbus: enable the CheckResponse logic for the lexus controller

kevin-y-wang:


modules/canbus/vehicle/lexus/lexus_controller.cc
```diff
@@ -633,8 +633,52 @@ void LexusController::SecurityDogThreadFunc() {
...
+      --retry_num;
+      std::this_thread::sleep_for(
+          std::chrono::duration<double, std::milli>(20));
+    }
+  } while (need_wait && retry_num);
+
+  AINFO << "check_response fail: is_eps_online:" << is_eps_online
+        << ", is_vcu_online:" << is_vcu_online
+        << ", is_esp_online:" << is_esp_online;
   return false;
```
jinghaomiao: Use ADEBUG when appropriate.

kevin-y-wang: Thanks for the advice. Fixed per comment with respect to the actual needs.

PR #9675, opened 2019-09-13
Dreamview: camera view frontend

kiboliu:
* Add a new websocket client to request camera data every 100ms when the point of view is CameraView
* When CameraView is on:
  * set all map elements's zOffset to 0, since the zOffset will affect the matching between hdmap and image
  * set ground to invisible, so that the 3d scene will be transparent


modules/dreamview/frontend/src/components/Layouts/MainView.js
```diff
@@ -22,12 +22,27 @@ class SceneView extends React.Component {
                 options, trafficSignal, video, hmi } = this.props.store;
 
         const sceneHeightOffset = OFFLINE_PLAYBACK ? 40 /* height of playback control */ : 0;
+
+        let width = null;
+        let height = null;
+        if (options.cameraAngle === 'CameraView') {
+            // Set width/height same as camera image ratio to ensure scene align to image
+            // thus the height won't change no mattter whether there's menu under scene
```
unacao: 'mattter'->'matter'

modules/dreamview/frontend/src/store/websocket/websocket_camera.js
```diff
@@ -0,0 +1,59 @@
...
+        this.websocket.onmessage = event => {
+            this.worker.postMessage({
+                source: 'camera',
+                data: event.data,
+            });
+        };
+        this.worker.onmessage = event => {
+            const message = event.data;
+            switch (message.type) {
+                case "CameraData":
```
unacao: maybe add 'default' with a warning message?

PR #9677, opened 2019-09-13
record file cleanup

Fla3inH0tCheet0s:
Continued from https://github.com/ApolloAuto/apollo/pull/9108

@xiaoxq Thanks for the feedback. There was a bad merge and some of the modules disabled for compilation were reenabled accidentally. Still not sure about CI.

### DIFF
--- a/cyber/record/file/record_file_integration_test.cc
+++ b/cyber/record/file/record_file_integration_test.cc
@@ -1,0 +1,126 @@
+/******************************************************************************
+ * Copyright 2018 The Apollo Authors. All Rights Reserved.
+ *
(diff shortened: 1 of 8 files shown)


cyber/record/file/record_file_integration_test.cc
```diff
@@ -0,0 +1,126 @@
+/******************************************************************************
+ * Copyright 2018 The Apollo Authors. All Rights Reserved.
```
xiaoxq: 2019.

Fla3inH0tCheet0s: Sorry for the delay. The change is made.

PR #9690, opened 2019-09-16
Rename install_nvidia_docker2.sh to install_nvidia_docker.sh

moizsajid:
Otherwise, the command for installing nvidia-docker won't run properly which is mentioned in the README.md inside docker/scripts


moizsajid:
Pull request has already been accepted.

PR #9701, opened 2019-09-16
Planning: enable smarter lane-change logic and control with gflags

JasonXu12:


modules/planning/tasks/deciders/path_bounds_decider/path_bounds_decider.cc
```diff
@@ -148,10 +147,8 @@ Status PathBoundsDecider::Process(
       AERROR << msg;
       return Status(ErrorCode::PLANNING_ERROR, msg);
     }
-    if (!lanechange_path_bound.empty()) {
```
panjiacheng: Why removing this check?

JasonXu12: There is a lanechange_path_bound.empty() check on line 146 and returning with an Error status if empty. No need to do redundant check here. @panjiacheng

panjiacheng: Got it. thx.

PR #9710, opened 2019-09-18
Docker: add openh264 and autopep8 libraries

michael4338:


anchans:
this installs the shared object files to /usr/local/lib which is not in the LD_LIBRARY_PATH, please add it in apollo_base.sh

docker/build/installers/install_openh264.sh
```diff
@@ -0,0 +1,38 @@
...
+# Prepare
+PACKAGE="v2.0.0.tar.gz"
+OPEN_H264="openh264-2.0.0"
+wget https://github.com/cisco/openh264/archive/${PACKAGE}
+tar zxf ${PACKAGE}
+
+# Build and install.
+pushd ${OPEN_H264}
+  make
+  make install
```
Capri2014: default path not in `LD_LIBRARY_PATH`, suggest add ` export LD_LIBRARY_PATH=/usr/local/lib:$LD_LIBRARY_PATH` in `apollo_base.sh`, or try install this one under `/usr/local/openh264` or `/usr/local/apollo/openh264` as the other libs do.

hugomatic:
It looks like `make clean` can be invoked after make install (it won't remove the installed files).

docker/build/installers/install_openh264.sh
michael4338: thinking about adding it in next change, but okay

Capri2014:
will merge in and verify

PR #9733, opened 2019-09-19
Dreamview: adjust camera param

kiboliu:
* Display camera parameter adjust when choosing camera view, support edit and reset. For rotation, display static rotation (localization->camera) and dynamic rotation (world->localization) respectively as they may wanna adjust them separately.

* Add camera_data to store and save camera data to it instead of saving them in renderer

Screenshot from 2019-09-19 15-45-06


modules/dreamview/frontend/src/components/CameraParam/index.js
```diff
@@ -0,0 +1,200 @@
...
+        Roll: 'x'
+    }
+};
+
+@inject("store") @observer
+export default class CameraParam extends Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            hidden: true,
```
vlin17: is this used?

PR #9734, opened 2019-09-20
Tools: rtk_player code clean

sjiang2018:


modules/tools/record_play/rtk_player.py
```diff
@@ -182,8 +186,8 @@ def next_gear_switch_time(self, start, end):
         for i in range(start, end):
             # trajectory with gear switch
             # include gear_neutral at the beginning of a trajectory
-            if((i < end - 1) and (self.data['gear'][i] == 1 or self.data['gear'][i] == 2) and
-                    (self.data['gear'][i + 1] != self.data['gear'][i]) ):
+            if((i < end - 1) and (self.data['gear'][i] == 1 or self.data['gear'][i] == 2)
```
xiaoxq: Space after `if`. And to make the whole sentence more readable, I prefer something like:
```
if (i < end - 1 and
    self.data['gear'][i] in {1, 2} and
    self.data['gear'][i] != self.data['gear'][i + 1]):
```

PR #9744, opened 2019-09-21
Planning: adjust ST chart style and add speed profile

JasonXu12:


modules/planning/on_lane_planning.cc
```diff
@@ -609,9 +609,14 @@ void PopulateChartOptions(double x_min, double x_max, std::string x_label,
 }
 
 void AddSTGraph(const STGraphDebug& st_graph, Chart* chart) {
-  chart->set_title(st_graph.name());
-  PopulateChartOptions(-2.0, 10.0, "t (second)", 0.0, 80.0, "s (meter)", true,
-                       chart);
+  AERROR << "ST Name: " << st_graph.name();
```
yifeijiang: ADEBUG?

JasonXu12: Fixed. Thanks!

PR #9757, opened 2019-09-23
cyber:add raw message human-readable size

mickeyouyou:
image


cyber/tools/cyber_monitor/general_channel_message.cc
```diff
@@ -239,6 +239,21 @@ void GeneralChannelMessage::RenderDebugString(const Screen* s, int key,
         s->AddStr(0, lineNo++, "RawMessage Size: ");
         outStr.str("");
         outStr << channelMsg->message.size() << " Bytes";
+        if (channelMsg->message.size() >= (1024 * 1024 * 1024)) {
+          outStr << " ("
+                 << static_cast<double>(channelMsg->message.size()) /
+                        (1024 * 1024 * 1024)
+                 << " GB)";
```
xiaoxq: 1. Float is enough.
2. Is the output well formatted, like "%.2f"? It's more readable than a long decimal: "3.1415926 GB".
3. Use `static constexpr` to avoid recalculation and magic numbers.
4. Use bits in the constexpr, for example:
```
static constexpr size_t kGB = 2 << 30;
static constexpr size_t kMB = 2 << 20;
static constexpr size_t kKB = 2 << 10;
```

mickeyouyou: thanks good tips. I will update.

xiaoxq: > 1. context have set out precise  at line 232 `outStr << std::fixed << std::setprecision(FrameRatio_Precision) << frame_ratio();`

Got it :)

PR #9777, opened 2019-09-24
Planning: add categorized object status into History.

panjiacheng:


modules/planning/common/history.cc
```diff
@@ -123,6 +123,35 @@ const HistoryObjectDecision* HistoryFrame::GetObjectDecisionsById(
...
+  object_status_ = object_status;
+}
+
+
+////////////////////////////////////////////////
+// HistoryStatus
+
+void HistoryStatus::SetObjectStatus(
+    const std::string& id, const ObjectStatus& object_status) {
+  object_id_to_status_[id] = object_status;
```
JasonXu12: Will a DeleteObjectStatus function be necessary for the future to prevent the map from being too large?

panjiacheng: It will be used in History and as list of tables. If it's longer than a threshold then older ones will be removed.

PR #9781, opened 2019-09-25
Dreamview: batch fix

kiboliu:
As discussed:
* showCameraView changed to cameraAngle->hasCameraView, and rename isCameraView to showCameraView
* move routing editing check to Scene.js
* export min scene width and import it in renderer.js
* get image width and height from backend


modules/dreamview/frontend/src/styles/main.scss
```diff
@@ -1,5 +1,8 @@
 @import "media-queries";
 
+$min-scene-width: 600;
+
+:export { minSceneWidth: $min-scene-width; }
```
vlin17: Can upper case the variable (MIN_SCENE_WIDTH) to indicate it is a constant?

modules/dreamview/frontend/src/components/Scene/index.js
```diff
@@ -34,7 +34,8 @@ export default class Scene extends React.Component {
                     }}>
                     {options.showGeo && <Geolocation />}
                 </div>
-                { options.isCameraView && <img id="camera-image" /> }
+                { (options.showCameraView && !RENDERER.routingEditor.isInEditingMode()) &&
```
vlin17: the same can be achieve with "options.showRouteEditingBar"

modules/dreamview/frontend/src/components/Layouts/MainView.js
```diff
@@ -19,16 +19,16 @@ const Navigation = Loadable({
...
         const sceneHeightOffset = OFFLINE_PLAYBACK ? 40 /* height of playback control */ : 0;
 
         let width = null;
         let height = null;
-        if (options.isCameraView) {
+        if (options.showCameraView) {
             // Set width/height same as camera image ratio to ensure scene align to image
             // thus the height won't change no matter whether there's menu under scene
-            const ASPECT_RATIO = 1920 / 1080;
+            const ASPECT_RATIO = cameraData.imageWidth / cameraData.imageHeight;
```
vlin17: Let's camelCase for aspect ratio as it is not a constant.

modules/dreamview/proto/camera_update.proto
```diff
@@ -13,4 +13,8 @@ message CameraUpdate {
 
     // Camera image data
     optional bytes image = 3;
+
+    // Camera image width and height
+    optional uint32 img_height = 4;
+    optional uint32 img_width = 5;
```
vlin17: Why not simply set the aspect ratio here as img_height and img_width are used for calculating aspect ratio only.

PR #9797, opened 2019-09-26
Guardian: add the error_code message for future extension

kevin-y-wang:


Capri2014:
close per offline discussion.

PR #9804, opened 2019-09-26
cyber:fix code style

ctwfei:


cyber/base/for_each_test.cc
```diff
@@ -26,7 +26,7 @@ namespace base {
 
 TEST(ForEachTest, base) {
   std::vector<int> vec;
-  FOR_EACH(i, 0, 100) { vec.push_back(i); }
+  FOR_EACH(i, 0, 100) { vec.emplace_back(i); }
```
xiaoxq: When `in-place-construction` is not needed, prefer `push_back`.
https://abseil.io/tips/112

ctwfei: ok, thanks for your reminding.

ctwfei: > When `in-place-construction` is not needed, prefer `push_back`.
> https://abseil.io/tips/112

ok, thanks for your reminding.

Issue #9778, opened 2019-09-24
Looking for demo_multi_sensor_fusion localization data for apollo 3.5


harrycheng20:
Hi, I am following the [MSF localization instruction](https://github.com/ApolloAuto/apollo/blob/master/docs/howto/how_to_run_MSF_localization_module_on_your_local_computer.md) and from the instruction it requires localization data for apollo3.5:
-"In the downloaded data, you can find a folder named apollo3.5."
However I can only found the download link for localization data of [apollo2.5](http://data.apollo.auto/help?name=sensor%20data&data_key=multisensor&data_type=1&locale=en-us&lang=en), I am wondering if you can kindly point out where I can download the localization data for apollo3.5.

zhouyao4321:
Hi, The link to the corresponding dataset had been broken.
Temporary Solution: please contact Yao Zhou, zhouyao@baidu.com, to request the dataset.  Requests need contain the following:
    Email address and affiliation (business or school)
    Application purpose

harrycheng20:
Thank you!

natashadsouza:
Closing this issue as it appears to be resolved. Feel free to reopen if you have additional questions. Thanks!

ajay1606:
@natashadsouza Could anyone please provide me a localization dataset for apollo 3.5. Eagerly looking for that. 

From,
South korea

PR #9823, opened 2019-09-29
Localization: Update download link to MSF localization dataset

zhouyao4321:


zhouyao4321:
Related issue:
https://github.com/ApolloAuto/apollo/issues/9778

ajay1606:
@natashadsouza Am from south Korea and Is it possible to provide MSF localization dataset ?

PR #9830, opened 2019-09-30
planning: add interface to set vehicle signal

jmtao:


modules/planning/common/reference_line_info.cc
```diff
@@ -592,15 +599,33 @@ void ReferenceLineInfo::ExportTurnSignal(VehicleSignal* signal) const {
...
+void ReferenceLineInfo::SetEmergencyLight() {
+  vehicle_signal_.set_emergency_light(true);
+}
+
+void ReferenceLineInfo::ExportVehicleSignal(
+    common::VehicleSignal* vehicle_signal) const {
+  CHECK_NOTNULL(vehicle_signal);
+  vehicle_signal->Clear();
+
+  vehicle_signal->CopyFrom(vehicle_signal_);
```
xiaoxq: Simply `*vehicle_signal = vehicle_signal_;`, so you don't need to clear.

modules/planning/scenarios/park/pull_over_emergency/BUILD
```diff
@@ -6,9 +6,11 @@ cc_library(
     name = "pull_over_emergency_scenario",
     srcs = [
         "pull_over_emergency_scenario.cc",
+        "stage_approach.cc",
```
xiaoxq: Prefer adding its own BUILD target.

modules/planning/scenarios/park/pull_over_emergency/stage_approach_test.cc
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+
+#define protected public
+#define private public
```
xiaoxq: Seems not necessary?

modules/planning/common/reference_line_info.cc
jmtao: amended. thx!

modules/planning/scenarios/park/pull_over_emergency/BUILD
jmtao: there's circular dependency among the scenario and stages. 
will build them together as is.

modules/planning/scenarios/park/pull_over_emergency/stage_approach_test.cc
jmtao: keep it for future use. 
we may add more tests later.

PR #9845, opened 2019-10-04
stub for teleop BE

unacao:


modules/dreamview/backend/dreamview.cc
```diff
@@ -93,7 +93,10 @@ Status Dreamview::Init() {
   server_->addWebSocketHandler("/pointcloud", *point_cloud_ws_);
   server_->addWebSocketHandler("/camera", *camera_ws_);
   server_->addHandler("/image", *image_);
-
+#ifdef TELEOP
```
xiaoxq: I'm not sure if it's by design, but could it be handled by a FLAG instead of compile-time MACRO, which results in difference in binary?

unacao: teleop classes are not available to all developers, it depends on the mounting of private repo apollo-teleop

PR #9846, opened 2019-10-04
Planning: st-bounds-decider

panjiacheng:


modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
```diff
@@ -171,5 +212,27 @@ bool STObstaclesProcessor::IsPathPointAwayFromObstacle(
   return true;
 }
 
+bool STObstaclesProcessor::IsADCOverlappingWithObstacle(
+    const PathPoint& adc_path_point, const Box2d& obs_box,
+    const double l_buffer) const {
+  // Convert reference point from center of rear axis to center of ADC.
```
Capri2014: Would use the `ComputeCOMPosition` in `vehicle_state_provider` and convert to SL frame make the vehicle shape more precise?

modules/planning/proto/st_bounds_decider_config.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto2";
+
+package apollo.planning;
+
+message STBoundsDeciderConfig {
+  optional double total_time = 1 [default = 7.0];
```
xiaoxq: Seems like `float` is enough.

modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
```diff
@@ -42,109 +43,148 @@ using apollo::common::PathPoint;
...
-    // Sanity checks.
-    if (!obstacle.IsStatic()) {
-      AWARN << "Non-static obstacle[" << obstacle.Id()
-            << "] has NO prediction trajectory."
-            << obstacle.Perception().ShortDebugString();
-    }
-  } else {
-    // Processing a dynamic obstacle.
-  }
+  // std::vector<STPoint> lower_points;
```
xiaoxq: Add TODO to state when we should re-enable or retire this piece of code.

modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
```diff
@@ -42,109 +43,148 @@ using apollo::common::PathPoint;
...
+      vehicle_param_.front_edge_to_center(),
+      true, 0, static_cast<int>(adc_path_points.size()) - 2);
+  int pt_after_idx = GetSBoundingPathPointIndex(
+      adc_path_points, obstacle_instance,
+      vehicle_param_.back_edge_to_center(),
+      false, 0, static_cast<int>(adc_path_points.size()) - 2);
+  if (pt_before_idx == static_cast<int>(adc_path_points.size()) - 2)
+    return false;
+  if (pt_after_idx == 0)
+    return false;
```
xiaoxq: Always add brackets for `if`, `for`, `while`.

modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
panjiacheng: Done.

modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
panjiacheng: Done.

modules/planning/proto/st_bounds_decider_config.proto
panjiacheng: Yes, that's enough. But since all other places used double for time, I just wanted to be consistent. Anyway we are not that scarce in memory.

modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
panjiacheng: I checked that class. It seems that is providing the adc's current state based on localization and chassis. However, here I'm trying to calculate a possible collision for some futuristic position :)

PR #9849, opened 2019-10-04
Storytelling: Initial with README and proto.

xiaoxq:


modules/storytelling/proto/story.proto
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// Usage guide for action modules:
+// 1. Call `stories.has_XXX()` to check if a story you are interested is in
+//    charge.
+// 2. Access the story details if necessary, and take action accordingly.
+message Stories {
+  optional apollo.common.Header header = 1;
+
+  optional ProceedWithCaution proceed_with_caution = 2;
```
Capri2014: Is it better to use Enum here instead? The scenarios/states should be exclusive and only one shall be active at one time instance.

xiaoxq: Per offline discussion, keep it as of now.

PR #9852, opened 2019-10-04
planning: minor fix

jmtao:


modules/planning/scenarios/scenario_manager.cc
```diff
@@ -310,20 +310,18 @@ ScenarioConfig::ScenarioType ScenarioManager::SelectPadMsgScenario(
...
       }
       break;
-    // TODO(all): to be added
-    // case DrivingAction::STOP:
-    //  if (FLAGS_) {
-    //    return ScenarioConfig::STOP_EMERGENCY;
-    //  }
-    //  break;
+    case DrivingAction::STOP:
+      // TODO(all): to be added
```
xiaoxq: Specific assignee is preferred, like `TODO(jiaming, xx): `, other than `all` which means None :)

PR #9853, opened 2019-10-04
tools: updated imu speed and acc analysis.

yifeijiang:


modules/tools/plot_planning/imu_speed_acc.py
```diff
@@ -92,37 +92,54 @@ def plot_freq(x, y, ax, color):
...
-            x.extend(data_x)
-            y.extend(data_y)
+            data_x = lon_acc_processor.get_timestamp_list()
+            data_y = lon_acc_processor.get_acc_list()
 
-        if len(x) <= 0:
+            lon_time.extend(data_x)
+            lon_acc.extend(data_y)
+
+        if len(lat_time) <= 0:
```
xiaoxq: Use `==` as it cannot be negative.

PR #9861, opened 2019-10-07
planning: add a standby stage for emergency_pull_over scenario

jmtao:


modules/planning/scenarios/park/emergency_pull_over/BUILD
```diff
@@ -7,10 +7,12 @@ cc_library(
     srcs = [
         "emergency_pull_over_scenario.cc",
         "stage_approach.cc",
+        "stage_standby.cc",
```
xiaoxq: It keeps getting thicker :)

jmtao: yeah. just bear with it, mu-hahahaha

PR #9877, opened 2019-10-09
Control: design the simplest form of the MRAC controller

kevin-y-wang:


modules/control/common/mrac_controller.h
```diff
@@ -0,0 +1,154 @@
...
+   * @brief get saturation status for control system
+   * @return saturation status
+   */
+  int ControlSaturationStatus() const;
+
+ protected:
+  // reference model as a digital filter
+  common::DigitalFilter reference_model_;
+  // flag to indicate if the reference model is valid
+  bool ref_model_enabled_ = false;
```
sjiang2018: set as a gflag

modules/control/common/mrac_controller.cc
```diff
@@ -0,0 +1,178 @@
...
+      kd_ref_[2] = 4 - 2 * a_ref * Ts_ + b_ref * Ts_ * Ts_;
+    } else {
+      AWARN << "refrence model order beyond the designed range, "
+               "reference_model_order: "
+            << reference_model_order_;
+      ref_model_enabled_ = false;
+    }
+  }
+  ref_model_enabled_ = true;
+}  // namespace control
```
sjiang2018: remove  // namespace control

kevin-y-wang: fixed. Thanks

modules/control/common/mrac_controller.h
```diff
@@ -0,0 +1,154 @@
...
+  double gamma_theta_adp_;  // Nonlinear dynamics adaption gain
+
+  // Reference system coefficients in discrete-time domain
+  std::vector<double> kd_ref_{0.0, 0.0, 0.0};  // Denominator coefficients
+  std::vector<double> kn_ref_{0.0, 0.0, 0.0};  // Numerator coefficients
+
+  // Adaption system input variables in discrete-time domain
+  std::vector<double> r_des_{0.0, 0.0};       // Updated desired command vector
+  std::vector<double> x_act_{0.0, 0.0};       // Updated actuation states vector
+  std::vector<double> x_ref_{0.0, 0.0};       // Reference states vector
```
xiaoxq: Try to use more meaningful names instead of abbreviations. Many of them are not readable to me.

kevin-y-wang: Fixed per comment. Thanks.

modules/control/proto/BUILD
```diff
@@ -18,6 +18,7 @@ proto_library(
         "leadlag_conf.proto",
         "lon_controller_conf.proto",
         "mpc_controller_conf.proto",
+        "mrac_conf.proto",
```
xiaoxq: BTW, this BUILD target keeps piling up which is not good.

modules/control/common/mrac_controller.h
kevin-y-wang: fixed per comment. Thanks

modules/control/proto/BUILD
kevin-y-wang: Will re-organize the proto_library with a new pr

kevin-y-wang:
@Capri2014 Fixed the first two issues per your comment. For the third one (anti-windup design), I will fixed it with another new pr.

PR #9884, opened 2019-10-09
storytelling: Initial implementation of CloseToJunction teller.

xiaoxq:


modules/storytelling/story_tellers/close_to_junction_teller.cc
```diff
@@ -18,22 +18,96 @@
...
+      FLAGS_planning_trajectory_topic);
+  const auto trajectory = planning_reader->GetLatestObserved();
+  if (trajectory == nullptr || trajectory->trajectory_point().empty()) {
+    AERROR << "Planning trajectory not ready.";
+    return;
+  }
+
+  auto& logger = manager->LogBuffer();
+  const double distance = DistanceToPNCJunction(*trajectory);
+  const bool close_to_junction = distance >= 0;
```
Capri2014: The logic here is in junction, not close to junction right ?

xiaoxq: `distance == 0` could be taken as "in junction".

PR #9904, opened 2019-10-11
planning: add path length check in path_reuse_decider

sjiang2018:


modules/planning/tasks/deciders/path_reuse_decider/path_reuse_decider.h
```diff
@@ -67,6 +67,9 @@ class PathReuseDecider : public Decider {
   // check if path is collision free
   bool IsCollisionFree(ReferenceLineInfo* const reference_line_info);
 
+  // check path length
+  bool NotShortPath(DiscretizedPath* current_path);
```
xiaoxq: Use `const reference` as parameter if you don't change it.

PR #9914, opened 2019-10-14
planning: add place holder in path_bounds_decider for emergency_pull_over emergency

jmtao:


modules/planning/proto/planning_status.proto
```diff
@@ -52,6 +52,12 @@ message DestinationStatus {
   optional bool has_passed_destination = 1 [default = false];
 }
 
+message EmergencyPullOverStatus {
+  optional apollo.common.PointENU position = 1;
+  optional double theta = 2;
+  optional bool is_in_emergency_pull_over_scenario = 3 [default = false];
```
xiaoxq: FYI, it is `false` by default. But it's OK if you want to make it explicit.

modules/planning/tasks/deciders/path_bounds_decider/path_bounds_decider.cc
```diff
@@ -99,27 +99,34 @@ Status PathBoundsDecider::Process(
...
+                                     ->mutable_emergency_pull_over();
+  is_in_pull_over_scenario_ = pull_over_status->is_in_pull_over_scenario();
+  is_in_emergency_pull_over_scenario_ =
+      emergency_pull_over_status->is_in_emergency_pull_over_scenario();
+
+  if (is_in_pull_over_scenario_ || is_in_emergency_pull_over_scenario_) {
+    PathBound pull_over_path_bound;
     Status ret = GeneratePullOverPathBound(*frame, *reference_line_info,
-                                           &pullover_path_bound);
+                                           &pull_over_path_bound);
```
panjiacheng: "pullover" is actually one word. I would suggest stick with pullover_path_bound rather than pull_over_path_bound.

jmtao: 停车的”pull over“是两个字

PR #9915, opened 2019-10-14
Dreamview: teleop console FE

vlin17:
teleopConsole


modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -87,14 +98,9 @@ void TeleopService::RegisterMessageHandlers() {
       });
   // Request to get updated modem info for client display
   websocket_->RegisterMessageHandler(
-      "GetModemInfo",
+      "RequestTeleopStatus",
       [this](const Json &json, WebSocketHandler::Connection *conn) {
-        std::string to_send;
-        {
-          boost::shared_lock<boost::shared_mutex> reader_lock(mutex_);
-          MessageToJsonString(modem_info_, &to_send);
```
unacao: I think we need to preserve this part and save it as a field in teleop_status_ either when receiving modem info or SendStatus.

modules/dreamview/frontend/src/store/websocket/websocket_teleop.js
```diff
@@ -0,0 +1,63 @@
+import STORE from "store";
+import Worker from 'utils/webworker.js';
+
+export default class PointCloudWebSocketEndpoint {
```
unacao: need to rename this class to TeleopWebSocketEndpoint

PR #9922, opened 2019-10-15
Control: anti-windup design for the mrac controller

kevin-y-wang:


kevin-y-wang:
> where did you set the anti-wind up gain?

Missed the gain initialization in init(). Just fix it. Thanks

PR #9924, opened 2019-10-16
Planning: implementing sweep-line algorithms helper in st-obstacles-processor.

panjiacheng:


modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
```diff
@@ -123,6 +147,106 @@ STObstaclesProcessor::GetAllSTBoundaries() {
   return obs_id_to_st_boundary_;
 }
 
+bool STObstaclesProcessor::GetSBoundsFromDecisions(double t,
+    std::vector<std::pair<double, double>>* const available_s_bounds,
+    std::vector<std::vector<std::pair<std::string, ObjectDecisionType>>>*
+        const available_obs_decisions) {
+  // Sanity checks.
+  CHECK_NOTNULL(available_s_bounds);
+  CHECK_NOTNULL(available_obs_decisions);
```
ycool: let's avoid CHECK in runtime function. only exception to use CHECK is in init function. thanks.

modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
```diff
@@ -329,5 +453,59 @@ bool STObstaclesProcessor::IsADCOverlappingWithObstacle(
...
+        prev_open_s = obs_s_edge.first;
+    }
+  }
+
+  return s_gaps;
+}
+
+ObjectDecisionType STObstaclesProcessor::DetermineObstacleDecision(
+    const double obs_s_min, const double obs_s_max, const double s) const {
+  CHECK(s <= obs_s_min || s >= obs_s_max);
```
ycool: ditto

modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
panjiacheng: Thanks. I fixed this.

PR #9944, opened 2019-10-17
Scripts: Use system protoc instead of that from bazel workspace.

xiaoxq:


xiaoxq:
> if we are going to use system protoc, consider change this as well?
> 
> https://github.com/ApolloAuto/apollo/blob/945535b9444dfc052637c639142d679b77c9d4c1/WORKSPACE.in#L223

Nope. Here we only change the python compiling part, which is safe. For Bazel WORKSPACE we still stick to the live version.

This change fixes the failure of running `apollo.sh build_py` in a fresh container.

PR #9963, opened 2019-10-18
Teleop: added readers and writers for both daemons and each modem.

hugomatic:
UI switches turn on when streaming.

Note: the audio switch seems to be linked to the video streaming on the UI.


modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -39,11 +41,36 @@ TeleopService::TeleopService(WebSocketHandler *websocket)
...
+  operator_daemon_cmd_writer_ = node_->CreateWriter<DaemonServiceCmd>(
+      "/apollo/teleop/car/daemon_service/cmd");
+
+  car_daemon_rpt_reader_ = node_->CreateReader<DaemonServiceRpt>(
+      "/apollo/teleop/car/daemon_service/rpt",
+       [this](const std::shared_ptr<DaemonServiceRpt> &msg) {
+           UpdateCarDaemonRpt(msg);
+       });
+
+  car_daemon_rpt_reader_ = node_->CreateReader<DaemonServiceRpt>(
```
unacao: this should be 'operator_daemon_rpt_reader' according to the context

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -84,6 +111,7 @@ void TeleopService::RegisterMessageHandlers() {
         {
           boost::unique_lock<boost::shared_mutex> writer_lock(mutex_);
           teleop_status_["video"] = !teleop_status_["video"];
+          AINFO << "ToggleVideo" << std::endl;
```
unacao: std::endl is not needed for logging output

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
-    teleop_status_["modems"][modem_info->provider()] = modem_info->technology();
-  }
-}
 
 void TeleopService::SendStatus(WebSocketHandler::Connection *conn) {
   std::string to_send;
   {
     boost::shared_lock<boost::shared_mutex> reader_lock(mutex_);
     to_send = teleop_status_.dump();
+    AINFO << "status: " << to_send << std::endl;
```
unacao: std::endl is not needed for logging output

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
   }
   websocket_->SendData(conn, to_send);
 }
 
+
+void TeleopService::UpdateModem(unsigned int index,
+    const std::shared_ptr<ModemInfo> &modem_info) {
+
+    // AINFO << "UpdateModem: " << index << std::endl;
```
unacao: remove commented code

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
+    boost::unique_lock<boost::shared_mutex> writer_lock(mutex_);
```
unacao: this is duplicated as line 155

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
+        if (index == 1) {
+            modemId = "2";
+        }
+        if (index == 2) {
+            modemId = "3";
+        }
+
+        // teleop_status_["modems"][modem_info->provider()] =
+        //  modem_info->technology();
+        teleop_status_["modems"][modemId] = modem_info->technology();
```
unacao: only add writer_lock before updating teleop_status_ here

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
+    }
+}
+
+void TeleopService::UpdateCarDaemonRpt(
+    const std::shared_ptr<DaemonServiceRpt> &daemon_rpt) {
+  {
+      boost::unique_lock<boost::shared_mutex> writer_lock(mutex_);
```
unacao: move the writer_lock to line 190

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
+      // AINFO << "CarDaemonRpt" << std::endl;
+      bool aVideoEncoderIsRunning = false;
+      bool voipIsRunning = false;
+      for (int i = 0; i < daemon_rpt->services_size(); i++) {
+          std::string service =  daemon_rpt->services(i);
+          AINFO <<  "  *" << service << std::endl;
+          if (service.find("encoder") >= 0) {
```
unacao: if this is checked before 'voip_encoder', it will set 'aVideoEncoderIsRunning' even if it finds 'voip_encoder'.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
+      }
+      teleop_status_["video"] = aVideoEncoderIsRunning;
+      teleop_status_["audio"] = voipIsRunning;
+  }
+}
+
+void TeleopService::UpdateOperatorDaemonRpt(
+    const std::shared_ptr<DaemonServiceRpt> &daemon_rpt) {
+  {
+      boost::unique_lock<boost::shared_mutex> writer_lock(mutex_);
```
unacao: move writer_lock to line 208

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
+      // AINFO << "OperatorDaemonRpt" << std::endl;
+      bool voipIsRunning = false;
+      for (int i = 0; i < daemon_rpt->services_size(); i++) {
+          std::string service =  daemon_rpt->services(i);
+          AINFO <<  "  *" << service << std::endl;
+          if (service.find("voip_encoder") >= 0) {
+              voipIsRunning = true;
+          }
+      }
+      teleop_status_["mic"] = voipIsRunning;
```
unacao: if this means the report of operator side daemon outcome when remotely requested by car, then voipIsRunning should set both mic and audio?

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,84 @@ void TeleopService::RegisterMessageHandlers() {
...
+          AINFO <<  "  *" << service << std::endl;
+          if (service.find("encoder") >= 0) {
+              aVideoEncoderIsRunning = true;
+          }
+          if (service.find("voip_encoder") >= 0) {
+              voipIsRunning = true;
+          }
+      }
+      teleop_status_["video"] = aVideoEncoderIsRunning;
+      teleop_status_["audio"] = voipIsRunning;
```
unacao: voipIsRunning should also set mic?

modules/dreamview/backend/teleop/teleop.h
```diff
@@ -51,11 +53,36 @@ class TeleopService {
   WebSocketHandler *websocket_;
 
 #ifdef TELEOP
+
+  // modem info readers and callback
+  std::shared_ptr<cyber::Reader<modules::car1::network::ModemInfo>>
+      modem0_info_reader_;
```
unacao: i thought dreamview backend will listen to a combined modem signal summary instead of individual ones? Or does the modem monitor node computes the ranking together but still set the ranking info in separate modem signals?

modules/dreamview/backend/teleop/teleop.h
```diff
@@ -51,11 +53,36 @@ class TeleopService {
...
+  void UpdateOperatorDaemonRpt(
+      const std::shared_ptr<modules::car1::teleop::DaemonServiceRpt> &rpt);
+
+  std::shared_ptr<cyber::Writer<modules::car1::teleop::DaemonServiceCmd>>
+      car_daemon_cmd_writer_;
   std::shared_ptr<cyber::Writer<modules::car1::teleop::DaemonServiceCmd>>
-      daemon_cmd_writer_;
+      operator_daemon_cmd_writer_;
+  std::shared_ptr<cyber::Reader<modules::car1::teleop::DaemonServiceRpt>>
+      car_daemon_rpt_reader_;
```
unacao: need to have 'operator_daemon_rpt_reader_' instance, too.

modules/dreamview/backend/teleop/teleop.cc
hugomatic: Done, it is more efficient. Should I also check for a new value to avoid locking unless the value has changed?

modules/dreamview/backend/teleop/teleop.cc
hugomatic: I admit I did not think about the car requesting the operator daemon outcome, and 
we were only thinking about the operator's perspective for the UI.  We don't even relay the operator daemon's report to the car. But we can change it.

modules/dreamview/backend/teleop/teleop.cc
hugomatic: As I stated previously, this only works from the operator's perspective and this backend may not work the way you want (yet).

modules/dreamview/backend/teleop/teleop.h
hugomatic: At this moment each modem has a channel (ModemInfo), and its rank is in the message.
It sort of works, but it has limitations (for example, a fleet of cars where some cars don't have 3 modems).
We can combine this into a single channel with a repeated field, which is closer to what the original backend expected. This is probably something I would consider doing even if you hadn't brought it up.

modules/dreamview/backend/teleop/teleop.cc
hugomatic: Done

modules/dreamview/backend/teleop/teleop.cc
hugomatic: OK, I removed all of them.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +132,82 @@ void TeleopService::RegisterMessageHandlers() {
...
+    AINFO << "status: " << to_send;
   }
   websocket_->SendData(conn, to_send);
 }
 
+
+void TeleopService::UpdateModem(unsigned int index,
+    const std::shared_ptr<ModemInfo> &modem_info) {
+
+    boost::unique_lock<boost::shared_mutex> writer_lock(mutex_);
```
unacao: please move this lock to line 167

modules/dreamview/backend/teleop/teleop.cc
unacao: this issue is not addressed yet

modules/dreamview/backend/teleop/teleop.cc
unacao: please move the lock to line 188

modules/dreamview/backend/teleop/teleop.cc
unacao: If we don't consider the car requesting the operator daemon outcome, then why do we need two readers for daemon rpt and two callback functions handling them separately? I am ok with just showing car daemon outcome.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -39,11 +41,36 @@ TeleopService::TeleopService(WebSocketHandler *websocket)
...
   // TODO get topic names from proto
   // TODO update proto to get all modems' info combined with rank
-  modem_info_reader_ = node_->CreateReader<ModemInfo>(
-      "/apollo/networks/modems",
-      [this](const std::shared_ptr<ModemInfo> &msg) { UpdateModemInfo(msg); });
-  daemon_cmd_writer_ = node_->CreateWriter<DaemonServiceCmd>(
-      "/apollo/teleop/daemon_service/cmd");
+
+  modem0_info_reader_ = node_->CreateReader<ModemInfo>(
+      "/apollo/teleop/network/modem0",
```
jmtao: move topic strings def to a central place instead of hard-coded it here and there.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -39,11 +41,36 @@ TeleopService::TeleopService(WebSocketHandler *websocket)
...
+      [this](const std::shared_ptr<ModemInfo> &msg) { UpdateModem(0, msg); });
```
jmtao: better not use maggic number 0, 1, 2...
use macros or gflags

modules/dreamview/backend/teleop/teleop.cc
hugomatic: The car's daemon is used to start / stop / check the video encoding and the audio encoding in the car. The operator daemon is used to start / stop /check the audio in the operator computer.  

Without the operator daemon, controlling the operator voip encoder from the operator console computer would need to be done with a different mechanism or a separate UI (the task manager).

unacao: I understand we need two daemons working on each side. But here is about the reporting status from each daemon. In your previous comment, you mentioned there will not be operator daemon status reported to the car, which is ok for me. If that is the case, we only need one reader to subscribe to the car daemon status report, right?

hugomatic: Without the operator daemon report / command, how should we stop/start and check the status the voip_encoder in the local (operator) machine? This seamed like the solution to that problem.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -180,10 +180,8 @@ void TeleopService::UpdateCarDaemonRpt(
           if (service.find("voip_encoder") >= 0) {
               voipIsRunning = true;
           }          
-	  else {
-	      if (service.find("encoder") >= 0) {
-                  aVideoEncoderIsRunning = true; 
-	      }
+	  else if (service.find("encoder") >= 0) {
```
unacao: this line should follow the bracket on line 182

modules/dreamview/backend/teleop/teleop.cc
hugomatic: fixed [here](https://github.com/ApolloAuto/apollo/pull/9963/commits/c316aa8253165a9ced4524995c176569ca781045)

modules/dreamview/backend/teleop/teleop.cc
unacao: > I admit I did not think about the car requesting the operator daemon outcome, and
> we were only thinking about the operator's perspective for the UI. We don't even relay the operator daemon's report to the car. But we can change it.

All my discussion about not needing the report status of operator daemon is based on this comment. If we do have both daemons reporting its status to the remote side, then we just need to set the appropriate teleop_status_ fields correctly. I did not want to remove the report status of operator daemon in the first place. Also, have a daemon running vs. reporting its status are two separate issues, right? I was just saying that if you did not think about reporting the operator daemon outcome while it is indeed running, it is ok with me.

modules/dreamview/backend/teleop/teleop.cc
hugomatic: done [here](https://github.com/ApolloAuto/apollo/pull/9963/commits/075cdc40a1d7c2084e86aed38c980b4e5b433b83)

modules/dreamview/backend/teleop/teleop.cc
hugomatic: I removed the number and put the modem id in a string. 

This is a change to the original code that used the provider to name the modem. 
I changed it because the provider (ex Verizon or AT&T) can be the same for several modems.

```
 teleop_status_["modems"][modem_info->provider()] = modem_info->technology();
```

see new code [here](https://github.com/ApolloAuto/apollo/pull/9963/commits/182cb16555cbc3398c3cd162c885f924f3d05207)

modules/dreamview/backend/teleop/teleop.cc
hugomatic: should be fixed now.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +149,70 @@ void TeleopService::RegisterMessageHandlers() {
...
   websocket_->SendData(conn, to_send);
 }
 
+
+void TeleopService::UpdateModem(const std::string &modem_id,
+    const std::shared_ptr<ModemInfo> &modem_info) {
+
+    // TODO simplify data and only send necessary info for display
+    // update modem_info_
+    if (modem_info->has_provider() && modem_info->has_technology()) {
```
vlin17: if provider info isn't used, we don't need modem_info->has_provider() in the if statement.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -104,23 +149,70 @@ void TeleopService::RegisterMessageHandlers() {
...
+}
+
+void TeleopService::UpdateOperatorDaemonRpt(
+    const std::shared_ptr<DaemonServiceRpt> &daemon_rpt) {
+  {
+      bool voipIsRunning = false;
+      for (int i = 0; i < daemon_rpt->services_size(); i++) {
+          std::string service =  daemon_rpt->services(i);
+          if (service.find("voip_encoder") >= 0) {
+              voipIsRunning = true;
```
vlin17: maybe break the for loop once voipIsRunning is set?

modules/dreamview/backend/teleop/teleop.cc
hugomatic: fixed [here](https://github.com/ApolloAuto/apollo/pull/9963/commits/84332ac9b452b2a739a1db47fb25f1568fa85827)

modules/dreamview/backend/teleop/teleop.cc
hugomatic: added in this [commit](https://github.com/ApolloAuto/apollo/pull/9963/commits/84332ac9b452b2a739a1db47fb25f1568fa85827)

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -39,11 +57,37 @@ TeleopService::TeleopService(WebSocketHandler *websocket)
...
-      [this](const std::shared_ptr<ModemInfo> &msg) { UpdateModemInfo(msg); });
-  daemon_cmd_writer_ = node_->CreateWriter<DaemonServiceCmd>(
-      "/apollo/teleop/daemon_service/cmd");
+
+  modem0_info_reader_ = node_->CreateReader<ModemInfo>(
+      modem0_channel,
+      [this](const std::shared_ptr<ModemInfo> &msg) {UpdateModem(modem0_id, msg); 
+      });
+
+  modem1_info_reader_ = node_->CreateReader<ModemInfo>(
```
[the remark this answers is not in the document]
hugomatic: added in this [commit](https://github.com/ApolloAuto/apollo/pull/9963/commits/84332ac9b452b2a739a1db47fb25f1568fa85827)

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -39,11 +57,37 @@ TeleopService::TeleopService(WebSocketHandler *websocket)
...
   // TODO update proto to get all modems' info combined with rank
-  modem_info_reader_ = node_->CreateReader<ModemInfo>(
-      "/apollo/networks/modems",
-      [this](const std::shared_ptr<ModemInfo> &msg) { UpdateModemInfo(msg); });
-  daemon_cmd_writer_ = node_->CreateWriter<DaemonServiceCmd>(
-      "/apollo/teleop/daemon_service/cmd");
+
+  modem0_info_reader_ = node_->CreateReader<ModemInfo>(
+      modem0_channel,
+      [this](const std::shared_ptr<ModemInfo> &msg) {UpdateModem(modem0_id, msg);
```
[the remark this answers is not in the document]
hugomatic: added in this [commit](https://github.com/ApolloAuto/apollo/pull/9963/commits/84332ac9b452b2a739a1db47fb25f1568fa85827)

modules/dreamview/backend/teleop/teleop.cc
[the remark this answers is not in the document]
hugomatic: I agree that we're probably not going to want to touch the readers much during the life cycle of the backend. However if we don't keep the reader around, isn't there a chance that the [shutdown](https://github.com/ApolloAuto/apollo/blob/master/cyber/node/reader.h#L241) will be triggered at the end of the scope?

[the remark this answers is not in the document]
vlin17: Sure, if you think there's a need to manually trigger the shutdown later on, we can keep those variables.

PR #9999, opened 2019-10-23
planning: add cruise_speed as a param to speed optimizers

jmtao:


modules/planning/common/reference_line_info.cc
```diff
@@ -131,6 +131,10 @@ void ReferenceLineInfo::SetCandidatePathBoundaries(
   candidate_path_boundaries_ = std::move(path_boundaries);
 }
 
+const double ReferenceLineInfo::GetCruiseSpeed() const {
+  return cruise_speed_ > 0.0 ? cruise_speed_ : FLAGS_default_cruise_speed;
+}
```
xiaoxq: No need to return a `const` copy.

jmtao: amended. thx

PR #10004, opened 2019-10-23
Prediction: remove the pose container dependency from evaluators

kechxu:


modules/prediction/evaluator/pedestrian/pedestrian_interaction_evaluator.cc
```diff
@@ -44,6 +44,7 @@ PedestrianInteractionEvaluator::PedestrianInteractionEvaluator()
   LoadModel();
 }
 
+/* TODO(kechxu) figure out if this function is necessary. It is not being used
```
xiaoxq: Just remove.

PR #10011, opened 2019-10-24
Control: MRAC controller higher-order upgrade

kevin-y-wang:


modules/control/common/mrac_controller.cc
```diff
@@ -19,23 +19,33 @@
...
 
-double MracController::Control(const double command, const double state,
+using Matrix = Eigen::MatrixXd;
+
+double MracController::Control(const double command, const Matrix state,
                                const double dt) {
-  // check if the reference model well set up during the initilization
+  // check if the reference/adaption model well set up during the initilization
   if (!reference_model_enabled_) {
-    TransformReferenceModel(dt);
```
Capri2014: Don not build reference/adaption model in `::Control(...)`, it should only be done once in `Init()` and after that only update the model if needed.

kevin-y-wang: Fixed per comment. Thanks

modules/control/common/mrac_controller.cc
```diff
@@ -87,28 +98,25 @@ double MracController::Control(const double command, const double state,
...
-  state_reference_[1] = state_reference_[0];
-  state_action_[1] = state_action_[0];
-  input_desired_[1] = input_desired_[0];
+  state_reference_.col(1) = state_reference_.col(0);
+  state_action_.col(1) = state_action_.col(0);
+  input_desired_.col(1) = input_desired_.col(0);
   control_previous_ = control;
   return control;
-}
+}  // namespace control
```
sjiang2018: remove // ....

modules/control/common/mrac_controller.cc
Capri2014: merge these two `if` please

kevin-y-wang: Fixed

modules/control/common/mrac_controller.cc
kevin-y-wang: Thanks. Fixed per comment

Capri2014:
1. Separate `Reset` to `Reset`, `ResetStates` and `ResetGain`, sometimes we may need only clear the internal state variables.

2. For the gain matrix(continuous one), it is always good to have the `set` functions as well in case we may need to adjust gains online in the future (i.e. change of scenarios).

kevin-y-wang:
> 1. Separate `Reset` to `Reset`, `ResetStates` and `ResetGain`, sometimes we may need only clear the internal state variables.
> 2. For the gain matrix(continuous one), it is always good to have the `set` functions as well in case we may need to adjust gains online in the future (i.e. change of scenarios).

Refactored per comment. Thanks

PR #10013, opened 2019-10-24
Planning: dynamically update lane-change preparation distance

JasonXu12:


modules/planning/proto/planning_status.proto
```diff
@@ -23,7 +23,7 @@ message ChangeLaneStatus {
   enum Status {
     IN_CHANGE_LANE = 1;       // during change lane state
     CHANGE_LANE_FAILED = 2;   // change lane failed
-    CHANGE_LANE_SUCCESS = 3;  // change lane succeeded
+    CHANGE_LANE_FINISHED = 3;  // change lane finished
```
xiaoxq: Prefer not to rename an existing field or value, which could break forward-compatibility.

Prefer:
```
CHANGE_LANE_SUCCESS = 3 [deprecated = true];
CHANGE_LANE_FINISHED = 4;
```

But it's up to planning module owners. Please dismiss my comment if they feel OK.

jmtao: this is a proto used internally by planning (i.e. neither shared with other modules, nor write to outgoing msg/topic).  it's ok to update existing field.

JasonXu12: Thanks for the suggestion. As jmtao said, the proto was only used by the planning module and I think it's good to update this field.

PR #10022, opened 2019-10-25
planning: updated ParkAndGo configs

sjiang2018:


modules/planning/open_space/coarse_trajectory_generator/hybrid_a_star.cc
```diff
@@ -113,6 +114,10 @@ bool HybridAStar::ValidityCheck(std::shared_ptr<Node3d> node) {
       for (const common::math::LineSegment2d& linesegment :
            obstacle_linesegments) {
         if (bounding_box.HasOverlap(linesegment)) {
+          ADEBUG << "collision start at x: " << linesegment.start().x();
+          ADEBUG << "collision start at y: " << linesegment.start().y();
+          ADEBUG << "collision end at x: " << linesegment.end().x();
+          ADEBUG << "collision end at y: " << linesegment.end().y();
```
xiaoxq: I checked the implementation of LineSegment2d. The usage here is correct :)

But actually it should not be named as `end()`, which generally means a out-of-bound iterator/object in C++. It should use `last()` or `back()` IMO.

modules/planning/open_space/coarse_trajectory_generator/hybrid_a_star.cc
```diff
@@ -24,6 +24,7 @@
 
 namespace apollo {
 namespace planning {
+// #define ADEBUG AINFO
```
Capri2014: remove this line please.

modules/planning/scenarios/util/util.cc
```diff
@@ -305,7 +305,7 @@ bool CheckADCHeading(const common::math::Vec2d adc_position,
                      const double adc_heading,
                      const ReferenceLineInfo& reference_line_info,
                      const double heading_diff_to_reference_line) {
-  const double kReducedHeadingBuffer = 0.2;  // TODO(Shu) move to config
+  const double kReducedHeadingBuffer = 0.3;  // TODO(Shu) move to config
```
Capri2014: add comments for the units, otherwise people might get confused. (ratio? radiance? degree?)

modules/planning/open_space/coarse_trajectory_generator/hybrid_a_star.cc
sjiang2018: Fixed. Thx!

PR #10027, opened 2019-10-25
control & common: add control preprocessor topic as an internal topic for control submodules

sjiang2018:


modules/control/submodules/preprocessor_submodule.h
```diff
@@ -155,14 +148,15 @@ class PreprocessorSubmodule : public apollo::cyber::TimerComponent {
   std::shared_ptr<cyber::Reader<apollo::localization::LocalizationEstimate>>
       localization_reader_;
   std::shared_ptr<cyber::Reader<planning::ADCTrajectory>> trajectory_reader_;
-  std::shared_ptr<cyber::Writer<apollo::control::ControlCommand>>
-      control_command_writer_;
+
+  std::shared_ptr<cyber::Writer<apollo::control::Preprocessor>>
+      preprocessor_writer_;
```
xiaoxq: You don't need the `apollo::control::` prefix as you are already inside the namespace. Then it fits in one line.

PR #10028, opened 2019-10-25
Prediction: implement get container output

kechxu:


modules/prediction/submodules/container_output.cc
```diff
@@ -15,16 +15,49 @@
...
 
 #include "modules/prediction/submodules/container_output.h"
+#include "modules/prediction/common/prediction_gflags.h"
 
 namespace apollo {
 namespace prediction {
 
 ContainerOutput::ContainerOutput() {
-  // TODO(kechxu) implement
+  Clear();
```
xiaoxq: Why we need this?

modules/prediction/submodules/container_output.cc
```diff
@@ -15,16 +15,45 @@
...
+    const std::vector<int>& curr_frame_considered_obstacle_ids) {
+  curr_frame_considered_obstacle_ids_ = curr_frame_considered_obstacle_ids;
+}
+
+void ContainerOutput::Clear() {
+  curr_frame_obstacles_.clear();
+  curr_frame_perception_obstacles_.clear();
+  curr_frame_movable_obstacle_ids_.clear();
+  curr_frame_unmovable_obstacle_ids_.clear();
+  curr_frame_considered_obstacle_ids_.clear();
```
xiaoxq: Sorry that I didn't make it clear. I mean, even the `Clear()` is not needed. std containers are guaranteed to be cleared on destruction.

xiaoxq: If you need to explicitly clear them in some occasions, it's reasonable to have this function. But you still don't need to call it from the destructor.

PR #10060, opened 2019-10-30
Planning: use dp st warm start in speed opt

JasonZhou404:


modules/planning/tasks/optimizers/piecewise_jerk_speed/piecewise_jerk_speed_nonlinear_optimizer.cc
```diff
@@ -149,12 +149,31 @@ Status PiecewiseJerkSpeedNonlinearOptimizer::Process(
...
+  for (int i = 0; i < num_of_knots; ++i) {
+    SpeedPoint sp;
+    if (!reference_speed_data.EvaluateByTime(i * delta_t, &sp)) {
+      std::string msg("retriving dp st data for speed optimization failed!");
+      AERROR << msg;
+      speed_data->clear();
+      return Status(ErrorCode::PLANNING_ERROR, msg);
+    }
+    warm_start.emplace_back(
+        std::initializer_list<double>{sp.s(), sp.v(), sp.a()});
```
xiaoxq: I guess that the `std::initializer_list<double>` is not needed, as `{}` already means an initializer_list.

PR #10062, opened 2019-10-30
Prediction: pass some side information through submodules

kechxu:


modules/prediction/submodules/container_submodule.cc
```diff
@@ -69,7 +69,7 @@ bool ContainerSubmodule::Init() {
 bool ContainerSubmodule::Proc(
     const std::shared_ptr<PerceptionObstacles>& perception_message) {
   MessageProcess::ContainerProcess(*perception_message);
-
+  double frame_start_time = Clock::NowInSeconds();
```
xiaoxq: const

modules/prediction/submodules/evaluator_submodule.cc
```diff
@@ -46,10 +46,19 @@ bool EvaluatorSubmodule::Init() {
 
 bool EvaluatorSubmodule::Proc(
     const std::shared_ptr<ContainerOutput>& container_output) {
+  apollo::common::Header perception_header =
```
xiaoxq: const &

PR #10069, opened 2019-10-31
Control: refactor MRAC controller anti-windup design and bounds settings

kevin-y-wang:


modules/control/common/mrac_controller.h
```diff
@@ -135,6 +138,18 @@ class MracController {
...
 
+  /**
+   * @brief bound the system output with the given bound and change-rate bound
+   * @param output_unbounded original system output without bound
+   * @param previous_output system output in the last step
+   * @param dt sampling time interval
+   * @param saturation_status system saturation status indicator
+   * @return bounded system output
+   */
+  double BoundOutput(const double output_unbounded,
```
Capri2014: usually you want the status to be the return value of the function.

kevin-y-wang: Fixed per comment. Thanks

PR #10092, opened 2019-11-02
cyber:fixed log level

mickeyouyou:


mickeyouyou:
> Fix the commit message to `Cyber: Fixed log level.`, which is the most preferred format.
> 
> 1. No additional chars like `*`.
> 2. Space after `:`.
> 3. Title-case for both module name and message sentence.
> 4. As it's a sentence, end with `.`.

Done.

PR #10095, opened 2019-11-04
paladin: latency monitoring and aggregation across modules

michael4338:


modules/monitor/software/latency_monitor.cc
```diff
@@ -0,0 +1,106 @@
...
+
+  if (current_time - flush_time_ > FLAGS_latency_report_interval) {
+    flush_time_ = current_time;
+    PublishLatencyReport();
+  }
+}
+
+void LatencyMonitor::UpdateLatencyStat(
+    const std::shared_ptr<LatencyRecordMap>& records) {
+  if (latency_report_.stat_aggr().find(records->module_name()) ==
```
xiaoxq: Store `latency_report_.mutable_stat_aggr()` in a local pointer and reuse it. So the code is much simpler.

modules/monitor/software/latency_monitor.cc
```diff
@@ -0,0 +1,106 @@
...
+      latency_report_.stat_aggr().end()) {
+    LatencyStat stat;
+    latency_report_.mutable_stat_aggr()->insert({records->module_name(), stat});
```
xiaoxq: Try `emplace`.

modules/monitor/software/latency_monitor.cc
```diff
@@ -0,0 +1,106 @@
...
+  }
+  auto* stat = &latency_report_.mutable_stat_aggr()->at(records->module_name());
+
+  uint64_t min_duration = (1UL << 63), max_duration = 0,
+           total_duration = 0;
+  for (const auto record : records->latency_records()) {
```
xiaoxq: `auto&`

modules/monitor/software/latency_monitor.h
```diff
@@ -0,0 +1,51 @@
...
+#include <vector>
+
+#include "modules/common/latency_recorder/proto/latency_record.pb.h"
+#include "modules/monitor/common/recurrent_runner.h"
+
+namespace apollo {
+namespace monitor {
+
+using apollo::common::LatencyRecordMap;
+using apollo::common::LatencyReport;
```
xiaoxq: Avoid making alias in header file. They introduced `apollo::monitor::LatencyReport` into the binary which is confusing and could cause problem.

modules/monitor/software/latency_monitor.cc
michael4338: the protobuf map does not seem to have "emplace" member function?

PR #10097, opened 2019-11-04
Teleop:  Car Teleop & Reset button

vlin17:
- adding Car Teleop Monitor
- adding Resume feature to Console teleop
- setting monitor width based on monitor types


modules/dreamview/frontend/src/components/TeleopMonitor/CarTeleop.js
```diff
@@ -0,0 +1,57 @@
...
+        return (
+            <div className="monitor teleop">
+                <div className="monitor-header">
+                    <div className="title">Car Teleop Controls</div>
+                </div>
+
+                <div className="monitor-content">
+                    <MonitorSection title="Audio" icon={itemIcon}>
+                        <div className="section-content-row" key={name}>
+                            <CheckboxItem id='teleopAudio'
```
unacao: do we have mic mute button for car teleop?

hugomatic: I think it would be nice to have one

PR #10106, opened 2019-11-05
Start stop encoders

hugomatic:
This adds logic to start / stop video encoders (2 for now).

The first message to start is sent when the user toggles the switch (setting a "starting" status). During the daemon report callback, if the encoders are not all started, then more messages are sent. 

In this PR, the UI is not "aware" of the "starting" / "stopping" status. A single "stopped, starting, started, stopping" list could replace the 4 booleans value.  One thing to note is that if the video is "starting" and the user presses the switch, the backend will go to "stopping" mode. This behavior might surprise the user.

Next, the same recipe should be applicable to audio and "Pull over" (using the planner msg instead of the daemon report). These should be simpler since there's only one audio encoding node, and one pad planner message to send.

### DIFF
--- a/modules/dreamview/backend/teleop/teleop.cc
+++ b/modules/dreamview/backend/teleop/teleop.cc
@@ -30,7 +30,6 @@ using ::google::protobuf::util::MessageToJsonString;
 using modules::teleop::network::ModemInfo;
 using modules::teleop::teleop::DaemonServiceCmd;
 using modules::teleop::teleop::DaemonServiceRpt;
-using apollo::planning::PadMessage;
 using apollo::cyber::Time;
 
 // modem ids
--- a/modules/dreamview/backend/teleop/teleop.cc
+++ b/modules/dreamview/backend/teleop/teleop.cc
@@ -289,7 +289,6 @@ void TeleopService::UpdateOperatorDaemonRpt(
 }
 
 void TeleopService::SendVideoStreamCmd(bool start_stop) {
-  // auto msg = std::make_shared<DaemonServiceCmd>();
   DaemonServiceCmd msg;
   if (start_stop) {
     msg.set_cmd("start");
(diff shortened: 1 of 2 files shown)


modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -182,37 +216,94 @@ void TeleopService::UpdateModem(const std::string &modem_id,
...
+      teleop_status_["mic"] = voipIsRunning;
+  }
+}
+
+void TeleopService::SendVideoStreamCmd(bool start_stop) {
+  DaemonServiceCmd msg;
+  if (start_stop) {
+    msg.set_cmd("start");
+  }
+  else {
```
xiaoxq: Please fix the style issue with `scripts/clang-format.sh`.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -182,37 +216,94 @@ void TeleopService::UpdateModem(const std::string &modem_id,
...
+          if (service.find("voip_encoder") != std::string::npos) {
+              voipIsRunning = true;
+              break;
+          }
       }
-    }
-    boost::unique_lock<boost::shared_mutex> writer_lock(mutex_);
-    teleop_status_["mic"] = voipIsRunning;
+      boost::unique_lock<boost::shared_mutex> writer_lock(mutex_);
+      teleop_status_["mic"] = voipIsRunning;
```
xiaoxq: Wrong indents.

modules/dreamview/backend/teleop/teleop.cc
```diff
@@ -18,6 +18,8 @@
 
 #include "cyber/common/log.h"
 #include "google/protobuf/util/json_util.h"
+#include "cyber/time/time.h"
+#include "modules/common/util/message_util.h"
```
xiaoxq: Make the headers well sorted.

modules/dreamview/backend/teleop/teleop.cc
hugomatic: fixed [here](https://github.com/ApolloAuto/apollo/pull/10106/commits/e52de49fccb6d559e548bf24fda32d77c7930df3)

PR #10109, opened 2019-11-05
planning: modified lane_change process() function

sjiang2018:


panjiacheng:
Why remove all the ADEBUG messages?

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -200,13 +208,16 @@ void LaneChangeDecider::PrioritizeChangeLane(
...
+  return;
+  /* disable rest to avoid core dump */
+  ADEBUG << "removed change lane";
   auto iter = reference_line_info->begin();
   while (iter != reference_line_info->end()) {
     if (iter->IsChangeLanePath()) {
-      iter = reference_line_info->erase(iter);
-    } else {
-      ++iter;
+      reference_line_info->erase(iter);
```
JasonXu12: Please don't change here as there might be more than one lane_change reference_line in the list to remove.

JasonXu12:
> Why remove all the ADEBUG messages?
> I guess you can leave them there if they are helpful.

JasonXu12:
> > Why remove all the ADEBUG messages?
> > I guess you can leave them there if they are helpful.
Most of the messages can be found from planning context ChangeLaneStatus message. No need to do the duplicate logging here.

PR #10111, opened 2019-11-05
planning: modified lane_change process() function

sjiang2018:


modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -200,6 +208,9 @@ void LaneChangeDecider::PrioritizeChangeLane(
 
 void LaneChangeDecider::RemoveChangeLane(
     std::list<ReferenceLineInfo>* reference_line_info) const {
+  return;
```
JasonXu12: Please add a TODO here. The core dump will be fixed.

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -34,7 +34,9 @@ using apollo::common::util::StrCat;
 LaneChangeDecider::LaneChangeDecider(const TaskConfig& config)
     : Decider(config) {}
 
-Status LaneChangeDecider::Process(Frame* frame) {
+// added a dummy parameter to enable this task in ExecuteTaskOnReferenceLine
+Status LaneChangeDecider::Process(
+    Frame* frame, ReferenceLineInfo* current_reference_line_info) {
```
JasonXu12: add const whenever possible

sjiang2018: Fixed. Thank you!

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
sjiang2018: Fixed. Thank you!

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.h
```diff
@@ -65,7 +65,8 @@ class LaneChangeDecider : public Decider {
       const ReferenceLineInfo* const reference_line_info);
 
  private:
-  common::Status Process(Frame* frame) override;
+  common::Status Process(Frame* frame,
```
panjiacheng: If not modifying, better use const reference than pointer;
If modifying, better use * const.

xiaoxq: Agree on the const reference part. But IMO it's OK to use `*` instead of `* const`, as redirecting a pointer is virtually never allowed. So we just save several bytes :)

JasonXu12: Agree. Better use const ReferenceLineInfo& reference_line_info here.

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -34,7 +34,9 @@ using apollo::common::util::StrCat;
 LaneChangeDecider::LaneChangeDecider(const TaskConfig& config)
     : Decider(config) {}
 
-Status LaneChangeDecider::Process(Frame* frame) {
+// added a dummy parameter to enable this task in ExecuteTaskOnReferenceLine
+Status LaneChangeDecider::Process(
+    Frame* frame, ReferenceLineInfo* const current_reference_line_info) {
```
JasonXu12: Please keep the param name the same as your header file.

PR #10138, opened 2019-11-07
Prediction: set caution within a certain distance and refactor the APIs in prioritization

kechxu:
…


modules/prediction/common/prediction_gflags.cc
```diff
@@ -76,6 +76,8 @@ DEFINE_double(caution_search_distance_backward_for_overlap, 30.0,
               "in the case of overlap");
 DEFINE_double(caution_pedestrian_approach_time, 3.0,
               "The time for a pedestrian to approach adc trajectory");
+DEFINE_double(caution_distance_threshold, 60.0,
+              "The distance threshold for obstacles to be caution");
```
xiaoxq: I just learned from https://abseil.io/tips/45 that, if a FLAG is rarely changed and only used in one file, just make it a constant. Up to you.

PR #10144, opened 2019-11-08
control: minor fixes in mrac controller

sjiang2018:


modules/control/common/mrac_controller.cc
```diff
@@ -228,50 +228,53 @@ void MracController::BuildReferenceModel() {
...
-      AWARN << "reference model order beyond the designed range, "
-               "model_order: "
-            << model_order_;
-      reference_model_enabled_ = false;
-    }
+    return;
+  }
+
+  if (model_order_ > 2) {
+    AWARN << "reference model order beyond the designed range, "
```
Capri2014: This should be a ERROR

modules/control/common/mrac_controller.cc
```diff
@@ -228,50 +228,53 @@ void MracController::BuildReferenceModel() {
...
+    matrix_b_reference_(0, 0) = 1.0 / tau_reference_;
+  } else if (model_order_ == 2) {
+    matrix_a_reference_(0, 1) = 1.0;
+    matrix_a_reference_(1, 0) = -wn_reference_ * wn_reference_;
+    matrix_a_reference_(1, 1) = -2 * zeta_reference_ * wn_reference_;
+    matrix_b_reference_(1, 0) = wn_reference_ * wn_reference_;
   }
 }
 
 void MracController::BuildAdaptionModel() {
```
Capri2014: In next iteration, please return the function of StatusCode instead of `void`

PR #10174, opened 2019-11-12
Update Apollo_3.5_Software_Architecture.md

OCILY:


xiaoxq:
Good catch! Could you sign the CLA so we can merge it? Thanks!

PR #10178, opened 2019-11-12
planning: add tests

jmtao:


modules/planning/scenarios/emergency/emergency_stop/stage_approach_test.cc
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+
+#define protected public
+#define private public
```
xiaoxq: These hacks seem neither recommended nor needed.

jmtao: fixed in next commit. cleaned up this bad hack at all other places as well.

PR #10181, opened 2019-11-12
Planning: update st-bounds and v-limits info into st_graph_data.

panjiacheng:


modules/planning/common/st_graph_data.h
```diff
@@ -71,6 +90,9 @@ class StGraphData {
   double path_length_by_conf_ = 0.0;
   double total_time_by_conf_ = 0.0;
   planning_internal::STGraphDebug* st_graph_debug_ = nullptr;
+
+  std::vector<std::tuple<double, double, double>> s_boundary_;
+  std::vector<std::tuple<double, double, double>> v_limits_;
```
JasonZhou404: Could we possibly use a proto message on defining the the tuple included in s_boundary_? so that we can combine the tuples in s_boundary_ and v_limits_ together

modules/planning/tasks/deciders/st_bounds_decider/st_bounds_decider.cc
```diff
@@ -105,18 +109,22 @@ void STBoundsDecider::InitSTBoundsDecider(
...
     st_bound->emplace_back(curr_t, std::numeric_limits<double>::lowest(),
                            std::numeric_limits<double>::max());
+    vt_bound->emplace_back(curr_t, std::numeric_limits<double>::lowest(),
+                           std::numeric_limits<double>::max());
   }
 
   // Sweep-line to get detailed ST-boundary.
   for (size_t i = 0; i < st_bound->size(); ++i) {
-    double t, s_lower, s_upper;
+    double t, s_lower, s_upper, v_lower, v_upper;
```
JasonZhou404: what is the meanings of v_lower and v_upper? the speed limit of the adc?  a bit confusing here

modules/planning/common/st_graph_data.h
panjiacheng: fixed this.

PR #10185, opened 2019-11-13
Planning: Prioritize non lane-change reference line when optimization…

sjiang2018:
… failed instead of RemoveChangeLane

### DIFF
--- a/modules/planning/scenarios/lane_follow/lane_follow_stage.cc
+++ b/modules/planning/scenarios/lane_follow/lane_follow_stage.cc
@@ -114,7 +114,7 @@ Stage::StageStatus LaneFollowStage::Process(
   ADEBUG << "Number of reference lines:\t"
          << frame->mutable_reference_line_info()->size();
 
-  unsigned count = 0;
+  unsigned int count = 0;
 
   for (auto& reference_line_info : *frame->mutable_reference_line_info()) {
     // TODO(SHU): need refactor
@@ -194,11 +194,12 @@ Status LaneFollowStage::PlanOnReferenceLine(
 
     RecordDebugInfo(reference_line_info, optimizer->Name(), time_diff_ms);
 
+    // TODO(SHU): disable reference line order changes for now
     // updated reference_line_info, because it is changed in
     // lane_change_decider by PrioritizeChangeLane().
-    reference_line_info = &frame->mutable_reference_line_info()->front();
-    ADEBUG << "Current reference_line_info is IsChangeLanePath: "
-           << reference_line_info->IsChangeLanePath();
+    // reference_line_info = &frame->mutable_reference_line_info()->front();
+    // ADEBUG << "Current reference_line_info is IsChangeLanePath: "
+    //        << reference_line_info->IsChangeLanePath();
   }
 
   RecordObstacleDebugInfo(reference_line_info);
--- a/modules/planning/scenarios/lane_follow/lane_follow_stage.cc
+++ b/modules/planning/scenarios/lane_follow/lane_follow_stage.cc
@@ -114,7 +114,16 @@ Stage::StageStatus LaneFollowStage::Process(
   ADEBUG << "Number of reference lines:\t"
          << frame->mutable_reference_line_info()->size();
 
+  unsigned count = 0;
+
   for (auto& reference_line_info : *frame->mutable_reference_line_info()) {
+    // TODO(SHU): need refactor
+    if (count++ == frame->mutable_reference_line_info()->size()) {
+      break;
+    }
+    ADEBUG << "No: [" << count << "] Reference Line.";
+    ADEBUG << "IsChangeLanePath: " << reference_line_info.IsChangeLanePath();
+
     if (has_drivable_reference_line) {
       reference_line_info.SetDrivable(false);
       break;
@@ -126,6 +135,8 @@ Stage::StageStatus LaneFollowStage::Process(
     if (cur_status.ok()) {
       if (reference_line_info.IsChangeLanePath()) {
         ADEBUG << "reference line is lane change ref.";
+        ADEBUG << "FLAGS_enable_smarter_lane_change: "
+               << FLAGS_enable_smarter_lane_change;
         if (reference_line_info.Cost() < kStraightForwardLineCost &&
             (LaneChangeDecider::IsClearToChangeLane(&reference_line_info) ||
              FLAGS_enable_smarter_lane_change)) {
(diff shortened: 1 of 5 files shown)


modules/planning/scenarios/lane_follow/lane_follow_stage.cc
```diff
@@ -233,8 +252,8 @@ Status LaneFollowStage::PlanOnReferenceLine(
         }
       }
       if (add_stop_obstacle_cost) {
-        constexpr double kRefrenceLineStaticObsCost = 1e3;
-        reference_line_info->AddCost(kRefrenceLineStaticObsCost);
+        constexpr double kReferenceLineStaticObsCost = 1e3;
```
xiaoxq: static

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -188,37 +193,48 @@ void LaneChangeDecider::UpdateStatus(double timestamp,
 }
 
 void LaneChangeDecider::PrioritizeChangeLane(
-    std::list<ReferenceLineInfo>* reference_line_info) const {
+    std::list<ReferenceLineInfo>* reference_line_info,
+    const bool is_prioritize_change_lane) const {
```
xiaoxq: Put input param before output param. If the input list is empty, prefer
```
std::list<ReferenceLineInfo> LaneChangeDecider::PrioritizeChangeLane(const bool is_prioritize_change_lane) const;
```
which is even cleaner.

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.h
```diff
@@ -79,11 +79,12 @@ class LaneChangeDecider : public Decider {
   void UpdateStatus(double timestamp, ChangeLaneStatus::Status status_code,
                     const std::string& path_id);
 
-  void PrioritizeChangeLane(
-      std::list<ReferenceLineInfo>* reference_line_info) const;
+  void PrioritizeChangeLane(std::list<ReferenceLineInfo>* reference_line_info,
+                            const bool is_prioritize_change_lane) const;
```
xiaoxq: ditto

modules/planning/scenarios/lane_follow/lane_follow_stage.cc
sjiang2018: fixed. thank you!

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
sjiang2018: fixed. thank you!

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.h
sjiang2018: fixed. thank you!

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -188,37 +192,48 @@ void LaneChangeDecider::UpdateStatus(double timestamp,
...
   auto iter = reference_line_info->begin();
   while (iter != reference_line_info->end()) {
-    if (iter->IsChangeLanePath()) {
-      reference_line_info->splice(reference_line_info->begin(),
-                                  *reference_line_info, iter);
+    ADEBUG << "iter->IsChangeLanePath(): " << iter->IsChangeLanePath();
+    /* is_prioritize_change_lane == true: prioritize change_lane_reference_line
+       is_prioritize_change_lane == false: prioritize
+       non_change_lane_reference_line */
+    if (is_prioritize_change_lane == iter->IsChangeLanePath()) {
```
JasonXu12: The logic seems to work. However, I would not recommend writing comparison in this way. 'is_prioritize_change_lane' and 'iter->IsChangeLanePath()' represents totally different meanings. It's more clear to compare them separately with 'True' and 'False' in your logic.

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -188,37 +192,48 @@ void LaneChangeDecider::UpdateStatus(double timestamp,
...
+      ADEBUG << "is_prioritize_change_lane: " << is_prioritize_change_lane;
+      ADEBUG << "iter->IsChangeLanePath(): " << iter->IsChangeLanePath();
       break;
     }
     ++iter;
   }
+  reference_line_info->splice(reference_line_info->begin(),
```
JasonXu12: Please also double-check what influence on the list iteration order in lane_follow_stage will be caused by the splice operation here. Originally, there was no influence because the lane_change reference line always appears at the front of the list for some reason and the order was never changed. However, after your modification, the order of the reference_line list could be changed while somewhere else in the program is doing the iteration on the list. https://github.com/ApolloAuto/apollo/blob/f2fa04cc578a4b75785f50fbbd9c1c30b0804885/modules/planning/scenarios/lane_follow/lane_follow_stage.cc#L118

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
sjiang2018: Fixed

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
sjiang2018: Fixed

modules/planning/scenarios/lane_follow/lane_follow_stage.cc
```diff
@@ -114,7 +114,16 @@ Stage::StageStatus LaneFollowStage::Process(
   ADEBUG << "Number of reference lines:\t"
          << frame->mutable_reference_line_info()->size();
 
+  unsigned count = 0;
```
xiaoxq: `unsigned int`.

modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
```diff
@@ -188,37 +193,53 @@ void LaneChangeDecider::UpdateStatus(double timestamp,
...
 
 void LaneChangeDecider::PrioritizeChangeLane(
+    const bool is_prioritize_change_lane,
     std::list<ReferenceLineInfo>* reference_line_info) const {
   if (reference_line_info->empty()) {
     AERROR << "Reference line info empty";
     return;
   }
+  // TODO(SHU): disable the reference line order change for now
+  return;
```
xiaoxq: Add a FLAG_enable_reference_line_order which default to false, and
```
if (!FLAG_enable_reference_line_order) {
  return;
}
```
So this piece of code seems better.

PR #10213, opened 2019-11-15
Planning: adjust obstacle l_buffer when changing lane

JasonXu12:


modules/planning/tasks/deciders/speed_bounds_decider/st_boundary_mapper.cc
```diff
@@ -305,7 +313,7 @@ bool STBoundaryMapper::GetOverlapBoundaryPoints(
               const auto& point_high = discretized_path.Evaluate(
                   high_s + discretized_path.front().s());
               if (!CheckOverlap(point_high, obs_box,
-                                FLAGS_nonstatic_obstacle_nudge_l_buffer)) {
+                                l_buffer)) {
```
xiaoxq: Try to refine the layout after making change.

JasonXu12: Will fix. Thank you.

modules/planning/tasks/deciders/speed_bounds_decider/st_boundary_mapper.cc
```diff
@@ -198,6 +199,14 @@ bool STBoundaryMapper::GetOverlapBoundaryPoints(
     return false;
   }
 
+  const auto* planning_status = PlanningContext::Instance()
+                                ->mutable_planning_status()
+                                ->mutable_change_lane();
+
+  double l_buffer = FLAGS_nonstatic_obstacle_nudge_l_buffer;
+  if (planning_status->status() == ChangeLaneStatus::IN_CHANGE_LANE) {
+    l_buffer = 0.0;
```
jmtao: 0.0 is too aggressive.  e.g a vehicle in front of adc at the original lane may get ignored while status is IN_CHANGE_LANE?

JasonXu12: Set the value to 0.3m now. Thanks.

modules/planning/tasks/deciders/speed_bounds_decider/st_boundary_mapper.cc
```diff
@@ -198,6 +199,15 @@ bool STBoundaryMapper::GetOverlapBoundaryPoints(
...
   }
 
+  const auto* planning_status = PlanningContext::Instance()
+                                    ->mutable_planning_status()
+                                    ->mutable_change_lane();
+
+  double l_buffer =
+      planning_status->status() == ChangeLaneStatus::IN_CHANGE_LANE
+          ? FLAGS_lane_change_obstacle_nudge_l_buffer
+          : FLAGS_nonstatic_obstacle_nudge_l_buffer;
```
jmtao: do not use the gflag for nonstatic_obstacle_nudge_l_buffer just because they have the same value. 
that may introduce bugs when we change that gflag w/o being aware of the implact on change lane case. 
use another gflag or (existing gflag - a adjustment value).

PR #10214, opened 2019-11-15
Prediction: implement sigma ellipse params

kechxu:


modules/prediction/evaluator/vehicle/semantic_lstm_evaluator.cc
```diff
@@ -169,6 +169,19 @@ bool SemanticLSTMEvaluator::Evaluate(Obstacle* obstacle_ptr,
     point->mutable_gaussian_info()->set_sigma_y(sigma_y);
     point->mutable_gaussian_info()->set_correlation(corr);
 
+    constexpr int ellipse_gap = 5;
```
xiaoxq: static

PR #10230, opened 2019-11-16
Localization: Quick fix on std::for_each which doesn't support early return.

xiaoxq:


xiaoxq:
It's a partial revert to https://github.com/ApolloAuto/apollo/commit/2c1a147aa0b938d884a7cd1c60621b48f2222c7e

PR #10269, opened 2019-11-20
Prediction: add lock in base image drawing

kechxu:


modules/prediction/common/semantic_map.cc
```diff
@@ -52,27 +51,20 @@ void SemanticMap::RunCurrFrame(
...
     double x = ego_feature_.position().x();
     double y = ego_feature_.position().y();
     curr_base_x_ = x - FLAGS_base_image_half_range;
     curr_base_y_ = y - FLAGS_base_image_half_range;
     DrawBaseMap(x, y, curr_base_x_, curr_base_y_);
-    base_img_.copyTo(curr_base_img_);
+    base_img_.copyTo(curr_img_);
   } else {
-    base_img_.copyTo(curr_base_img_);
+    base_img_.copyTo(curr_img_);
```
xiaoxq: Move it out of `if, else` if it's executed anyway.

PR #10272, opened 2019-11-21
planning: fix a bug a virtual stop fence is created without STOP decision when destination point it not along reference line

jmtao:


modules/planning/tasks/deciders/rule_based_stop_decider/rule_based_stop_decider.h
```diff
@@ -65,6 +65,7 @@ class RuleBasedStopDecider : public Decider {
                       const common::PathPoint& stop_point);
 
  private:
+  static constexpr char const* const PATH_END_STOP_VO_ID = "PATH_END_STOP";
```
xiaoxq: Put it to `.cc`.

PR #10273, opened 2019-11-21
Control: refactor MRAC control algorithm with initial gains estimation

kevin-y-wang:


modules/control/common/mrac_controller.cc
```diff
@@ -262,6 +265,45 @@ bool MracController::CheckLyapunovPD(const Matrix matrix_a,
...
+    matrix_a_estimate(0, 1) = 1.0;
+    matrix_a_estimate(1, 0) = -wn_estimate * wn_estimate;
+    matrix_a_estimate(1, 1) = -2 * zeta_estimate * wn_estimate;
+    matrix_b_estimate(1, 0) = wn_estimate * wn_estimate;
+  } else {
+    AWARN << "No pre-known actuation dynamics; the initial states of the "
+             "adaptive gains are set as zeros";
+    return;
+  }
+  gain_state_adaption_.col(1) =
```
Capri2014: You are just calculating psedo inverse here, so try to re-use the function in `modules/common` https://github.com/ApolloAuto/apollo/blob/9e3c56a7342800a5e06aac46f9acf905071f03bc/modules/common/math/matrix_operations.h#L51

kevin-y-wang: re-use the common/math function per comment. Thanks.

PR #10307, opened 2019-11-23
Control: Mrac Debug message into Control Proto 

kevin-y-wang:


modules/control/proto/control_cmd.proto
```diff
@@ -222,6 +225,21 @@ message SimpleMPCDebug {
   optional double jerk_error = 49;
 }
 
+message SteerMracDebug {
```
Capri2014: Name it MARC debug to be more generic instead of `SteerMracDebug`

PR #10312, opened 2019-11-24
Prediction: refactor some logic of caution

kechxu:


modules/prediction/predictor/extrapolation/extrapolation_predictor.cc
```diff
@@ -211,14 +211,15 @@ double ExtrapolationPredictor::ComputeExtraplationSpeed(
...
-  double diff_y = last_point.path_point().y() - mid_point.path_point().y();
-  return std::hypot(diff_x, diff_y) / (num_tail_point * time_resolution);
+  CHECK_GT(num_tail_point, 0);
+  double v_sum = 0.0;
+  int v_count = 0;
+  int mid_index = num_trajectory_point - num_tail_point;
+  for (int i = mid_index; i < num_trajectory_point; ++i) {
+    v_sum += trajectory.trajectory_point(i).v();
+    ++v_count;
+  }
```
xiaoxq: You mean `v_count = num_tail_point`?

kechxu: Yes, you are right. I have fixed it.

Issue #10284, opened 2019-11-22
Old tf2 library Is Buggy


Fla3inH0tCheet0s:
It seems like the source of the tf2 library is old. Are there plans to rewrite or rebase it?

Take a look at this function https://github.com/ApolloAuto/apollo/blob/master/third_party/tf2/src/buffer_core.cpp#L479. It seems like there's an infinite loop, if `(*frame_chain)[n]` equals `reverse_frame_chain[m]` at 0 or if either size is 0.

```
unsigned int m = reverse_frame_chain.size() - 1;
unsigned int n = frame_chain->size() - 1;
for (; m >= 0 && n >= 0; --m, --n) {
  if ((*frame_chain)[n] != reverse_frame_chain[m]) break;
}
```

Newer version seem to address this issue: https://github.com/ros2/geometry2/blob/ros2/tf2/src/buffer_core.cpp#L531. I'm not sure what other issues they addressed.
```
int m = (int)reverse_frame_chain.size()-1;
int n = (int)frame_chain->size()-1;
for (; m >= 0 && n >= 0; --m, --n)
{
  if ((*frame_chain)[n] != reverse_frame_chain[m])
    break;
}
```

mickeyouyou:
@Fla3inH0tCheet0s thank you, this issue have fixed, we will keep our changes in the patch https://github.com/ApolloAuto/apollo/blob/master/third_party/tf2/geometry2_0.5.16_tf2.patch

thank you so much again.

PR #10318, opened 2019-11-25
ThirdParty: fix issue #10284

mickeyouyou:


third_party/tf2/src/buffer_core.cpp
```diff
@@ -476,8 +476,8 @@ int BufferCore::walkToTopParent(
   if (frame_chain) {
     // Pruning: Compare the chains starting at the parent (end) until they
     // differ
-    unsigned int m = reverse_frame_chain.size() - 1;
-    unsigned int n = frame_chain->size() - 1;
+    int m = (int)reverse_frame_chain.size() - 1;
+    int n = (int)frame_chain->size() - 1;
```
xiaoxq: Use `static_cast`.

PR #10340, opened 2019-11-27
Control: added header to submodules

sjiang2018:


modules/control/submodules/preprocessor_submodule.cc
```diff
@@ -84,8 +84,9 @@ bool PreprocessorSubmodule::Proc(const std::shared_ptr<LocalView> &local_view) {
     control_preprocessor.set_received_pad_msg(true);
   }
 
-  preprocessor_writer_->Write(
-      std::make_shared<Preprocessor>(control_preprocessor));
+  common::util::FillHeader(Name(), &control_preprocessor);
```
michael4338: do we need to populate the lidar_timestamp field?

PR #10346, opened 2019-11-27
Control: fix return status for lat_lon_controller_submodule

Capri2014:


Capri2014:
More return status needs to be fixed for other submodules.

PR #10349, opened 2019-11-27
Control: fix return status for mpc controller submodule

Capri2014:


modules/control/submodules/postprocessor_submodule.cc
```diff
@@ -41,51 +41,52 @@ bool PostprocessorSubmodule::Init() {
       << FLAGS_control_common_conf_file;
 
   postprocessor_writer_ =
-      node_->CreateWriter<ControlCommand>(FLAGS_control_command_topic);
+      node_->CreateWriter<ControlCommand>(FLAGS_control_core_command_topic);
```
sjiang2018: Control command

Capri2014: @sjiang2018 Hoops, you are right. fixed

PR #10352, opened 2019-11-27
paladin: add latency measurement for control

michael4338:


modules/common/latency_recorder/latency_recorder.cc
```diff
@@ -22,13 +22,29 @@
 namespace apollo {
 namespace common {
 
+namespace {
+
 using apollo::cyber::Time;
 
+uint64_t ToNanoSeconds(const double seconds) {
+  static constexpr uint64_t kNanoSecondsFactor = 1000000000UL;
+  return static_cast<uint64_t>(kNanoSecondsFactor * seconds);
```
xiaoxq: `static constexpr double kNanoSecondsFactor = 1e9;`
Otherwise, the uint64 will be converted to double first before multiplying `seconds`.

PR #10354, opened 2019-11-28
Map: hdmap_map_test runs in parallel

Fla3inH0tCheet0s:
## Description
It might be good to support the `--runs_per_test` feature for all tests. This change should make the test file unique and cleanup the file afterward.

## Test
```
bazel test --runs_per_test=50 //modules/map/hdmap:hdmap_map_test
```

### DIFF
--- a/modules/map/hdmap/hdmap_impl_test.cc
+++ b/modules/map/hdmap/hdmap_impl_test.cc
@@ -13,9 +13,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 =========================================================================*/
 
-#include "modules/map/hdmap/hdmap_impl.h"
-#include "cyber/common/file.h"
+#include <chrono>
+
+#include "absl/strings/str_cat.h"
 #include "gtest/gtest.h"
+
+#include "cyber/common/file.h"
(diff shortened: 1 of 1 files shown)


modules/map/hdmap/hdmap_impl_test.cc
```diff
@@ -414,13 +416,17 @@ TEST_F(HDMapImplTestSuite, GetLocalMap) {
   std::pair<double, double> range{10.0, 10.0};
   ASSERT_EQ(0, hdmap_impl_.GetLocalMap(point, range, &local_map));
 
-  const std::string output_bin_file = FLAGS_output_dir + "/base_map.bin";
+  const std::string time_since_epoch = std::to_string(
+      std::chrono::steady_clock::now().time_since_epoch().count());
+  const std::string output_bin_file =
+      FLAGS_output_dir + "/base_map_" + time_since_epoch + " .bin";
```
xiaoxq: Please use `absl::StrCat` to avoid many temporary intermediate string objects.
https://github.com/ApolloAuto/apollo/blob/master/modules/map/hdmap/hdmap_util.cc#L43

PR #10365, opened 2019-12-01
Robot: Weekly code clean.

xiaoxq:


modules/planning/open_space/trajectory_smoother/distance_approach_ipopt_fixed_dual_interface.cc
```diff
@@ -777,7 +777,6 @@ void DistanceApproachIPOPTFixedDualInterface::eval_constraints(int n,
   g[constraint_index + 2] = x[state_index + 2];
   g[constraint_index + 3] = x[state_index + 3];
   constraint_index += 4;
-  state_index += 4;
```
Capri2014: It might be better to leave it here to make it more easier to debug. In case equation matrix needs to be modified for re-formulated later.

modules/planning/tasks/optimizers/piecewise_jerk_speed/piecewise_jerk_speed_nonlinear_ipopt_interface.cc
```diff
@@ -496,8 +496,6 @@ bool PiecewiseJerkSpeedNonlinearIpoptInterface::eval_g(int n, const double *x,
       g[coffset_soft_upper_s + i] =
           s - soft_safety_bounds_[i].second - upper_s_slack;
     }
-
-    offset += num_of_points_;
```
Capri2014: Same here

PR #10374, opened 2019-12-03
Planning: relax following gap when adc is slow

JasonZhou404:


modules/planning/tasks/optimizers/piecewise_jerk_speed/piecewise_jerk_speed_nonlinear_optimizer.cc
```diff
@@ -74,7 +75,8 @@ Status PiecewiseJerkSpeedNonlinearOptimizer::Process(
     return Status::OK();
   }
 
-  const auto& problem_setups_status = SetUpStatesAndBounds(path_data);
+  const auto& problem_setups_status =
+      SetUpStatesAndBounds(path_data, *speed_data);
```
xiaoxq: `SetUpStatesAndBounds` returns a copy, so you need to hold it instead of reference.

PR #10388, opened 2019-12-04
Control: clean up error status logic

sjiang2018:


modules/control/submodules/preprocessor_submodule.cc
```diff
@@ -98,6 +99,17 @@ bool PreprocessorSubmodule::Proc(const std::shared_ptr<LocalView> &local_view) {
   control_preprocessor.mutable_header()->set_radar_timestamp(
       local_view->trajectory().header().radar_timestamp());
   common::util::FillHeader(Name(), &control_preprocessor);
+  if (control_preprocessor.local_view().has_pad_msg()) {
+    auto pad_message = control_preprocessor.local_view().pad_msg();
```
xiaoxq: const auto&

sjiang2018: Fixed. Thank you.

PR #10428, opened 2019-12-06
Drivers: add latency log for lidar

fengqikai1414:


modules/drivers/velodyne/compensator/compensator_component.cc
```diff
@@ -62,10 +62,14 @@ bool CompensatorComponent::Proc(
   }
   point_cloud_compensated->Clear();
   if (compensator_->MotionCompensation(point_cloud, point_cloud_compensated)) {
-    uint64_t diff = cyber::Time().Now().ToNanosecond() - start;
-    AINFO << "compenstator diff:" << diff
+    auto now = cyber::Time().Now().ToNanosecond();
```
Capri2014: use `const` here?

modules/drivers/velodyne/fusion/pri_sec_fusion_component.cc
```diff
@@ -64,6 +64,9 @@ bool PriSecFusionComponent::Proc(
     }
     std::this_thread::sleep_for(std::chrono::milliseconds(5));
   }
+  auto diff = Time::Now().ToNanosecond() -
+              static_cast<uint64_t>(target->header().lidar_timestamp());
```
xiaoxq: Seems like that it is `uint64_t`. So the `static_cast` is not needed.

fengqikai1414: Yes, you are right. google::protobuf::uint64 is an alias for uint64_t. Have fixed.

modules/drivers/velodyne/compensator/compensator_component.cc
fengqikai1414: Fixed.

PR #10429, opened 2019-12-06
All: Thinner targets for compilation optimization.

xiaoxq:


xiaoxq:
@vlin17 @unacao This change will require the recent pull-request to simulation. But it's safe for both old and new code after deployed.

unacao:
> @vlin17 @unacao This change will require the recent pull-request to simulation. But it's safe for both old and new code after deployed.

which pull request to simulation are you referring to?

xiaoxq:
> which pull request to simulation are you referring to?

608, which has been merged.

PR #10430, opened 2019-12-06
Prediction: use junction_scenario info from storytelling_container

HongyiSun:


modules/prediction/container/storytelling/storytelling_container.cc
```diff
@@ -32,6 +32,10 @@ std::shared_ptr<const hdmap::JunctionInfo> StoryTellingContainer::ADCJunction()
   return PredictionMap::JunctionById(adc_junction_id);
 }
 
+std::string StoryTellingContainer::ADCJunctionId() const {
+  return close_to_junction_.junction_id();
```
xiaoxq: How about returning a const reference?

PR #10431, opened 2019-12-06
planning: added configurations for  st bound decider

yifeijiang:


modules/planning/conf/scenario/park_and_go_config.pb.txt
```diff
@@ -132,4 +133,7 @@ stage_config:{
     task_config: {
         task_type: RULE_BASED_STOP_DECIDER
     }
+    task_config: {
+        task_type: ST_BOUNDS_DECIDER
```
xiaoxq: Seems like all other config files have indents as 2.

modules/planning/conf/scenario/valet_parking_config.pb.txt
```diff
@@ -64,6 +65,9 @@ stage_config: {
     task_config: {
         task_type: RULE_BASED_STOP_DECIDER
     }
+    task_config: {
+        task_type: ST_BOUNDS_DECIDER
```
xiaoxq: ditto

PR #10438, opened 2019-12-07
Prediction: avoid deep copy of perception in lego

kechxu:


modules/prediction/prediction_component.cc
```diff
@@ -107,7 +110,7 @@ bool PredictionComponent::Proc(
 
 bool PredictionComponent::ContainerSubmoduleProcess(
     const std::shared_ptr<PerceptionObstacles>& perception_obstacles) {
-  const double frame_start_time = Clock::NowInSeconds();
+  const double frame_start_time = apollo::cyber::Time::Now().ToNanosecond();
```
xiaoxq: 1. ToNanosecond() returns uint64.
2. Try `absl::ToUnixNanos(absl::Now())` if you need a system time.

modules/prediction/submodules/predictor_submodule.cc
```diff
@@ -73,6 +76,10 @@ bool PredictorSubmodule::Proc(
   predictor_writer_->Write(
       std::make_shared<PredictionObstacles>(prediction_obstacles));
 
+  double end_time = apollo::cyber::Time::Now().ToNanosecond();
+  double start_time = submodule_output->frame_start_time();
+  ADEBUG << "End to end time = " << (end_time - start_time) / 1e6 << " ms";
```
xiaoxq: Use high level timer API instead of hard-code the calculation, with magic numbers like `1e6`, `1e9`.
```
const auto start_time = absl::Now();
...
const auto end_time = absl::Now();
ADEBUG << "End to end time = " << absl::ToDoubleMilliseconds(end_time - start_time) << " ms";
```

PR #10452, opened 2019-12-09
Prediction: add OnStoryTelling and use it for junction info

HongyiSun:


modules/prediction/prediction_component.cc
```diff
@@ -133,6 +133,16 @@ bool PredictionComponent::ContainerSubmoduleProcess(
...
     MessageProcess::OnPlanning(trajectory_msg);
   }
+
+  // Read storytelling message and call OnStorytelling to update the
+  // StoryTellingContainer
+  storytelling_reader_->Observe();
+  auto ptr_storytelling_msg = storytelling_reader_->GetLatestObserved();
+  if (ptr_storytelling_msg != nullptr) {
+    auto storytelling_msg = *ptr_storytelling_msg;
+    MessageProcess::OnStoryTelling(storytelling_msg);
```
xiaoxq: So you make an unnecessary copy here. Use `const auto&`, or just call `OnStoryTelling(*ptr_storytelling_msg)`.

modules/prediction/prediction_component.cc
```diff
@@ -152,8 +162,8 @@ bool PredictionComponent::ContainerSubmoduleProcess(
   ADCTrajectoryContainer adc_container = *adc_trajectory_container_ptr;
   adc_container_writer_->Write(
       std::make_shared<ADCTrajectoryContainer>(adc_container));
-  perception_obstacles_writer_->Write(std::make_shared<PerceptionObstacles>(
-      perception_msg));
+  perception_obstacles_writer_->Write(
+      std::make_shared<PerceptionObstacles>(perception_msg));
```
xiaoxq: `write(perception_msg)` works the same.

PR #10474, opened 2019-12-11
Revert "Docker: fix getting openGL version failed when using cyber_vi…

xiaoxq:
…sualizer (#10426)"

This reverts commit c8f429f89fa1d609904edc12178bd4ed28e65e48.

### DIFF
--- a/docker/scripts/dev_start.sh
+++ b/docker/scripts/dev_start.sh
@@ -349,12 +349,6 @@ function main(){
 
     set -x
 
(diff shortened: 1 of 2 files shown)


xiaoxq:
This change makes assumption of HOST which may not be true. And if both host and container have "/usr/lib/x86_64-linux-gnu", there will be conflict or even weird behavior. @muleisheng is working on a solid fix which is self-contained within the container.

PR #10483, opened 2019-12-12
Planning: st-decider

panjiacheng:


modules/planning/tasks/deciders/st_bounds_decider/st_obstacles_processor.cc
```diff
@@ -188,20 +208,40 @@ Status STObstaclesProcessor::MapObstaclesToSTBoundaries(
...
-          closest_stop_obs_boundary.min_s() <=
-              std::get<1>(clear_zone).max_s()) {
-        std::tie(closest_stop_obs_id, closest_stop_obs_boundary,
-                 closest_stop_obs_ptr) = clear_zone;
-        ADEBUG << "Clear zone " << closest_stop_obs_id << " is closer.";
-        break;
+    if (closest_stop_obs_id.find("Side_Pass_Stop") == std::string::npos &&
+        closest_stop_obs_id.find("DEST") == std::string::npos) {
+      for (const auto& clear_zone : candidate_clear_zones_) {
+        auto clear_zone_boundary = std::get<1>(clear_zone);
```
xiaoxq: As the returned value is an object, plz keep it in a const reference to avoid making copy.
If it's scalar value, we'll make a copy as it's cheap.

panjiacheng: fixed :)

PR #10490, opened 2019-12-12
Planning: skip path_reuse when not in lane_follow scenario.

sjiang2018:


modules/planning/tasks/deciders/path_reuse_decider/path_reuse_decider.cc
```diff
@@ -44,6 +44,20 @@ Status PathReuseDecider::Process(Frame* const frame,
   CHECK_NOTNULL(frame);
   CHECK_NOTNULL(reference_line_info);
 
+  auto* mutable_path_reuse_decider_status = PlanningContext::Instance()
+                                                ->mutable_planning_status()
+                                                ->mutable_path_reuse_decider();
+
+  // skip path reuse if not in LANE_FOLLOW_SCENARIO
+  const ScenarioConfig_ScenarioType& scenario_type =
```
xiaoxq: Make a local copy for scalar types.

PR #10517, opened 2019-12-16
Control: reset the initial MRAC adaptive gains if no given vehicle_param latency

kevin-y-wang:


modules/control/common/mrac_controller.cc
```diff
@@ -454,6 +454,30 @@ void MracController::SetInitialCommand(const double command_init) {
   input_desired_(0, 1) = command_init;
 }
 
+void MracController::SetInitialStateAdaptionGain(
+    const Matrix gain_state_adaption_init) {
```
xiaoxq: Matrix&

modules/control/common/mrac_controller.h
```diff
@@ -187,6 +187,26 @@ class MracController {
    */
   void SetInitialCommand(const double command_init);
 
+  /**
+   * @brief set initial values of state adaption gains for mrac control
+   * @param gain_state_adaption_init initial state adaption gains
+   */
+  void SetInitialStateAdaptionGain(
+      const Eigen::MatrixXd gain_state_adaption_init);
```
xiaoxq: ditto

modules/control/common/mrac_controller.cc
kevin-y-wang: Fixed per comment. Thanks.

PR #10523, opened 2019-12-16
Planning: updated path_reuse logic

sjiang2018:


modules/planning/tasks/deciders/path_reuse_decider/BUILD
```diff
@@ -14,6 +14,7 @@ cc_library(
         "//modules/planning/common:planning_gflags",
         "//modules/planning/common:reference_line_info",
         "//modules/planning/tasks/deciders:decider_base",
+        "//modules/planning/proto:planning_proto",
```
xiaoxq: Sort.

modules/planning/tasks/deciders/path_reuse_decider/path_reuse_decider.cc
```diff
@@ -34,6 +35,7 @@ using apollo::common::math::Vec2d;
 
 int PathReuseDecider::reusable_path_counter_ = 0;
 int PathReuseDecider::total_path_counter_ = 0;
+bool PathReuseDecider::path_reusable_ = false;
```
xiaoxq: You may init value inline in the header file.

modules/planning/tasks/deciders/path_reuse_decider/path_reuse_decider.cc
```diff
@@ -84,21 +86,32 @@ Status PathReuseDecider::Process(Frame* const frame,
...
+  // 3. failed to trim previous path
+  // 4. speed optimization failed on previous path
+  // const auto history_trajectory_type = FrameHistory::Instance()
+  //                                          ->Latest()
+  //                                          ->reference_line_info()
+  //                                          .front()
+  //                                          .trajectory_type();
+  if (path_reusable_) {
+    if (!frame->current_frame_planned_trajectory().is_replan() &&
+        CheckPathReusable(frame, reference_line_info) &&
```
Capri2014: replace the name `CheckPathReusable` to avoid confusion, it look like we are expecting false here, util I saw in implementation that you are just returning the `IsCollisionFree` results.

modules/planning/tasks/deciders/path_reuse_decider/path_reuse_decider.cc
sjiang2018: Can you please give me an example?

xiaoxq: Oops, not work for non-const static member.

PR #10578, opened 2019-12-25
clean & clear code

storypku:
clean and clear code


storypku:
How to restart the teamcity build check when "git early eof" issues was encountered?

PR #10580, opened 2019-12-26
canbus: added option to select socket CAN interface (native, virtual or slcan)

### LINKED ISSUE #10546: Add vcan support for socket_can_client_raw
Adding an option for 'vcan' (virtual socket can) would be very beneficial for running the system in our simulation. We currently use this and would be a great public feature. 

The update would only require changing "can" to "vcan" on line 100 shown in figure below.

image

This could either be set during the build or an additional socket client for virtual busses could be added to the component for this.

james-goin:
Added support for virtual and serial CAN interfaces when using the socket CAN client. Tested with our virtual CAN bus based drive by wire simulator. 

closes #10546

### DIFF
--- a/modules/canbus/conf/canbus_conf.pb.txt
+++ b/modules/canbus/conf/canbus_conf.pb.txt
@@ -8,6 +8,7 @@ can_card_parameter {
   brand: ESD_CAN
   type: PCI_CARD
   channel_id: CHANNEL_ID_ZERO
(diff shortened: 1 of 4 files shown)


Capri2014:
@james-goin Thanks for the contribution, please fix the lint issue. You can do a self-check via `apollo.sh lint`

PR #10623, opened 2020-01-14
scripts: add -n option to skip pull docker image

zhxt:
Provide a way to skip docker pulling procedure if we know exactly those images are up to date.

Call for test!

And any better option name than '-n' ?


xiaoxq:
The `--local` option is for using local image instead of pulling from remote. Could you try if it works for you?

zhxt:
@xiaoxq yeah, I know that, --local is added at the time when we only have one docker image(dev-xxx),
it just skips the main image, we have five images/volumes now, the later added images/volumes do not follow the rule.

Well, I am thinking if we can merge those two options into one.

PR #10627, opened 2020-01-14
common: python 2 to 3 for modules/tools

michael4338:


modules/tools/mobileye_viewer/location_server.py
```diff
@@ -19,7 +19,7 @@
 import os
 import rospy
 import math
-import thread
+import _thread
```
xiaoxq: The next step is to fix the header orders to satisfy coding style requirement. Maybe with tools like autopep8.

PR #10635, opened 2020-01-16
[D-kit] Add config files for D-kit

SeasoulChris:
1、add config files for D-kit
2、add gflag for vehicle width 
3、add python tools for odometry generator

### DIFF
--- a/modules/calibration/data/dev_kit/Config_files/camera.dag
+++ b/modules/calibration/data/dev_kit/Config_files/camera.dag
@@ -1,0 +1,43 @@
+# Define all coms in DAG streaming.
+module_config {
+    module_library : "/apollo/bazel-bin/modules/drivers/camera/libcamera_component.so"
+
+    components {
+      class_name : "CameraComponent"
+      config {
+        name : "camera_front_6mm"
+        config_file_path : "/apollo/modules/drivers/camera/conf/camera_front_6mm.pb.txt"
+      }
+    }
+    components {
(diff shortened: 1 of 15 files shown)


modules/drivers/velodyne/launch/velodyne16.launch
```diff
@@ -0,0 +1,7 @@
+<cyber>
+    <module>
+        <name>gnss</name>
```
zhxt: Module name  is "gnss"?

modules/calibration/data/dev_kit/Config_files/camera.dag
```diff
@@ -0,0 +1,43 @@
+# Define all coms in DAG streaming.
```
zhxt: The dir name "Config_files" makes people a little bit confused since nearly all files under dev_kit is config files :P

Looking at the existing directory names, you may find some thing interesting to follow.

modules/drivers/velodyne/launch/velodyne16.launch
SeasoulChris: I referred to modules/drivers/velodyne/launch/velodyne128.launch , so velodyne128.launch should be corrected too?
 ```
<cyber>
    <module>
        <name>gnss</name>
        <dag_conf>/apollo/modules/drivers/velodyne/dag/velodyne128.dag</dag_conf>
        <process_name>gnss</process_name>
    </module>
</cyber>

```

PR #10649, opened 2020-01-20
python: switch to python3 for applications

michael4338:


docker/build/installers/py3_requirements.txt
```diff
@@ -1,6 +1,32 @@
+# system utils
+supervisor
+
 # Google infras.
 absl-py
+glog
+grpcio-tools == 1.2.0
```
Capri2014: Could you remind me why we need grpcio-tools with a fixed version of 1.2.0... I remember I introduced this one previously due to Bazel/Protobuf/Grpc version issue but not sure if this is the only place needs a `grpcio-tools == 1.2.0`

docker/build/installers/install_python_modules.sh
```diff
@@ -23,13 +23,15 @@ cd "$(dirname "${BASH_SOURCE[0]}")"
...
     python-pip \
     python-psutil \
     python-scipy \
-    software-properties-common \
+    python-matplotlib \
     python3-pip \
-    python3-psutil
+    python3-psutil \
+    python3-scipy \
+    python3-matplotlib
```
xiaoxq: sort

docker/build/installers/py3_requirements.txt
michael4338: confirmed this one is not needed anymore, so will just remove it

PR #10653, opened 2020-01-21
[Drivers/Velodyne]add velodyne16.dag velodyne16.launch

SeasoulChris:
add veldoyne16.dag 、velodyne16.launch


modules/drivers/velodyne/launch/velodyne16.launch
```diff
@@ -0,0 +1,7 @@
+<cyber>
+    <module>
+        <name>gnss</name>
```
Capri2014: The process name shall not be `gnss`

PR #10657, opened 2020-01-22
Planning&Scripts: add ntp pool options and enrich planning information

Capri2014:


scripts/time_sync.sh
```diff
@@ -20,10 +20,23 @@ sudo apt-get install -y ntpdate
...
 
+REGIN="us"
+
+if [ "$#" -ne 1 ]
+then
+      echo "Usage:      ./time_sync.sh <ntp_pool_regin>"
+      echo "Example:    ./time_sync.sh cn                               # Set ntp pool to China server"
+      echo "Default using us as ntp pool regin"
+else
+      REGIN=$1
```
xiaoxq: Fix indents.

modules/planning/on_lane_planning.cc
```diff
@@ -211,7 +211,9 @@ void OnLanePlanning::RunOnce(const LocalView& local_view,
   VehicleState vehicle_state =
       VehicleStateProvider::Instance()->vehicle_state();
   const double vehicle_state_timestamp = vehicle_state.timestamp();
-  DCHECK_GE(start_timestamp, vehicle_state_timestamp);
+  DCHECK_GE(start_timestamp, vehicle_state_timestamp)
+      << "start_time_stamp is behind vehicle_state_timestamp by"
+      << start_time_stamp - vehicle_state_timestamp << "secs";
```
xiaoxq: Leave some space before and after numbers.

scripts/time_sync.sh
Capri2014: Done

modules/planning/on_lane_planning.cc
Capri2014: Done

PR #10669, opened 2020-01-23
build_py: build grpc interface

vlin17:


apollo.sh
```diff
@@ -282,6 +282,9 @@ function build_py_proto() {
   find modules/ cyber/ -name "*.proto" \
       | grep -v node_modules \
       | xargs protoc --python_out=py_proto
+  find modules/ cyber/ -name "*_service.proto" \
+      | grep -v node_modules \
```
Capri2014: @vlin17 @michael4338 Do we need to  added/updated   the grpc_tool version in docker before this one can be merged in?

michael4338: will add the library into python3 requirements soon, but it should not block this change I think.  We have python2 in current docker image already.

PR #10684, opened 2020-01-29
Control: using system time for time latency

sjiang2018:


modules/control/control_component.cc
```diff
@@ -394,8 +398,18 @@ bool ControlComponent::Proc() {
   }
 
   const auto end_time = Clock::Now();
+  const auto end_system_timestamp =
+      std::chrono::duration<double>(
+          std::chrono::system_clock::now().time_since_epoch())
+          .count();
+  double time_diff_ms;
+  if (FLAGS_use_system_time) {
+    time_diff_ms = (end_system_timestamp - start_system_timestamp) * 1000;
```
xiaoxq: 1. Move the `end_system_timestamp` var into the if scope. [Minimum scope rule]
2. The flag name is too common. Note that a FLAG is a global var. Rename it to things like "use_system_time_in_control"
3. If I understand correctly, control should always use system time. If that is the case, just replace `Clock::Now()` to `absl::Now()` everywhere.

modules/control/control_component.cc
```diff
@@ -277,6 +277,10 @@ Status ControlComponent::ProduceControlCommand(
 
 bool ControlComponent::Proc() {
   const auto start_time = Clock::Now();
+  const double start_system_timestamp =
```
kevin-y-wang: suggest "const auto" to keep consistent with the others

modules/control/control_component.cc
sjiang2018: Fixed. Thank you.

modules/control/control_component.cc
sjiang2018: Fixed. Thank you.

PR #10742, opened 2020-02-19
planning: remove unnecessary dependencies on prediction method

jmtao:


modules/planning/common/util/util.cc
```diff
@@ -128,6 +128,28 @@ bool CheckInsidePnCJunction(const ReferenceLineInfo& reference_line_info) {
...
+    return;
+  }
+  if (boost::filesystem::is_directory(path)) {
+    for (auto& entry : boost::make_iterator_range(
+             boost::filesystem::directory_iterator(path), {})) {
+      GetFilesByPath(entry.path(), files);
+    }
+  }
+}
+
```
xiaoxq: How about:
https://github.com/ApolloAuto/apollo-fuel/blob/v2/apps/k8s/spark_submitter/client.py#L70
https://github.com/ApolloAuto/apollo/blob/master/cyber/common/file.h#L140

PR #10820, opened 2020-03-09
Dreamview: replacing parking_space_id with parking_info for POI

vlin17:


modules/routing/proto/poi.proto
```diff
@@ -2,12 +2,14 @@ syntax = "proto2";
 
 package apollo.routing;
 
+import "modules/common/proto/geometry.proto";
```
unacao: is this needed?

modules/routing/proto/BUILD
```diff
@@ -29,6 +29,7 @@ py_proto(
     name = "poi_pb2",
     src = "poi.proto",
     deps = [
+        "//modules/common/proto:geometry_pb2",
```
unacao: is this needed?

vlin17: good catch.

PR #10835, opened 2020-03-12
routing: implemented get parking id from parking point

sjiang2018:


modules/routing/routing.cc
```diff
@@ -144,6 +146,55 @@ double Routing::GetRoutingLength(const RoutingResponse& routing_response) {
   return length;
 }
 
+bool Routing::GetParkingID(const PointENU& parking_point,
+                           std::string* parking_space_id) {
+  // search current parking space id associated with parking point.
+  constexpr double kDistance = 0.01;  // meter
+  std::vector<ParkingSpaceInfoConstPtr> parking_spaces;
+  if (!hdmap_->GetParkingSpaces(parking_point, kDistance, &parking_spaces)) {
```
jmtao: 条件写反了？

modules/routing/routing.cc
```diff
@@ -144,6 +146,55 @@ double Routing::GetRoutingLength(const RoutingResponse& routing_response) {
...
+    return false;
+  }
+
+  const bool has_parking_point =
+      has_parking_info && routing_request.parking_info().has_parking_point();
+  // not need to extract parking ID when parking_point is not available
+  if (!has_parking_point) {
+    ADEBUG << "routing has no parking point info";
+    return false;
+  }
```
jmtao: consolidate the above 3 checks into one. 
make it less length.

modules/routing/routing.cc
sjiang2018: https://github.com/ApolloAuto/apollo/blob/654f16e4e32e2ea863402fa248748c5fb7a2b80f/modules/map/hdmap/hdmap_impl.cc#L456
Here. It returns 0 when gets parking space successfully.

modules/routing/routing.cc
sjiang2018: Fixed. Thank you.

modules/routing/routing.cc
```diff
@@ -144,6 +146,41 @@ double Routing::GetRoutingLength(const RoutingResponse& routing_response) {
...
+}
+
+bool Routing::FillParkingID(RoutingResponse* routing_response) {
+  // set parking space ID when
+  //  has parking info && has parking point && NOT has parking space id && get
+  //  ID successfully
+  const auto& routing_request = routing_response->routing_request();
+  const bool has_parking_info = routing_request.has_parking_info();
+  if (!has_parking_info ||
+      routing_request.parking_info().has_parking_space_id() ||
```
jmtao: if (has_parking_space_id) , shall return true

unacao: I think she means no need to fill parking space id anymore if already provided.

vlin17: IMHO, returning true here can indicate that true always means routing response has parking id filled. Might be better from caller's perspective; otherwise, caller needs to check has_parking_space_id again.

sjiang2018: @unacao Yes. That is what I mean. Thank you for the explanation.
@jmtao @vlin17 In that case, I think the previous "three if's" version is better. if there is no parking info, there is no need to move on check parking id or parking point. So it returns false. And if there is parking id, it returns true. And then check parking_point ... I think it's the best way to organize the logic but I am open for any comments or suggestions anyway.

jmtao: the caller only cares about if "parking_space_id" has been filled up or not.  they don't have to know the logic inside, when and how you process this id. 
true means YES (=parking_space_id is in place fine). 
if you return FALSE, what do you expect the caller to do? unless the caller cares about if the parking_space_id was originally there, or filled by this method indeed.  I don't think that's the case here.

PR #10839, opened 2020-03-13
Dreamview: closes cleanly

Fla3inH0tCheet0s:
## Description
Dreamview sefaults after SIGINT. There are two issues.
1. `~Dreamview()` is called after `apollo::cyber::Clear()`. This causes a segfault because AINFO is called, but the logger is shutdown.
2. The SimulationWorldUpdater timer callback can be run while it is shutting down. This means that it can access out of scope data.

## Test
1. `bazel-bin/modules/dreamview/dreamview`
2. Wait 5 seconds
3. Press ctrl-c
4. Assert that the program returns within 5 seconds
5. `echo $?`
6. Assert that the exit code is 0

### DIFF
--- a/modules/dreamview/backend/main.cc
+++ b/modules/dreamview/backend/main.cc
@@ -31,7 +31,5 @@ int main(int argc, char *argv[]) {
     return -1;
   }
   apollo::cyber::WaitForShutdown();
-  dreamview.Stop();
-  apollo::cyber::Clear();
   return 0;
 }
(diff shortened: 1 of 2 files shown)


unacao:
Does executing dreamview with `$ ./scripts/dreamview.sh start` and `$ ./scripts/dreamview.sh stop` give you the same segfault?

Fla3inH0tCheet0s:
> Does executing dreamview with `$ ./scripts/dreamview.sh start` and `$ ./scripts/dreamview.sh stop` give you the same segfault?

Yes.

You can run this procedure before and after thee changes.
```
ls data/core | wc -w
# Note number of core dumps
bazel build -c opt //modules/dreamview
scripts/dreamview.sh
# Wait 5 seconds for program to start
scripts/dreamview.sh stop
# Wait 5 seconds for program to stop
ls data/core | wc -w
# Compare number of core dumps from before
```

PR #10864, opened 2020-03-23
Bazel2.x csh

changsh726:
Bazel:command "bazel build //modules/planning/..." passed


jinghaomiao:
@changsh726 please check teamcity and see why the build failed. Thanks!

PR #10947, opened 2020-04-04
RecordWriter: cleanup

storypku:


cyber/record/record_writer.cc
```diff
@@ -177,17 +177,15 @@ bool RecordWriter::IsNewChannel(const std::string& channel_name) const {
 void RecordWriter::OnNewChannel(const std::string& channel_name,
                                 const std::string& message_type,
                                 const std::string& proto_desc) {
-  if (IsNewChannel(channel_name)) {
-    channel_message_number_map_[channel_name] = 0;
-    channel_message_type_map_[channel_name] = message_type;
-    channel_proto_desc_map_[channel_name] = proto_desc;
-  }
+  channel_message_number_map_[channel_name] = 0;
```
jinghaomiao, line 180: Better check to save some computation time.

storypku:
Since the OnNewChannel member func is private,  maybe class author(s)
should ensure that it is called after some check?

Calvin Miao <notifications@github.com> 于2020年4月7日周二 上午2:17写道：

cyber/record/record_writer.cc
storypku: Since the OnNewChannel member func is private,  maybe class author(s) should ensure that it is called after some check?

storypku: > Better check to save some computation time.

Since the OnNewChannel member func is private,  maybe class author(s) should ensure that it is called after some check?

jinghaomiao: I get your point. Please do a cross check to see when this function is called and if it's called in the desired manner.

storypku: > I get your point. Please do a cross check to see when this function is called and if it's called in the desired manner.

Yep, already checked  before making this change.

PR #10973, opened 2020-04-11
R5.5.0 vehicle_controller.cc

Gavin-Tao:
if (EnableSteeringOnlyMode() != ErrorCode::OK) {
        AERROR << "Failed to enable steering only mode."; 
        return ErrorCode::CANBUS_ERROR;


Gavin-Tao:
speed should be corrected by steering

PR #10989, opened 2020-04-12
Update ch_controller.cc

Gavin-Tao:
Forget to output "Protocol can sender is null".


jinghaomiao:
@Gavin-Tao I am closing all your commits. As communicated earlier, please check into master head and we'll do the cherry picking to 5.5 branch. And if possible, please combine your commits into a single or fewer pull requests. Thank you again for your contribution.

PR #11025, opened 2020-04-14
Smartereye

lishuanghua:
This is a hard device driver which is added to apollo as a third party perception device.


modules/drivers/smartereye/compress_component.cc
```diff
@@ -0,0 +1,84 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
jinghaomiao, line 2: 2020 and so on and so forth.

jinghaomiao:
@lishuanghua please split your pull requests with multiple commits for easy code review and management purpose.

lx18233184051:
[06:01:55][Step 2/6] //modules/drivers/smartereye:smartereye_component_lib_cpplint            FAILED 
[06:01:55][Step 2/6] //modules/drivers/smartereye:smartereye_device_cpplint                   FAILED in 0.3s
[06:01:55][Step 2/6] //modules/drivers/smartereye:smartereye_handler_cpplint                  FAILED in 0.4s

Capri2014:
Also looks like they are depending on external libs. (*.so)? Please correct me if I am wrong. If so, we generally prefer  to host  those as  third party libs and manage them with bazel or third party libs installers, in stead of directly check those in. @lishuanghua

PR #11039, opened 2020-04-15
remove redundant options

storypku:
Remove two options from the `test` command as they can be inherited from the `build` command.


storypku:
Closed Temporarily as for later commit to bazel 2.x branches.

PR #11042, opened 2020-04-15
third party refactor

panxuechao:


modules/third_party_perception/proto/third_party_perception_component.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto2";
+
+package apollo.third_party_perception;
+
+message ThirdPartyPerceptionDevice {
+  optional string device_names = 1 [default = "ThirdPartyPerceptionSmartereye"];
```
Capri2014, line 6: It is probably better to use Enum here since you candidates are "ThirdPartyPerceptionSmartereye" and "ThirdPartyPerceptionMobileye", which are mutually exclusive.

modules/third_party_perception/tools/conversion_smartereye.cc
```diff
@@ -0,0 +1,273 @@
...
+namespace third_party_perception {
+namespace conversion_smartereye {
+
+using apollo::canbus::Chassis;
+using apollo::localization::LocalizationEstimate;
+using apollo::perception::PerceptionObstacle;
+using apollo::perception::PerceptionObstacles;
+using Point = apollo::common::Point3D;
+
+std::map<std::int32_t, apollo::hdmap::LaneBoundaryType_Type>
```
Capri2014, line 47: If the mapping are the same, we can probably share this one and the one in mobileye.

modules/third_party_perception/tools/conversion_smartereye.cc
```diff
@@ -0,0 +1,273 @@
...
+    const apollo::localization::LocalizationEstimate& localization,
+    const apollo::canbus::Chassis& chassis) {
+  PerceptionObstacles obstacles;
+  // retrieve position and velocity of the main vehicle from the localization
+  // position
+
+  obstacles.mutable_header()->CopyFrom(smartereye_obstacles.header());
+
+  // Fullfill lane_type information
+  std::int32_t mob_left_lane_type =
```
Capri2014, line 68: the world 'mob' was really the acronym for 'mobileye', so please rename it accordingly.

modules/third_party_perception/tools/conversion_smartereye.cc
```diff
@@ -0,0 +1,273 @@
...
+
+  for (int index = 0; index < smartereye_obstacles.num_obstacles() &&
+                      index < smartereye_obstacles.output_obstacles_size();
+       ++index) {
+    auto* mob = obstacles.add_perception_obstacle();
+    const auto& data_obstacle =
+        smartereye_obstacles.output_obstacles().at(index);
+    int mob_id = data_obstacle.trackid() + FLAGS_smartereye_id_offset;
+
+    double mob_x = data_obstacle.avgdistancez();
```
Capri2014, line 145: same for all 'mob' in this file.

modules/third_party_perception/tools/conversion_smartereye.cc
```diff
@@ -0,0 +1,273 @@
...
+      converted_vx = converted_speed * std::cos(mob->theta());
+      converted_vy = converted_speed * std::sin(mob->theta());
+    } else {
+      converted_x = data_obstacle.real3dcenterx() - mob_l / 2.0;
+      converted_y = mob_y;
+      converted_vx = mob_vel_x + chassis.speed_mps();
+      converted_vy = 0.0;
+      converted_z = data_obstacle.avgdistancez();
+
+      mob->set_theta(0.0);
```
Capri2014, line 228: why set theta to 0.0 here? this would be overwritten by line 244

Capri2014:
Also for the commit msgs we would usually follow the rule of `ModuleName: Context for detailed changes`  for higher tracing efficiency.

PR #11043, opened 2020-04-15
Smartereye

lishuanghua:
This is a hard device driver which is added to apollo as a third party perception device.

### DIFF
--- a/modules/drivers/smartereye/smartereye_component.cc
+++ b/modules/drivers/smartereye/smartereye_component.cc
@@ -1,0 +1,287 @@
+/******************************************************************************
+ * Copyright 2020 The Apollo Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+
+#include "modules/drivers/smartereye/smartereye_component.h"
+
+#include "modules/drivers/smartereye/smartereye_handler.h"
+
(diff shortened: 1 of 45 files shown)


scripts/smartereye.sh
```diff
@@ -0,0 +1,71 @@
...
+sudo ln -s libQt5Widgets.so.5.12.2 libQt5Widgets.so.5
+
+sudo cp /apollo/third_party/camera_library/smartereye/qtlib5.12.2/libQt5X11Extras.so.5.12.2 /usr/lib/x86_64-linux-gnu/
+sudo rm -f /usr/lib/x86_64-linux-gnu/libQt5X11Extras.so.5
+sudo ln -s libQt5X11Extras.so.5.12.2 libQt5X11Extras.so.5
+
+sudo cp /apollo/third_party/camera_library/smartereye/qtlib5.12.2/libQt5XcbQpa.so.5.12.2 /usr/lib/x86_64-linux-gnu/
+sudo rm -f /usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5
+sudo ln -s libQt5XcbQpa.so.5.12.2 libQt5XcbQpa.so.5
+
```
jinghaomiao, line 71: Please check with @ycool and @Capri2014 to see if this is an appropriate way to manage libraries.

Capri2014: No, we talked offline and I do not think this is a proper way of doing it, this might potentially affect other applications depending on qt also.

Capri2014:
@lishuanghua what is the license under  `third_party/camera_library/smartereye/include/`?

Capri2014:
license looks good now, but we should not directly commit `*.so` into the code.

modules/drivers/smartereye/launch/smartereye.lauch
```diff
@@ -0,0 +1,7 @@
+<cyber>
```
xiaoxq, line 1: Wrong file extension.

modules/drivers/smartereye/smartereye_component.cc
```diff
@@ -0,0 +1,288 @@
...
+ * limitations under the License.
+ *****************************************************************************/
+
+#include "modules/drivers/smartereye/smartereye_component.h"
+
+#include "modules/drivers/smartereye/smartereye_handler.h"
+
+#include "modules/common/adapters/adapter_gflags.h"
+
+
```
xiaoxq, line 23: Remove extra line.

PR #11121, opened 2020-04-24
planning: use different algorithm to find current lane and current ro…

jmtao:
…uting index for learning data

### DIFF
--- a/modules/planning/pipeline/feature_generator.cc
+++ b/modules/planning/pipeline/feature_generator.cc
@@ -264,38 +264,43 @@ void FeatureGenerator::OnRoutingResponse(
   }
 }
 
-apollo::hdmap::LaneInfoConstPtr FeatureGenerator::GetLane(
-    const apollo::common::PointENU& position,
-    int* routing_index) {
-  constexpr double kRadius = 0.1;
+int FeatureGenerator::GetADCCurrentRoutingIndex() {
+  static constexpr double kRadius = 4.0;
+  const auto& pose = localizations_.back().pose();
   std::vector<std::shared_ptr<const apollo::hdmap::LaneInfo>> lanes;
-  for (int i = 0; i < 10; ++i) {
+  apollo::hdmap::HDMapUtil::BaseMapPtr()->GetLanes(
+      pose.position(), kRadius, &lanes);
+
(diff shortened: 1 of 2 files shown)


modules/planning/pipeline/feature_generator.cc
```diff
@@ -491,6 +496,11 @@ void FeatureGenerator::GenerateRoutingFeature(
   }
 
   if (routing_index < 0) {
+    std::ostringstream msg;
+    msg << "no LOCAL routing. frame_num["
+        << learning_data_frame->frame_num() << "]";
```
xiaoxq, line 501: Try `StrCat`.

modules/planning/pipeline/feature_generator.cc
```diff
@@ -515,6 +525,13 @@ void FeatureGenerator::GenerateRoutingFeature(
     i++;
   }
 
+  if (local_routing_lane_ids.empty()) {
+    std::ostringstream msg;
+    msg << "no LOCAL routing. frame_num["
+        << learning_data_frame->frame_num() << "]";
```
xiaoxq, line 531: ditto

modules/planning/pipeline/feature_generator.cc
```diff
@@ -264,40 +264,45 @@ void FeatureGenerator::OnRoutingResponse(
   }
 }
 
-apollo::hdmap::LaneInfoConstPtr FeatureGenerator::GetLane(
-    const apollo::common::PointENU& position,
-    int* routing_index) {
-  constexpr double kRadius = 0.1;
+int FeatureGenerator::GetADCCurrentRoutingIndex() {
+  constexpr double kRadius = 4.0;
```
xiaoxq, line 268: static

modules/planning/pipeline/feature_generator.cc
jmtao: if performance is not a concern(well, it is not ,here), stream seems better here. It IS a logging output with  flexible text. 
please approve. :)

modules/planning/pipeline/feature_generator.cc
jmtao: fixed. amended. thx!

PR #11126, opened 2020-04-25
apollo_dev image build:  option for the local_dev tag

storypku:
add "-l/--local" option to `build_dev.sh` to facilitate local development.

### DIFF
--- a/docker/build/build_dev.sh
+++ b/docker/build/build_dev.sh
@@ -17,9 +17,11 @@
 ###############################################################################
 
 # Usage:
-#   ./build_dev.sh ./dev.x86_64.dockerfile
+#   ./build_dev.sh ./dev.x86_64.dockerfile [options]
+# Options:
+#   -l/--local   also tag the newly built image as apolloauto/apollo:local_dev
+#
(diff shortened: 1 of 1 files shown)


docker/scripts/dev_start.sh
```diff
@@ -21,8 +21,8 @@ FAST_BUILD_MODE="no"
 FAST_TEST_MODE="no"
 VERSION=""
 ARCH=$(uname -m)
-VERSION_X86_64="dev-18.04-x86_64-20200422_2150"
-VERSION_AARCH64="dev-aarch64-20170927_1111"
+VERSION_X86_64="dev-18.04-x86_64-latest"
+VERSION_AARCH64="dev-aarch64-latest"
```
xiaoxq, line 25: I'd prefer keeping exact tag, so that the image version and code version are well matched. Otherwise history versions of code would always break.

storypku: > I'd prefer keeping exact tag, so that the image version and code version are well matched. Otherwise history versions of code would always break.

Good point!

PR #11131, opened 2020-04-26
change dag proto to suit new replay engine for multi adc

lx18233184051:


cyber/proto/dag_conf.proto
```diff
@@ -7,11 +7,17 @@ import "cyber/proto/component_conf.proto";
 message ComponentInfo {
   optional string class_name = 1;
   optional ComponentConfig config = 2;
+  optional uint64 adc_index = 3;
+  optional string episode_name = 4;
+
```
xiaoxq, line 12: No additional line here. Try `./scripts/clang-format.sh` to do auto-format and learn from it :)

cyber/proto/dag_conf.proto
```diff
@@ -7,11 +7,17 @@ import "cyber/proto/component_conf.proto";
 message ComponentInfo {
   optional string class_name = 1;
   optional ComponentConfig config = 2;
+  optional uint64 adc_index = 3;
+  optional string episode_name = 4;
```
xiaoxq, line 11: Not sure if it's a good idea to add specific simulation support in CyberRT, which is designed to be general and lightweight. Please refer to @unacao @vlin17 .

unacao: I don't think we need to add any field in dag proto. Everything should be added to task.h.

PR #11138, opened 2020-04-27
planning: write remainder of a record file (< 100 frames_ to learning…

jmtao:
… data)

### DIFF
--- a/modules/planning/pipeline/feature_generator.cc
+++ b/modules/planning/pipeline/feature_generator.cc
@@ -114,6 +114,17 @@ void FeatureGenerator::WriteOutLearningData(
   ++learning_data_file_index_;
 }
 
+void FeatureGenerator::WriteRemainderData() {
+  if (learning_data_.learning_data_size() > 0) {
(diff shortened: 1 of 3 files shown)


modules/planning/pipeline/feature_generator.cc
```diff
@@ -114,6 +114,17 @@ void FeatureGenerator::WriteOutLearningData(
   ++learning_data_file_index_;
 }
 
+void FeatureGenerator::WriteRemainderData() {
+  if (learning_data_.learning_data_size() > 0) {
+    WriteOutLearningData(learning_data_, learning_data_file_index_);
+  }
+  std::ostringstream msg;
+  msg << "record_file[" << record_file_name_
+      << "] frame_num[" << total_learning_data_frame_num_ << "]";
```
xiaoxq, line 123: Prefer `StrCat`.

jmtao: performance is not a concern here. these are log messages with flexible text, stream is better imo. please approve.

PR #11180, opened 2020-05-04
Prediction: add semantic lstm pedestrian model

kechxu:


modules/prediction/container/obstacles/obstacle.h
```diff
@@ -89,6 +89,8 @@ class Obstacle {
    */
   perception::PerceptionObstacle::Type type() const;
 
+  bool IsPedestrian();
```
xiaoxq, line 92: bool IsPedestrian() const;

PR #11186, opened 2020-05-05
LGSVL Simulator change (part 2)

rongguodong:
Add calibration data for Lincoln2017MKZ.


rongguodong:
> Prefer renaming to things like `LincolnLGSVL`, to avoid misleading users that it's a general profile for any 2017 MKZ.

This name matches the name of our asset bundle in the simulator. To avoid confusion, how about adding a suffix, like Lincoln2017MKZ_LGSVL? So users can easier to match the map to asset bundle using the prefix.

xiaoxq:
> > Prefer renaming to things like `LincolnLGSVL`, to avoid misleading users that it's a general profile for any 2017 MKZ.
> 
> This name matches the name of our asset bundle in the simulator. To avoid confusion, how about adding a suffix, like Lincoln2017MKZ_LGSVL? So users can easier to match the map to asset bundle using the prefix.

Looks good :)

rongguodong:
Changed the name of Lincoln2017MKZ to Lincoln2017MKZ_LGSVL.

PR #11187, opened 2020-05-05
LGSVL Simulator change (part 3)

rongguodong:
Add cyber_bridge.


rongguodong:
> Prefer moving to `modules/contrib` as we try to make CyberRT as compact and minimum as possible.

Is `modules/contrib` the place holding 3rd party contributed codes?

rongguodong:
I have moved the code to "modules/contrib/cyber_bridge" and fixed the lint failures.

modules/contrib/cyber_bridge/LICENSE
```diff
@@ -0,0 +1,59 @@
+SIMULATOR SOFTWARE LICENSE AGREEMENT
```
xiaoxq, line 1: Not sure if we are compatible with this. I'll leave it to @jinghaomiao @jmtao

modules/contrib/cyber_bridge/BUILD
```diff
@@ -0,0 +1,27 @@
...
+    srcs = [
+        "bridge.cc",
+        "node.cc",
+        "node.h",
+        "client.cc",
+        "client.h",
+        "clients.cc",
+        "clients.h",
+        "server.cc",
+        "server.h",
```
xiaoxq, line 16: Prefer slim targets. Each pair of `.cc` and `.h` should be in its own cc_library or cc_binary, and claims dependency relationship clearly.

rongguodong: I split out two libraries: bridge_client and bridge_server.
There are circular inclusion in node.cc, client.cc, and clients.cc. So I put them in one library.

rongguodong:
I see this build error:

modules/contrib/cyber_bridge/LICENSE
jmtao: I have no idea.  let's see if @jinghaomiao may has any input. otherwise, we'll have to consult legal.

modules/contrib/cyber_bridge/BUILD
```diff
@@ -0,0 +1,46 @@
...
+        ":bridge_client",
+    ],
+)
+
+cc_library(
+    name = "bridge_client",
+    srcs = [
+        "client.cc",
+        "clients.cc",
+        "node.cc",
```
xiaoxq, line 34: Please continue decoupling. In most cases we want at most 1 file in `srcs` and 1 in `hdrs`. Thanks!

xiaoxq:
> I see this build error:
> [ERROR] Failed to pull docker image : apolloauto/apollo:map_volume-sunnyvale_loop-latest
> 
> What does it mean?

This could happen time to time due to docker hub pulling issue. It's OK as the CI will retry later.

modules/contrib/cyber_bridge/BUILD
rongguodong: How do you deal with circular reference? I.e. class A uses class B, and class B uses class A too. In this case, do we have to put A.cc and B.cc in one build target?

xiaoxq: Actually one of the benefit of slim-target-principle is to avoid circular reference, which indicates some refactoring.
Or maybe you can try forward-declaration, while it's still not that good though.

rongguodong: I separated out client.h, and now there are no more than 1 file in `srcs` and/or `hdrs`. The dependency graph is like the following:

cyber_bridge
 |
 |--bridge_server
     |
     |--bridge_client
         |
         |--bridge_clients
         |   |
         |   |--bridge_client_header
         |
         |--bridge_node
             |
             |--bridge_client_header

modules/contrib/cyber_bridge/LICENSE
rongguodong: I have replaced the license with Apache 2.0.

PR #11189, opened 2020-05-05
LGSVL Simulator change (part 5)

rongguodong:
Changed config files to have the simulator running correctly.

### DIFF
--- a/scripts/bootstrap_lgsvl.sh
+++ b/scripts/bootstrap_lgsvl.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+###############################################################################
+# Copyright 2020 The Apollo Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+###############################################################################
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+DREAMVIEW_URL="http://localhost:8888"
+
+cd "${DIR}/.."
+
+# Make sure supervisord has correct coredump file limit.
+ulimit -c unlimited
+
+source "${DIR}/apollo_base.sh"
+
+function start() {
+    nohup cyber_launch start modules/drivers/tools/image_decompress/launch/image_decompress.launch & 
+    ./scripts/monitor.sh start
+    ./scripts/dreamview.sh start
+    if [ $? -eq 0 ]; then
+        sleep 2  # wait for some time before starting to check
+        http_status="$(curl -o /dev/null -I -L -s -w '%{http_code}' ${DREAMVIEW_URL})"
+        if [ $http_status -eq 200 ]; then
+            echo "Dreamview is running at" $DREAMVIEW_URL
+        else
+            echo "Failed to start Dreamview. Please check /apollo/data/log or /apollo/data/core for more information"
+        fi
+    fi
+}
+
+function stop() {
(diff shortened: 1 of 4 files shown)


lx18233184051:
please add at least one reviewer before merge.

modules/dreamview/conf/hmi_modes/mkz_standard_debug.pb.txt
```diff
@@ -32,7 +32,7 @@ cyber_modules {
 cyber_modules {
   key: "Localization"
   value: {
-    dag_files: "/apollo/modules/localization/dag/dag_streaming_msf_localization.dag"
+    dag_files: "/apollo/modules/localization/dag/dag_streaming_rtk_localization.dag"
```
xiaoxq, line 35: We may prefer MSF localization by default.

scripts/bootstrap.sh
```diff
@@ -27,6 +27,7 @@ ulimit -c unlimited
 source "${DIR}/apollo_base.sh"
 
 function start() {
+    nohup cyber_launch start modules/drivers/tools/image_decompress/launch/image_decompress.launch &
```
xiaoxq, line 30: IIRC, this tool is pretty heavy for online use.
We can leave these kind of steps in the instruction document as a start.

xiaoxq:
Also add perception folks to review the config change.

modules/dreamview/conf/hmi_modes/mkz_standard_debug.pb.txt
rongguodong: LGSVL Simulator does not work with MSF localization for some reason. Since MSF is not open sourced, we need your help to figure out why it does not work. Before this is fixed, we have to use RTK for now.

scripts/bootstrap.sh
rongguodong: Since our simulator send out compressed images, we need this decompress module. Otherwise, traffic light perception does not work correctly.

xiaoxq: Yeah I understand. How about adding a `scripts/lgsvl_bootstrap.sh` instead of changing the one we use for road test?

modules/dreamview/conf/hmi_modes/mkz_standard_debug.pb.txt
xiaoxq: Ditto. We adapt to road test by default. Prefer adding dedicated mode for simulation purpose.

rongguodong: Added a new mode `mkz_lgsvl.pb.txt`.

scripts/bootstrap.sh
rongguodong: Added `scripts/bootstrap_lgsvl.sh`.

scripts/bootstrap_lgsvl.sh
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env bash
+
+###############################################################################
+# Copyright 2017 The Apollo Authors. All Rights Reserved.
```
xiaoxq, line 4: 2020

rongguodong: Corrected.

modules/perception/production/conf/perception/lidar/velodyne128_segmentation_conf.pb.txt
```diff
@@ -1,5 +1,5 @@
 sensor_name: "velodyne128"
 enable_hdmap: true
-lidar_query_tf_offset: 0
+lidar_query_tf_offset: 200
```
xiaoxq, line 3: Need to be confirmed by @gchen-apollo @weidezhang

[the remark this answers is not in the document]
rongguodong: In the latest version of this PR, I restored this file and created a copy just for LGSVL Simulator, and changed this value there.

Without this change, the obstacle detection does not work well. It complains timestamp error. The reason may be because the timestamp of our sensors uses the simulation time, which is different to real time. With a large value of `lidar_query_tf_offset`, the obstacle detection is more stable.

Once you started to use our simulator, maybe you can tell me a better way to resolve this issue. We can change it further by then.

PR #11192, opened 2020-05-06
Fix user account issue with docker

panxuechao:


docker/scripts/dev_start.sh
```diff
@@ -86,6 +85,12 @@ do
   fi
 done
 }
+function set_registry_mirrors()
+{
+sed -i '$aDOCKER_OPTS=\"--registry-mirror=http://hub-mirror.c.163.com\"' /etc/default/docker
+sed -i '$i  ,"registry-mirrors": [ "http://hub-mirror.c.163.com"]' /etc/docker/daemon.json
```
storypku, line 91: L91 doesn't work if /etc/docker/daemon.json doesn't exist

PR #11195, opened 2020-05-06
Prediction: add offline mode 6 to dump record with new prediction messages

kechxu:


modules/tools/prediction/data_pipelines/scripts/records_to_dump_records.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+
+###############################################################################
+# Copyright 2019 The Apollo Authors. All Rights Reserved.
```
xiaoxq, line 4: 2020

PR #11215, opened 2020-05-08
introduce bazel  "--output_user_root" startup option

storypku:
This fixes #issues/11198. We don't put bazel cache in $HOME/.cache outside container; Instead, we take advantage of  bazel's `--output_user_root` startup option (pointing to /apollo/bazel-root inside container, $APOLLO_ROOT/bazel-root outside).  

Although bazel-0.5.3 doesn't support `.bazelignore` file yet, it makes no harm. As soon as we moved to bazel 3.x, `.bazelignore` will come to life!

### DIFF
--- a/tools/bazel.rc
+++ b/tools/bazel.rc
@@ -12,6 +12,7 @@ startup --batch_cpu_scheduling
 
 startup --host_jvm_args=-XX:-UseParallelGC
 
+startup --output_user_root=/apollo/.cache/bazel
 # +------------------------------------------------------------+
 # | Test Configurations                                        |
 # +------------------------------------------------------------+
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,0 +1,4 @@
+.cache
+docs
+docker
+scripts
--- a/.gitignore
+++ b/.gitignore
@@ -70,3 +70,6 @@ docs/demo_guide/*.record
 
 # Bash Environment
 .dev_bash_hist
+
+# bazel cache and others
+.cache/
--- a/apollo.sh
+++ b/apollo.sh
@@ -298,7 +298,8 @@ function warn_proprietary_sw() {
 }
 
 function release() {
-  RELEASE_DIR="${HOME}/.cache/apollo_release"
+  CACHE_DIR="/apollo/.cache"
+  RELEASE_DIR="${CACHE_DIR}/apollo_release"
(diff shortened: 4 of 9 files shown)


tools/bazel.rc
```diff
@@ -12,6 +12,7 @@ startup --batch_cpu_scheduling
 
 startup --host_jvm_args=-XX:-UseParallelGC
 
+startup --output_user_root=/apollo/bazel-root
```
xiaoxq, line 15: Please also retire the old cache mechanism, such as:
https://github.com/ApolloAuto/apollo/blob/master/docker/scripts/dev_start.sh#L211
https://github.com/ApolloAuto/apollo/blob/master/docker/scripts/dev_start.sh#L283

And as it's a big change, please conduct intensive testing to verify that everything works as expected :)

jinghaomiao: Good catch. It's important to let the team test internally to an extensive extent before pushing to the public.

storypku: > Good catch. It's important to let the team test internally to an extensive extent before pushing to the public.

Got it. Actually I tested on bazel-0.5.3 and bazel-3.x systems. But didn't realize  that `apollo.sh` script have that directory mounted.  Thanks Xiangquan for pointing it out.

xiaoxq:
> We have many modules, and great number of developers working together. To make it easier to refer to,
> please always add module name in your commit message. "<module>: msg...".
> In your case, probably it's "build: "? please consult with @xiaoxq . thx

Agree! @storypku Let's use "Bazel: xxx yyy zzz." and "Docker: xxx yyy zzz." for all commit message and pull request title.

storypku:
> > We have many modules, and great number of developers working together. To make it easier to refer to,
> > please always add module name in your commit message. ": msg...".
> > In your case, probably it's "build: "? please consult with @xiaoxq . thx
> 
> Agree! @storypku Let's use "Bazel: xxx yyy zzz." and "Docker: xxx yyy zzz." for all commit message and pull request title.

OK.

storypku:
> We have many modules, and great number of developers working together. To make it easier to refer to,
> please always add module name in your commit message. "<module>: msg...".
> In your case, probably it's "build: "? please consult with @xiaoxq . thx

Great idea! Thanks to Jiaming Tao from Jiaming Liu. :-)

PR #11217, opened 2020-05-10
Update Waypoint_Following--Operation_And_Questions_cn.md

daohu527:
the "bash gps.sh" and "bash localization.sh" need "start" parameter

### DIFF
--- a/docs/specs/D-kit/Waypoint_Following--Operation_And_Questions_cn.md
+++ b/docs/specs/D-kit/Waypoint_Following--Operation_And_Questions_cn.md
@@ -78,7 +78,7 @@ enable_sender_log: true
 
 在命令行中执行
 ```
-bash bootstrap.sh
+bash bootstrap.sh start
 ```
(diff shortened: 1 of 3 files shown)


daohu527:
thx! fixed the commit log

daohu527:
Thx for you reply,  it will work without the "start" but maybe some error in the feature.  

#### question
if the parameter number is less than 3, the "cmd" is empty and will enter the "*)" branch. But the '$@' parameter will be "drivers/gnss gnss" (this maybe the shift command bug, there are just two parameters!!! but shift 3 so the shift 3 will not work)
https://github.com/ApolloAuto/apollo/blob/0b6b21742c6204625b87c0cd333c81579944aa16/scripts/apollo_base.sh#L417
so if you enter "bash gps.sh", the "start_customized_path" input parameter will be
**"drivers/gnss gnss drivers/gnss gnss"** beacuse the shift 3 will not work(there are just two parameters!!!).

for now the "start_customized_path" will just read the front 2 paramter, so there is no problem to use "bash gps.sh", but it might be a bit strange to use this way.
https://github.com/ApolloAuto/apollo/blob/0b6b21742c6204625b87c0cd333c81579944aa16/scripts/apollo_base.sh#L247-L250

#### solution
there are two solutions to do with this
1. if "cmd" in "run_customized_path" is empty then filled it with default "start" then there is no question.
2. add start parameter while run the bash script

PR #11224, opened 2020-05-11
Planning: skip collision check with virtual obstacle when do scenario…

Capri2014:
… switch


Capri2014:
> this method is not even in use.
> u wanna delete it?

It is currently only used in `park_and_go` scenario and is causing the stage transition bug for `adjust` stage and its next stage. I am fixing it now and thinking about refactoring this out or at least moving it to within `park_and_go` scenario only.

PR #11235, opened 2020-05-13
Docker: Fix copt compiling warnings.

xiaoxq:


xiaoxq:
cc1: warning: '-Werror=' argument '-Werror=reorder' is not valid for C

PR #11249, opened 2020-05-14
Octopus: modify planning component to support multi-adc simulation

changsh726:
I'm not sure this is a proper approach to change the output channels of octopus dependent components, and if not, I'd like your advice.

### DIFF
--- a/cyber/common/global_data.cc
+++ b/cyber/common/global_data.cc
@@ -67,6 +67,8 @@ GlobalData::GlobalData() {
   is_reality_mode_ =
       config_.run_mode_conf().run_mode() == proto::RunMode::MODE_REALITY;
 
+  is_octopus_mode_ = false;
+
   const char* run_mode_val = ::getenv("CYBER_RUN_MODE");
   if (run_mode_val != nullptr) {
     std::string run_mode_str(run_mode_val);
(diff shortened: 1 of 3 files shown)


cyber/common/global_data.cc
```diff
@@ -67,6 +67,8 @@ GlobalData::GlobalData() {
   is_reality_mode_ =
       config_.run_mode_conf().run_mode() == proto::RunMode::MODE_REALITY;
 
+  is_octopus_mode_ = false;
```
xiaoxq, line 70: Please refer to https://github.com/ApolloAuto/apollo/pull/11131#discussion_r415933639
We'd better avoid such change in Cyber, unless approved by @unacao and @vlin17.

cyber/common/global_data.h
```diff
@@ -58,6 +58,12 @@ class GlobalData {
   void EnableSimulationMode();
   void DisableSimulationMode();
 
+  void EnableOctopusMode();
+  bool IsOctopusMode() const;
+
+  void SetADCId(const int& adc_id);
+  int ADCId() const;
```
unacao, line 65: cyber global data is a singleton, if we set the adc_id here, it cannot be set to different values for different planning modules.

cyber/common/global_data.cc
jmtao: better not use code name/project name as variable name here. 
it's hard to understand what "octopus" is.  poor readability

unacao:
According to @ycool, you can add the adc_id field in https://github.com/ApolloAuto/apollo/blob/master/modules/planning/proto/planning_config.proto so that config for each ADC can be different.

modules/planning/planning_component.cc
```diff
@@ -56,45 +57,89 @@ bool PlanningComponent::Init() {
...
-      });
-
-  pad_msg_reader_ = node_->CreateReader<PadMessage>(
-      FLAGS_planning_pad_topic,
-      [this](const std::shared_ptr<PadMessage>& pad_msg) {
-        ADEBUG << "Received pad data: run pad callback.";
-        std::lock_guard<std::mutex> lock(mutex_);
-        pad_msg_.CopyFrom(*pad_msg);
-      });
+  if (GlobalData::Instance()->IsOctopusMode()) {
```
jmtao, line 60: non OctopusMode, these readers shall be created as well.

PR #11251, opened 2020-05-14
dreamview times out on system_status

Fla3inH0tCheet0s:
Since the monitor can crash, it would be nice if there is an indication in dreamview that it wasn't receiving updates.

### DIFF
--- a/modules/dreamview/backend/hmi/hmi_worker.cc
+++ b/modules/dreamview/backend/hmi/hmi_worker.cc
@@ -574,15 +574,29 @@ void HMIWorker::ResetComponentStatusTimer() {
 }
 
 void HMIWorker::UpdateComponentStatus() {
-  static constexpr uint64_t kSecondsTillComponentReset(2.5);
-  if (cyber::Time::Now().ToSecond() - last_status_received_s_.load() >
-      kSecondsTillComponentReset) {
-    WLock wlock(status_mutex_);
-    for (auto& monitored_component : *status_.mutable_monitored_components()) {
-      monitored_component.second.set_status(ComponentStatus::UNKNOWN);
-      monitored_component.second.set_message("Status not reported by Monitor.");
+  static constexpr double kSecondsTillTimeout(2.5);
+  const double now = cyber::Time::Now().ToSecond();
+  if (now - last_status_received_s_.load() > kSecondsTillTimeout) {
+    if (!monitor_timed_out_) {
+      WLock wlock(status_mutex_);
+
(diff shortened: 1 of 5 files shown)


Fla3inH0tCheet0s:
I don't have a specific error case for the monitor in mind. It's mostly about being defensive and handling the crash if it does happen. Is there another way to determine that the monitor crashed in dreamview?

xiaoxq:
> I don't have a specific error case for the monitor in mind. It's mostly about being defensive and handling the crash if it does happen. Is there another way to determine that the monitor crashed in dreamview?

You're right that it's important to also check monitor's status. But as of now it seems pretty simple and under careful caring. The 1 sec interval is kind of too frequent IMO. So I'd like to make a TODO but keep as it is. I think it's better to add in both Dreamview and Guardian to make sure the car is safely handled if monitor module is down.

Fla3inH0tCheet0s:
>  as of now it seems pretty simple and under careful caring.

What is the current state of determining monitor is not running?
```
pgrep -f dag/monitor
```
? It would be nice to move away from querying the terminal actively and move towards being notified.


> The 1 sec interval is kind of too frequent IMO

What is the concern of a frequent status message? The message does not seem big.

xiaoxq:
> > as of now it seems pretty simple and under careful caring.
> 
> What is the current state of determining monitor is not running?
> 
> ```
> pgrep -f dag/monitor
> ```
> 
> ? It would be nice to move away from querying the terminal actively and move towards being notified.

That's why I'm kind of wondering if you have real failing cases. In the past we do count on that the monitor is simple, well maintained and intensively tested. Besides that, it's a real time monitor, which means it can be off at offline when people are doing simulation or replaying bags. In such cases the notification would be noisy. So we'd like to take a second think on this change. It's nice though and appreciated!

> > The 1 sec interval is kind of too frequent IMO
> 
> What is the concern of a frequent status message? The message does not seem big.

Actually the monitor sends update immediately when the status changed. And if there is no change it sends by interval as a backup. We just try to make it as quiet and lightweight as possible.

Fla3inH0tCheet0s:
There was an issue with the previous commit. hmi_worker would set `status_changed_` and cause the frontend to go out of sync with the backend. Now, dreamview should turn on modules even if the monitor is not running.

Also, I added similar functionality to guardian. It will start out timed out, which is by design. I hope that doesn't interrupt operational flow.

My main motivation with this pull request is to facilitate developing custom monitors. Unfortunately, they are still experimental and can crash. The person developing those monitors isn't the same as the person testing them. Therefore, having feedback in dreamview is helpful.

PR #11302, opened 2020-05-21
planning: add a task to call inference model and get learning data

jmtao:


modules/planning/tasks/learning_model/learning_model_inference_task.cc
```diff
@@ -0,0 +1,104 @@
...
+
+  std::vector<common::TrajectoryPoint> trajectory_points;
+  ConvertTrajectory(learning_data_frame.output(), &trajectory_points);
+
+  frame->set_learning_data_adc_future_trajectory_points(trajectory_points);
+
+  return Status::OK();
+}
+
+void LearningModelInferenceTask::ConvertTrajectory(
```
JasonZhou404, line 70: current model only fill the following field of TrajectoryPointFeature (x, y, v, a, relative_time, timestamp_sec)

PR #11305, opened 2020-05-21
storytelling: modify storytelling to support multi-adc simulation.

changsh726:


modules/storytelling/dag/storytelling.dag
```diff
@@ -4,6 +4,7 @@ module_config {
         class_name : "Storytelling"
         config {
             name: "storytelling"
+            config_file_path: "apollo/modules/storytelling/conf/storytelling_conf.pb.txt"
```
unacao, line 7: I think you need "/apollo"

Issue #11296, opened 2020-05-20
apollo/scripts directory is no longer added to path for $USER_ID 1000


lemketron:
**Describe the bug**
apollo/scripts directory is no longer added to path for $USER_ID 1000

**To Reproduce**
1. docker/scripts/dev_start.sh
2. docker/scripts/dev_into.sh
3. bootstrap.sh
4. bootstrap.sh script not found in apollo master since "/apollo/scripts" not in path (for $USER_ID 1000) but worked fine in apollo 5.0

**Expected behavior**
/apollo/scripts should be in PATH and bootstrap.sh (and other scripts) should launch without full path.

In apollo, docker/scripts/dev_start.sh calls scripts/docker_adduser.sh which (besides adding the user) appears to export /apollo/scripts into the PATH.

In apollo 5.0, calling docker_adduser.sh is only skipped if $USER is "root".

In apollo master, calling docker_adduser.sh is now also skipped if $USER is "apollo" or if $USER_ID is 1000 (which is typically the user id of the first user who creates and installs the system).

This change happened in [commit a9fb059d](https://github.com/ApolloAuto/apollo/commit/a9fb059d7daa82aeb1f6956c29eef23bdcdfb3e7) to apollo master in April.

There are several other things that happen in docker_adduser.sh (setting user file ownership, setting access permissions on GPS and camera devices, etc.) that appear to also be skipped now for USER_ID 1000 so I'm curious if that might cause problems for some users.

Even if my guess as to why this is happening is wrong, the issue in the bug title (that /apollo/scripts is no longer in the PATH in apollo master) is still the main issue that I am reporting here.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04

lemketron:
Another possible fix for this might be to add `/apollo/scripts` into the export PATH line in `scripts/apollo_base.sh`...?

storypku:
> Another possible fix for this might be to add `/apollo/scripts` into the `export PATH` line in `scripts/apollo_base.sh`...?


> **Describe the bug**
> apollo/scripts directory is no longer added to path for $USER_ID 1000
> 
> **To Reproduce**
> Steps to reproduce the behavior:
> 
> 1. docker/scripts/dev_start.sh
> 2. docker/scripts/dev_into.sh
> 3. bootstrap.sh
> 4. bootstrap.sh script not found in apollo master since "/apollo/scripts" not in path (for $USER_ID 1000) but worked fine in apollo 5.0
[...]

PR #11306, opened 2020-05-21
Docker: execute docker_adduser.sh if USER is not root

SeasoulChris:
execute docker_adduser.sh if USER is not root

### DIFF
--- a/docker/scripts/dev_start.sh
+++ b/docker/scripts/dev_start.sh
@@ -424,9 +424,7 @@ function main(){
     fi
     set +x
 
-    # User with uid=1000 or username=apollo excluded
-    if [[ "${USER}" != "root" ]] && [[ "${USER}" != "apollo" ]] \
-        && [[ $USER_ID -ne 1000 ]]; then
+    if [[ "${USER}" != "root" ]]; then
         docker exec -u root $APOLLO_DEV bash -c '/apollo/scripts/docker_adduser.sh'
     fi
 


docker/scripts/dev_start.sh
```diff
@@ -424,9 +424,7 @@ function main(){
     fi
     set +x
 
-    # User with uid=1000 or username=apollo excluded
-    if [[ "${USER}" != "root" ]] && [[ "${USER}" != "apollo" ]] \
-        && [[ $USER_ID -ne 1000 ]]; then
+    if [[ "${USER}" != "root" ]]; then
```
xiaoxq, line 427: We may need to keep the `"${USER}" != "apollo"` part, as there is already a `apollo` user in the image, @storypku correct?

SeasoulChris: here is a situation: if  default user_name of system is `apollo`, when he execute dev_start.sh, docker_adduser.sh will not be executed，then some devices such as ttyACM* will not get proper permissions

SeasoulChris: sadly , now the default system user_name of D-kit is apollo

storypku: > We may need to keep the `"${USER}" != "apollo"` part, as there is already a `apollo` user in the image, @storypku correct?

Yeap. So we should re-factore `docker_add_user.sh` asap to fix https://github.com/ApolloAuto/apollo/issues/11296

PR #11307, opened 2020-05-21
Perception: allow multi RPN heads and a set of anchors for each

jeroldchen:


modules/perception/lidar/lib/detection/lidar_point_pillars/point_pillars.cc
```diff
@@ -42,21 +42,30 @@ namespace apollo {
...
+const int PointPillars::kGridZSize =
+    static_cast<int>((kMaxZRange - kMinZRange) / kPillarZSize);
 const int PointPillars::kRpnInputSize = 64 * kGridXSize * kGridYSize;
-const int PointPillars::kNumAnchorXInds = kGridXSize * 0.5;
-const int PointPillars::kNumAnchorYInds = kGridYSize * 0.5;
-const int PointPillars::kNumAnchorPerLoc = 6;
-// kNumAnchorPerLoc need to be equal to length of kAnchorRo
-const int PointPillars::kNumAnchor =
-    kNumAnchorXInds * kNumAnchorYInds * kNumAnchorPerLoc;
+const int PointPillars::kNumAnchor = 100 * 100 * 12 + 160 * 160 * 8;
```
jinghaomiao, line 68: Regardng the const variables, if they don't change, you can consider static variable. If they may change with different settings, better use a config file to pass the parameters. @xiaoxq what's your take on this?

jinghaomiao:
@jeroldchen lint check failed. Please also correct that.

modules/perception/lidar/lib/detection/lidar_point_pillars/point_pillars.cc
jeroldchen: Currently they have to be static const, cause memory malloced to cuda arrays at the build-time needs a fixed size provided by those build-time static variables. Is it possible to move these const variables into a config file while cuda arrays can be malloced memory at the build-time?

jeroldchen:
> @jeroldchen lint check failed. Please also correct that.

Fixed in a new commit.

modules/perception/lidar/lib/detection/lidar_point_pillars/point_pillars.cc
```diff
@@ -318,37 +342,43 @@ void PointPillars::GenerateAnchors(float* anchors_px_, float* anchors_py_,
...
-  anchor_x_count[0] = 0;
-  for (int i = 0; i < kNumAnchorXInds; ++i) {
-    anchor_x_count[i] = static_cast<float>(i) * x_stride + x_offset;
-  }
-  float anchor_y_count[kNumAnchorYInds];
-  anchor_y_count[0] = 0;
-  for (int i = 0; i < kNumAnchorYInds; ++i) {
-    anchor_y_count[i] = static_cast<float>(i) * y_stride + y_offset;
-  }
+  int ind = 0;
```
lx18233184051, line 345: you could define this nearest where it's used. L367

lx18233184051: sorry I look messed.

modules/perception/lidar/lib/detection/lidar_point_pillars/point_pillars.cc
jinghaomiao: Good question. Please try to group all parameters need by CUDA into a separate class with static const variables (e.g. class CudaParams { const static kVar1 ...... } and reference them in your PointPillars class so that we can **separate** parameters setting (with comments for each individual parameter) and workflow logic for a cleaner code organization purpose. What do you think?

xiaoxq: > Regardng the const variables, if they don't change, you can consider static variable. If they may change with different settings, better use a config file to pass the parameters. @xiaoxq what's your take on this?

Agree. Prefer `static constexpr`.

weidezhang: how about we change the style in a separate commit ? I need some performance number this week.

jeroldchen: @jinghaomiao @weidezhang Parameters are now assigned values from a new class Params's static member variables and functions.

jinghaomiao:
@jeroldchen please resolve the conflicts by rebasing with the master branch. Thanks!

PR #11311, opened 2020-05-22
Script: apollo.bashrc split out from apollo_base.sh and bugfix for clang-format.sh

storypku:
1) `apollo.bashrc` split out from `apollo_base.sh`:  Use the former as a general `.bashrc` file, consisting of functions and enviroment variables for general purpose only.  The later can be used for apollo specific functionalities and settings.
2) Bufix for clang-format.sh: Enhanced check and only lint c/c++/cuda files using clang-format/llvm

### DIFF
--- a/scripts/apollo.bashrc
+++ b/scripts/apollo.bashrc
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+
+###############################################################################
+# Copyright 2020 The Apollo Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


scripts/apollo.bashrc
```diff
@@ -0,0 +1,102 @@
...
+  echo "=============================================="
+}
+
+function get_now() {
+  echo "$(date +%s)"
+}
+
+function print_time() {
+  END_TIME=$(get_now)
+  ELAPSED_TIME=$(echo "$END_TIME - $START_TIME" | bc -l)
```
storypku, line 55: Here is a dangling START_TIME variable that was never defined across the `${APOLLO_ROOT_DIR}/scripts/` directory ...

PR #11312, opened 2020-05-22
Planning: implementation of trajectory imitation model inference

JasonZhou404:


modules/planning/common/planning_gflags.cc
```diff
@@ -550,3 +550,7 @@ DEFINE_int32(learning_data_obstacle_history_time_sec, 3.0,
              "time sec (second) of history trajectory points for a obstacle");
 DEFINE_int32(learning_data_frame_num_per_file, 100,
              "number of learning_data_frame to write out in one data file.");
+DEFINE_bool(planning_use_cuda, true, "whether use cuda for model inference");
+DEFINE_string(test_model_path,
```
jmtao, line 554: use conf file.  https://github.com/ApolloAuto/apollo/blob/master/modules/planning/proto/task_config.proto#L46

PR #11314, opened 2020-05-22
Script: docker_start_user for non-root users

storypku:


diegohu:
Suggest use `git mv` to rename the file

storypku:
> Suggest use `git mv` to rename the file

Yeap. However, as there are many changes, `git` considers it a new file when i forgot to `git mv` and using `rm old.file` and `git add  new.file` instead...

xiaoxq:
> Suggest use `git mv` to rename the file

Should be fine. Actually git will do the calculation on content and decide if it's `renaming` or `new`. And it doesn't matter how user created the file.
https://stackoverflow.com/questions/433111/how-to-make-git-mark-a-deleted-and-a-new-file-as-a-file-move

PR #11327, opened 2020-05-25
Docker: user permission bugfix the easy way

storypku:


docker/build/user_dev.x86_64.dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM apolloauto/apollo:dev-18.04-x86_64-20200507_0633
+ENV DEBIAN_FRONTEND=noninteractive
+
+WORKDIR /apollo
+USER root
+
+RUN deluser --remove-home apollo
+RUN deluser --remove-home changsonghong
+RUN deluser --remove-home liujiaming
```
xiaoxq, line 9: Should not include these in a Dockerfile which look weird. If the base image is dirty, just clean it and overwrite the tag.
And to reduce layers, you can RUN commands in one line.

storypku: > Should not include these in a Dockerfile which look weird. If the base image is dirty, just clean it and overwrite the tag.
> And to reduce layers, you can RUN commands in one line.

Done.

PR #11329, opened 2020-05-25
Delete the unuseful pictures which are added by me

lishuanghua:
Docs : Delete the unuseful pictures which are added by me


xiaoxq:
CLA not signed. Plz check if you committed with a different email.

PR #11343, opened 2020-05-27
storytelling: refactor to support muli-adc simulation

changsh726:


modules/storytelling/story_tellers/close_to_junction_teller.h
```diff
@@ -47,6 +53,7 @@ class CloseToJunctionTeller : public BaseTeller {
   std::string yield_sign_id_;
   double yield_sign_distance_;
   StorytellingConfig config_;
+  std::shared_ptr<FrameManager> frame_manager_;
```
unacao, line 56: is it better to save this private field in BaseTeller? otherwise we need to add this for every new type of story teller

changsh726: ok, I just not sure the frame manager is also needed by other type of teller

jinghaomiao:
@changsh726 @unacao : I merged the code by mistake. @changsh726 please fix the issues as @unacao suggested and make another pull request. Thanks both.

unacao:
> @changsh726 @unacao : I merged the code by mistake. @changsh726 please fix the issues as @unacao suggested and make another pull request. Thanks both.

No worries, it does not break the code.

PR #11351, opened 2020-05-28
prediction: refactor to support muli-adc simulation

changsh726:


modules/prediction/common/message_process.cc
```diff
@@ -143,11 +150,12 @@ void MessageProcess::ContainerProcess(
...
   ptr_obstacles_container->Insert(perception_obstacles);
 
+  ObstaclesPrioritizer obstacles_prioritizer(container_manager);
   // Ignore some obstacles
-  ObstaclesPrioritizer::Instance()->AssignIgnoreLevel();
+  obstacles_prioritizer.AssignIgnoreLevel();
 
   // Scenario analysis
-  ScenarioManager::Instance()->Run();
+  ScenarioManager::Instance()->Run(container_manager.get());
```
unacao, line 158: Just to double check, scenario manager can still run as singleton?

modules/prediction/common/message_process.h
```diff
@@ -36,29 +41,44 @@ class MessageProcess {
...
       const perception::PerceptionObstacles &perception_obstacles);
 
   static void OnPerception(
       const perception::PerceptionObstacles &perception_obstacles,
+      const std::shared_ptr<ContainerManager> &container_manager,
+      EvaluatorManager *evaluator_manager, PredictorManager *predictor_manager,
       PredictionObstacles *const prediction_obstacles);
 
   static void OnLocalization(
+      ContainerManager *container_manager,
```
unacao, line 68: Curious to know why we use raw pointer of ContainerManager in some functions, and  std::shared_ptr<ContainerManager> in others

changsh726: in some function std::shared_ptr of ContainerManager is used to construct an ObstaclesPrioritizer instance, and nothing needs to be constructed by ContainerManager in others.

modules/prediction/common/message_process.cc
kechxu: I think we'd better avoid using singleton for ScenarioManager here because it has a private member variable inside. Also, to make the whole thing consistent.

changsh726: Ok.

PR #11352, opened 2020-05-28
planning: support routing with multiple paralle passages when(change …

jmtao:
…lanes) generating local_routing for learning data

### DIFF
--- a/modules/planning/common/message_process.cc
+++ b/modules/planning/common/message_process.cc
@@ -238,19 +238,7 @@ void MessageProcess::OnRoutingResponse(
   const apollo::routing::RoutingResponse& routing_response) {
   ADEBUG << "routing_response received at frame["
         << total_learning_data_frame_num_ << "]";
-  routing_lane_segment_.clear();
-  for (int i = 0; i < routing_response.road_size(); ++i) {
-    for (int j = 0; j < routing_response.road(i).passage_size(); ++j) {
-      for (int k = 0; k < routing_response.road(i).passage(j).segment_size();
-          ++k) {
-        const auto& lane_segment =
-            routing_response.road(i).passage(j).segment(k);
-        routing_lane_segment_.push_back(
-            std::make_pair(lane_segment.id(),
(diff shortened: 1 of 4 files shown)


modules/planning/common/message_process.cc
```diff
@@ -326,8 +314,9 @@ void MessageProcess::ProcessOfflineData(const std::string& record_file) {
   }
 }
 
-int MessageProcess::GetADCCurrentRoutingIndex() {
-  if (localizations_.empty()) return -1;
+bool MessageProcess::GetADCCurrentRoutingIndex(
+    int* road_index, double* road_s) {
+  if (localizations_.empty()) return false;
```
xiaoxq, line 319: Add `{
}`

modules/planning/common/message_process.cc
```diff
@@ -358,12 +359,19 @@ apollo::hdmap::LaneInfoConstPtr MessageProcess::GetCurrentLane(
   }
 
   for (auto& lane : lanes) {
-    for (size_t i = 0; i < routing_lane_segment_.size(); ++i) {
-      if (lane->id().id() == routing_lane_segment_[i].first) {
-        return lane;
+    for (int i = 0; i < routing_response_.road_size(); ++i) {
+      for (int j = 0; j < routing_response_.road(i).passage_size(); ++j) {
+        for (int k = 0;
```
xiaoxq, line 364: Loop with for-each style code, to avoid i/j/k and long row.

modules/planning/common/message_process.cc
jmtao: "Short conditional statements may be written on one line... "
the sample has no bracket. :)
https://google.github.io/styleguide/cppguide.html#Conditionals

modules/planning/common/message_process.cc
jmtao: I use index to loop almost everywhere for routing protobuf.  
sometimes I need the index for debugging purpose.

PR #11364, opened 2020-06-02
Fix: change to cyber python2

jeroldchen:
These scripts can only be run by python2, not python3, due to the limitation of pypcd library.


storypku:
> These scripts can only be run by python2, not python3, due to the limitation of pypcd library.

As PY2 support will be abandoned soon, please consider adding its PY3 support.

weidezhang:
i agree. How about we do the opposite. Upgrade to use python 3 instead of using python2? Is there a lot of efforts ?

jeroldchen:
@weidezhang I will try another pypcd package for python3.

PR #11375, opened 2020-06-03
Docker dev: build and test localization module

lfcarol:


storypku:
Try rebasing first before commit.

PR #11390, opened 2020-06-04
DockerDev: Fix bugs and reduce layer size by cleaning cache.

xiaoxq:


docker/build/installers/install_adolc.sh
```diff
@@ -30,5 +30,6 @@ apt-get -y update && \
     libcolpack-dev \
     libadolc-dev
 
-# clean up
-apt-get clean && rm -rf /var/lib/apt/lists/*
+# Clean up cache to reduce layer size.
+apt-get clean && \
```
storypku, line 34: Please note not every install_xxx needs a `rm -rf /var/lib/apt/lists/*` operation, as some of them will be part of a "larger" installer.sh.

xiaoxq: Yeah, so I only added to top-level installers such as install_gpu_support.sh, while skip it's sub-installers.

PR #11394, opened 2020-06-05
Bazel: make modules/prediction build passed.

changsh726:
0. fix some rebase errors
1. `bazel test //modules/prediction...` hasn't work yet, maybe some code need to be changed due to libtorch update.


external/libtorch_cpu.BUILD
```diff
@@ -4,13 +4,15 @@ licenses(["notice"])
...
 cc_library(
     name = "libtorch_cpu",
+    srcs = glob(["lib/*.so"]),
+    hdrs = glob(["*"]),
     includes = [
-        ".",
-        "torch/csrc/api/include",
+        "include",
+        "include/torch/csrc/api/include",
+        "include/torch/csrc/api/include/torch",
```
storypku, line 12: include directories sharing  same ancestor  should be as few as possible. Flexibility is a bad idea here.

modules/prediction/predictor/junction/BUILD
```diff
@@ -24,6 +24,10 @@ cc_test(
         "//modules/prediction:prediction_data",
         "//modules/prediction:prediction_testdata",
     ],
+    linkopts = [
+        "-L/usr/lib/x86_64-linux-gnu/",
```
storypku, line 28: Same as above.

modules/prediction/evaluator/vehicle/BUILD
```diff
@@ -137,6 +144,10 @@ cc_test(
         "//modules/prediction:prediction_data",
         "//modules/prediction:prediction_testdata",
     ],
+    linkopts = [
+        "-L/usr/lib/x86_64-linux-gnu/",
```
storypku, line 148: What if running on an aarch64 machine. Please don't link architecture-specific code here. If it needs to do so, try using bazel's `select` rule.

modules/prediction/BUILD
```diff
@@ -35,6 +35,10 @@ cc_test(
         ":prediction_data",
         ":prediction_testdata",
     ],
+    linkopts = [
+        "-L/usr/lib/x86_64-linux-gnu/",
```
storypku, line 39: And here.

external/python37.BUILD
```diff
@@ -0,0 +1,32 @@
...
+        },
+        no_match_error = "Please Build with an ARM or Linux x86_64 platform",
+    ),
+    hdrs = glob([
+        "include/python3.7/*.h",
+    ]),
+    includes = ["include/python3.7"],
+)
+
+config_setting(
```
storypku, line 24: Don't define custom config settings here. It's hard to maintain.
Use `//tools/platform:x86_mode` and `//tools/platform:aarch64_mode` instead.

external/libtorch_gpu.BUILD
```diff
@@ -4,14 +4,15 @@ licenses(["notice"])
...
+    hdrs = glob(["*"]),
     includes = [
-        ".",
-        "torch/csrc/api/include",
+        "include",
+        "include/torch/csrc/api/include",
+        "include/torch/csrc/api/include/torch",
     ],
     linkstatic = False,
-    linkopts = [
```
storypku, line 12: I prefer this form. As libtorch was pre-installed, and we don't need them  until the linking stage.

changsh726: You are right, and I find that it seems like python3.7 is no longer needed in this way...

PR #11407, opened 2020-06-08
Docs: D-Kit: update the vehicle calibration online contents and common Q&A

macDure:
Update the vehicle calibration online contents and add some common Q&A and tips.


macDure:
@xiaoxq  Please help merge this commit because build failed stopping me merging it.  I just change the doc instead of code. 
The build fault reason is that Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: TLS handshake timeout, [06:01:22][ERROR] Failed to pull docker image : apolloauto/apollo:map_volume-sunnyvale_loop-latest.

PR #11410, opened 2020-06-08
add SensorMeasurement at backend

lj0919:
add SensorMeasurement Message in object at backend


modules/dreamview/proto/simulation_world.proto
```diff
@@ -152,6 +153,12 @@ message Object {
   // perception confidence level. Range: [0,1]
   optional double confidence = 31 [default = 1];
   optional apollo.prediction.ObstaclePriority obstacle_priority = 33;
+
+  //map for sensor measurement
+  map<string, Object> sensors= 36;
```
unacao, line 158: This should be created in message SimulationWorld instead of Object. One object == One Obstacle

PR #11415, opened 2020-06-08
Planning: init inference at inference task init time to avoid slow in…

JasonZhou404:
…ference at first frame


lx18233184051:
//modules/planning/learning_based/model_inference:model_inference_test   FAILED in 19.9s

//modules/planning/tasks/learning_model:learning_model_inference_task_test FAILED in 2.0s

PR #11432, opened 2020-06-11
fix bug in path.cc

owphoo:
start_interpolation_index will equal to num_segments_  when hueristic_start_s >= length_ in function 'GetIndexFromS', which will cause vector out of range.

### DIFF
--- a/modules/map/pnc_map/path.cc
+++ b/modules/map/pnc_map/path.cc
@@ -746,7 +746,7 @@ bool Path::GetProjectionWithHueristicParams(const Vec2d& point,
   int start_interpolation_index = GetIndexFromS(hueristic_start_s).id;
   int end_interpolation_index = static_cast<int>(
       std::fmin(num_segments_, GetIndexFromS(hueristic_end_s).id + 1));
-  int min_index = start_interpolation_index;
+  int min_index = std::min(start_interpolation_index, num_segments_ - 1);
   for (int i = start_interpolation_index; i < end_interpolation_index; ++i) {
     const double distance = segments_[i].DistanceSquareTo(point);
     if (distance < *min_distance) {


modules/map/pnc_map/path.cc
```diff
@@ -746,7 +746,7 @@ bool Path::GetProjectionWithHueristicParams(const Vec2d& point,
   int start_interpolation_index = GetIndexFromS(hueristic_start_s).id;
   int end_interpolation_index = static_cast<int>(
       std::fmin(num_segments_, GetIndexFromS(hueristic_end_s).id + 1));
-  int min_index = start_interpolation_index;
+  int min_index = std::min(start_interpolation_index, num_segments_ - 1);
```
jmtao, line 749: seems not necessary. 
GetIndexFromS() returns (num_segments_ - 1) when start_s exceed upper bound. 
have you run into cases when start_interpolation_index is greater than (num_segments_ - 1)?

jmtao, line 749: seems not necessary.
GetIndexFromS() returns (num_segments_ - 1) when start_s exceed upper bound.
https://github.com/ApolloAuto/apollo/blob/master/modules/map/pnc_map/path.cc#L657

have you run into cases when start_interpolation_index is greater than (num_segments_ - 1)?

owphoo: in
```
InterpolatedIndex Path::GetIndexFromS(double s) const {
...
if (s >= length_) {
    return {num_points_ - 1, 0.0};
  }
  const int sample_id = static_cast<int>(s / kSampleDistance);
  if (sample_id >= num_sample_points_) {
    return {num_points_ - 1, 0.0};
  }
...
}
```
It will return num_points - 1, where 
```
void Path::InitPoints() {
...
num_segments_ = num_points_ - 1;
...
}
```
So the index of **segments_** will be **[0, num_points_ -1)**,  **num_points - 1** is out of bound

PR #11444, opened 2020-06-11
Python3: py_wrapper renamed to //cyber/python/internal

storypku:


cyber/py_wrapper/py_cyber.cc
```diff
@@ -16,11 +16,7 @@
 
 #include "cyber/py_wrapper/py_cyber.h"
 
-#if PY_MAJOR_VERSION >= 3
 #include <python3.6m/Python.h>
```
storypku, line 19: Better if we can get rid of specific python3 versions.

PR #11475, opened 2020-06-16
Planning: updated path optimizer for hybrid model 

sjiang2018:


modules/planning/common/frame.h
```diff
@@ -258,6 +258,10 @@ class Frame {
   std::vector<common::TrajectoryPoint>
       learning_data_adc_future_trajectory_points_;
   bool is_valid_learning_trajectory_ = false;
+
+  // hybrid model related parameters
+  double path_reference_end_pose;
+  size_t trimmed_path_bound_size;
```
xiaoxq, line 264: `path_reference_end_pose_` and `trimmed_path_bound_size_`

sjiang2018: Removed. Thank you!

PR #11512, opened 2020-06-19
routing: fixed a failure by adding acuracy tolarence on s

jmtao:


modules/routing/core/black_list_range_generator.cc
```diff
@@ -129,21 +129,30 @@ void BlackListRangeGenerator::AddBlackMapFromTerminal(
     double end_s, TopoRangeManager* const range_manager) const {
   double start_length = src_node->Length();
   double end_length = dest_node->Length();
-  if (start_s < 0.0 || start_s > start_length) {
+
+  static constexpr double kEpsilon = 1e-2;
+  const double start_s_adjusted =
+      (start_s > start_length && start_s - start_length <= kEpsilon) ?
```
unacao, line 135: should this be "start_s < start_length && start_length - start_s <= kEpsilon"?

jmtao: NO. when start_s <= start_length, it's legitimate. 
we are covering illegitimate cases when (start_s > start_length) and will be reported error and return.
We are giving accuracy tolerance so that when start_s exceed start_length not longer than 0.01m, start_s is adjusted to   start_length.

PR #11525, opened 2020-06-22
Bazel: introduce bazel management for modules/tools/prediction etc.

changsh726:


modules/tools/prediction/data_pipelines/data_preprocessing/combine_features_and_labels.py
```diff
@@ -20,7 +20,7 @@
 import os
 import re
 
-from features_labels_utils import *
+from modules.tools.prediction.data_pipelines.data_preprocessing.features_labels_utils import *
```
storypku, line 23: Avoid use of `import *`  in python.

changsh726: done

changsh726: done

PR #11527, opened 2020-06-22
Planning: using path reference as part of optimization objective.

sjiang2018:


modules/planning/conf/planning.conf
```diff
@@ -52,3 +52,4 @@
 --use_soft_bound_in_nonlinear_speed_opt=false
 
 # --use_road_boundary_from_map=true
+--planning_learning_mode=3
```
jmtao: need revert to 0

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
```diff
@@ -55,6 +55,11 @@ Status PathReferenceDecider::Execute(Frame *frame,
 
 Status PathReferenceDecider::Process(Frame *frame,
                                      ReferenceLineInfo *reference_line_info) {
+  // skip using path reference during lane changing
+  if (reference_line_info->size() > 1) {
```
jmtao: next time, if you haven't pushed this commit yet and find it's not valid, please revert it, instead of pushing an invalid commit, together with your upcoming commits.

modules/planning/conf/scenario/lane_follow_config.pb.txt
```diff
@@ -29,15 +27,6 @@ stage_config: {
     }
   }
```
jmtao: You previous commit to add these tasks haven't been pushed yet. 
Now you are deleting them, it's better to consolidate these changes since you are actually making no changes.

modules/planning/conf/planning.conf
sjiang2018: Fixed. Thank you.

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
sjiang2018: Fixed. Thank you.

modules/planning/conf/scenario/lane_follow_config.pb.txt
sjiang2018: Fixed. Thank you.

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
```diff
@@ -235,7 +240,8 @@ bool PiecewiseJerkPathOptimizer::OptimizePath(
 
   // TODO(Hongyi): update end_state settings
   piecewise_jerk_problem.set_end_state_ref({1000.0, 0.0, 0.0}, end_state);
-  if (end_state[0] != 0) {
+  // pull over scenarios
+  if (end_state[0] != 0 && !is_valid_path_reference) {
```
jmtao, line 244: "end_state[0] != 0" means we are in pull-over now.  
seems questionable here when (this condition && !is_valid_path_reference)
please double check

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
```diff
@@ -122,17 +123,22 @@ common::Status PiecewiseJerkPathOptimizer::Process(
...
+
     if (path_boundary.label().find("regular") != std::string::npos &&
-        path_data.is_valid_path_reference()) {
+        is_valid_path_reference) {
       // when path reference is ready
-      // set end lateral to be at the path reference
-      common::SLPoint path_reference_end_sl;
-      reference_line.XYToSL(path_data.path_reference_end_pose(),
-                            &path_reference_end_sl);
-      end_state[0] = path_reference_end_sl.l();
```
jmtao, line 132: this line can't be deleted?
end_state[0] will be used by a check later on.

sjiang2018: Updated on line 142. Thank you.

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
sjiang2018: Added  comments about this. Thank you.

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
sjiang2018: Added on line 142. Thank you.

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
sjiang2018: Added more explanation here as comments. Thank you.

modules/planning/common/path/path_data.h
```diff
@@ -94,24 +94,16 @@ class PathData {
...
-  const common::PointENU path_reference_end_pose() const {
-    return path_reference_end_pos_;
-  }
-  void set_path_reference_end_pose(
-      const common::PointENU &path_reference_end_pose);
-
   const bool is_valid_path_reference() const {
     return is_valid_path_reference_;
   }
   void set_valid_path_reference(bool is_valid_path_reference) {
```
jmtao, line 100: function name use: set_<member variable name>

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
jmtao: if so, should be OR ??

sjiang2018: Path reference has its own x_ref setting method, which is different from pull-over. Here I intend to restrict the following codes only applicable to pull-over. If using OR, when a non-valid path reference presents, it will also perform pull over scenario optimization process.

modules/planning/common/path/path_data.h
sjiang2018: Can you be more specific? Am I not supposed to use set here?

jmtao: oh I previously typed "set_\<member_variable\>".  seems got truncated as marker. 
to be specific, use "set_is_valid_path_reference()"

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
jmtao: so when we are in PULL OVER, but is_valid_path_reference = TRUE, we don't adjust weight_x_ref based on PULL-OVER? why?
based on your comments "Here I intend to restrict the following codes only applicable to pull-over. ", shall we juse use "if (end_state[0] != 0) {" ?

sjiang2018: We are not using path reference in pull-over scenario, so  is_valid_path_reference is always false in pull-over scenario.

modules/planning/common/path/path_data.h
sjiang2018: Fixed. Thank you.

PR #11550, opened 2020-06-24
Docker_dev: Build all.

xiaoxq:


docker/build/installers/install_bazel.sh
```diff
@@ -24,7 +24,7 @@ cd "$(dirname "${BASH_SOURCE[0]}")"
 . installer_base.sh
 
 TARGET_ARCH=$(uname -m)
-VERSION="3.2.0"
+VERSION="3.3.0"
```
storypku, line 27: Please also update the sha256sum values involved.
Note also for aarch64, we prebuilt bazel  binary of version 3.2.0, and hosted it on bcebos. So either keep version 3.2.0 for aarch64, or build one for 3.3.0 and host it on bcebos.

scripts/apollo_lint.sh
```diff
@@ -34,7 +34,7 @@ function run_cpp_lint() {
     _cpp_lint_impl "${targets}"
 
     if [[ "${STAGE}" == "dev" ]]; then
-        _cpp_lint_impl "$(bazel query //modules/... except //modules/tools/visualizer/...)"
+        _cpp_lint_impl "//modules/..."
```
storypku, line 37: Good job.

docker/build/installers/install_bazel.sh
xiaoxq: done

xiaoxq: For aarch64, I'd like to keep current version and revisit when we have a machine :)

xiaoxq:
@storypku Locked to v3.2 for aarch64. Thanks!

PR #11555, opened 2020-06-25
Docker: updated dev.x86_64 image

storypku:
However, perception failures increased to 24. Please help check them.


xiaoxq:
Not work on my side.

And I also fount an interesting thing:
`bazel test //modules/perception/common/geometry:basic_test` will fail as of now.
But if I rename the TEST name from `calculate_dist_and_dir_to_boundary_list` to `calculate_dist_and_dir_to_boundary_lists` or something else, it will PASS :(

jinghaomiao:
C++17seems resolve the alignment issue, but need to some extensive study to see how it's implemented. Could be explicitly specified?

storypku:
> C++17seems resolve the alignment issue, but need to some extensive study to see how it's implemented. Could be explicitly specified?

We will investigate this later to make it work with c++14 and 17 both.

Issue #9321, opened 2019-08-06
Dreamview's SimControl separated into its own module


Fla3inH0tCheet0s:
As of 099f135a464f85b10599040073806fcb6b2d6c25, Apollo's open source simulator is located in dreamview/backend/sim_control. It would be nice if it was in its own module. That way it can be started and stopped without a UI. This feature is useful for process level, real-time integration frameworks.

I know there are a few solutions to running dreamview without a UI in https://github.com/ApolloAuto/apollo/issues/6795. But it seems simpler to decouple the simulation from the visualizer.

Also, the dreamview UI should probably be changed at the same time. Instead of having the SimControl button on the front page, it should be started like the other programs. There can be another config located in modules/dreamview/conf/hmi_modes/simulation.pb.txt.

xmyqsh:
@Fla3inH0tCheet0s 
That's a good suggestion.
The `dreamview/backend/sim_control` is `PerfectControlModel`.
Do you know where is the `Control-in-loop Simulation`?
@natashadsouza [Dynamic Model - Control-in-loop Simulation](https://github.com/ApolloAuto/apollo/blob/master/docs/specs/dynamic_model.md)

askariz:
Hi
@xmyqsh do you find out where is the Control -in-loop Simulation ? 
@natashadsouza  the announce of apollo that mention r5.0 have the feature of it ? but it seems that i can find out any instruction of it ?

unacao:
control-in-loop can be enabled on Apollo Simulation Service http://apollo.auto/platform/simulation.html

xmyqsh:
@unacao 
Got it. Thanks.

Andor233:
So,it is invisible for us?
Can I find anything about the model in this repo?

unacao:
We provide two dynamic models: Echo and Intelli. Echo is a rule-based well-tuned dynamic model for Lincoln MKZ, and offered by third-party via a license. Due to the license restrictions, we are not able to disclose the details. Intelli is a dynamic model trained by machine learning methods. The corresponding paper will be available on IROS 2019.

daohu527:
Closed due to inactivity. If the problem persists, pls feel free to reopen it or create a new one and refer to it.

PR #11556, opened 2020-06-25
sim_2d prototype

Fla3inH0tCheet0s:
Here's a protoype to address https://github.com/ApolloAuto/apollo/issues/9321.

I'm not sure the best way to fix the documentation. Also, I'm not sure if all the configuration files should have the module added. It might make sense to create a separate simulation configuration.

Dreamview's change in interface is jarring at first, but I think it make more sense.


Fla3inH0tCheet0s:
It's meant to say that the simulation is in two dimensions. I was confused about the naming of sim_control, when the input is planning. Really this two dimensional simulation could take either.

vlin17:
> It's meant to say that the simulation is in two dimensions. I was confused about the naming of sim_control, when the input is planning. Really this two dimensional simulation could take either.

I see. Thanks for the explanation. 
The reason we named it sim_control is because it simulates a 'perfect control' algorithm, which assumes an ideal world where the car can be perfectly placed wherever the planning asks it to be. As control module can't be used offline, this component is a replacement. We apologize for the confusion in the first sight; however, it would be great if we can keep the idea of control module replacement in the name.

modules/dreamview/frontend/src/store/index.js
```diff
@@ -158,10 +158,6 @@ class DreamviewStore {
     }
 
     handleDrivingModeChange(wasAutoMode, isAutoMode) {
-        if (this.options.enableSimControl) {
```
vlin17, line 161: We still want to keep this logic. Just this time, we check if SimControl is on from HMI's componentStatus: hmi.componentStatus.get(XXX)

modules/sim_2d/sim_2d_component.cc
```diff
@@ -293,28 +272,26 @@ void SimControl::OnPlanning(const std::shared_ptr<ADCTrajectory>& trajectory) {
...
 
-void SimControl::Freeze() {
+void Sim2DComponent::Freeze() {
   next_point_.set_v(0.0);
   next_point_.set_a(0.0);
   prev_point_ = next_point_;
 }
 
-void SimControl::RunOnce() {
+void Sim2DComponent::SimStep() {
```
vlin17, line 281: In the apollo repo, we usually use RunOnce() or Proc(). Would be great if we can follow the convention.

modules/sim_2d/sim_2d_component.cc
```diff
@@ -293,28 +272,26 @@ void SimControl::OnPlanning(const std::shared_ptr<ADCTrajectory>& trajectory) {
...
   std::lock_guard<std::mutex> lock(mutex_);
 
   TrajectoryPoint trajectory_point;
   Chassis::GearPosition gear_position;
   if (!PerfectControlModel(&trajectory_point, &gear_position)) {
     AERROR << "Failed to calculate next point with perfect control model";
-    return;
```
vlin17, line 309: How come we don't want to return anymore?

modules/sim_2d/sim_2d_component.cc
```diff
@@ -100,93 +96,69 @@ void SimControl::InitTimerAndIO() {
...
 }
 
-void SimControl::Init(bool set_start_point, double start_velocity,
-                      double start_acceleration) {
+void Sim2DComponent::InitState(bool set_start_point, double start_velocity,
+                               double start_acceleration) {
   if (set_start_point && !FLAGS_use_navigation_mode) {
     InitStartPoint(start_velocity, start_acceleration);
   }
+  InitTimerAndIO();
```
vlin17, line 111: InitTimerAndIO() shall be done once while InitState can be invoked multiple times. I think it is more appropriate to move InitTimerAndIO() to Init().

modules/sim_2d/sim_2d_component.cc
```diff
@@ -100,93 +96,69 @@ void SimControl::InitTimerAndIO() {
...
-    point.mutable_path_point()->set_y(start_point.y());
-    point.mutable_path_point()->set_z(start_point.z());
-    double theta = 0.0;
-    double s = 0.0;
-    map_service_->GetPoseWithRegardToLane(start_point.x(), start_point.y(),
-                                          &theta, &s);
-    point.mutable_path_point()->set_theta(theta);
-    point.set_v(start_velocity);
-    point.set_a(start_acceleration);
-  } else {
```
vlin17, line 140: Wondering why the else statement is removed? There are use cases of simulating control from the existing ADC location.

modules/sim_2d/sim_2d_component.cc
Fla3inH0tCheet0s: Sorry. That was mistake. Bad refactor.

modules/dreamview/frontend/src/store/index.js
Fla3inH0tCheet0s: I'm having some difficulty here. It seems `this.hmi.componentStatus.has("SimControl")` always returns false.

modules/sim_2d/sim_2d_component.cc
Fla3inH0tCheet0s: I don't expect the synchronous version on initialization to work often. There is a use case where someone may want to switch from a recorded pose to a simulated pose. In a later commit, I added a callback for accepting localization.

modules/dreamview/frontend/src/store/index.js
vlin17: Oops, apologize for pointing to the wrong variable.  This will do the job: `this.hmi.moduleStatus.get("SimControl")`

modules/sim_2d/sim_2d_component.cc
vlin17: Agree, there is a use case of switching from a recorded pose to a simulated pose. From my understand, I see issues of handling this use case with the OnLocalization() in the latest commit:
1. While the component is able to get the latest recorded pose from OnLocalization(), the InitStartPoint() will ignore the SetStartPoint() from OnLocalization() and override it with the dummy point from map. Hence, the start point is always from dummy point from map.
2. This minor, but from performance perspective, we only need to get the data from localization when setting init start point. With the logic from the latest commit, we are reading the localization every 0.01 sec, which can be avoid.

lx18233184051:
please solve the conflicts@Fla3inH0tCheet0s

modules/sim_2d/sim_2d_component.cc
Fla3inH0tCheet0s: That is by design. Sorry, I didn't explain it well.

When this library was part of dreamview, it was getting localization messages in the background before sim_control was enabled. In that case, when it eventually called init there was a realistic duration where a localization message could be received: https://github.com/ApolloAuto/apollo/blob/10286356fc2c16be0b3d03e6ad6bd27220df95ce/modules/dreamview/backend/sim_control/sim_control.cc#L254. In the case where Init is called when the process just starts, there is very little time for localization messages to be received. Therefore, it is only handled asynchronously. This change also has the advantage of allowing for manually playing a bag to reset the position to a precise localization without toggling sim_control.

Fla3inH0tCheet0s:
I tried it again, but I'm seeing problems with `mainboard -d modules/planning/dag/planning.dag` unable to find `/apollo/bazel-bin/modules/planning/libplanning_component.so` and bootstrap.sh returning
```
monitor start
/apollo/scripts/apollo.bashrc: line 176: monitor: command not found
dreamview start
/apollo/scripts/apollo.bashrc: line 176: dreamview: command not found

vlin17:
We are upgrading our infra recently. Many lib changes inside the docker. Could you try if the latest docker can solve the issue?

Fla3inH0tCheet0s:
> Could you try if the latest docker can solve the issue?

`./apollo.sh build_fe` fails on master due to indentation problems. It would be nice if the webpack parts were part of the bazel build and not committed to the repo.

PR #11558, opened 2020-06-26
Planning: skipped path reference during side pass and set valid path re…

sjiang2018:
…ference counter

### DIFF
--- a/modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
+++ b/modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
@@ -44,6 +44,9 @@ using common::math::Box2d;
 using common::math::LineSegment2d;
 using common::math::Vec2d;
 
+int PathReferenceDecider::valid_path_reference_counter_ = 0;
+int PathReferenceDecider::total_path_counter_ = 0;
+
 PathReferenceDecider::PathReferenceDecider(const TaskConfig &config)
     : Task(config) {}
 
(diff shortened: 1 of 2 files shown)


modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.h
```diff
@@ -109,6 +109,10 @@ class PathReferenceDecider : public Task {
   bool IsADCBoxAlongPathReferenceWithinPathBounds(
       const std::vector<common::TrajectoryPoint> &path_reference,
       const PathBoundary *regular_path_bound);
+
+ private:
+  static int valid_path_reference_counter_;  // count valid path reference
+  static int total_path_counter_;            // count total path
```
jmtao, line 115: what are these two member variables used for? just for debugging/statistics purpose?

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
```diff
@@ -61,8 +64,20 @@ Status PathReferenceDecider::Process(Frame *frame,
     reference_line_info->mutable_path_data()->set_is_valid_path_reference(
         false);
     ADEBUG << "Skip path reference when changing lane.";
+    ADEBUG << "valid_path_reference_counter[" << valid_path_reference_counter_
+           << "] total_path_counter[" << total_path_counter_ << "]";
+    return Status::OK();
+  }
+
+  // skip using path reference during side pass
+  if (reference_line_info->is_path_lane_borrow()) {
```
jmtao, line 73: why skip?
is sidepass one of the major usages of learning path_reference?

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.h
sjiang2018: Yes. They are. For Now.

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
sjiang2018: This is for static obstacles. We were talking about moving obstacles.

jmtao: NO. path_reference from learning is supposed to work more effectively (than plain lane follow) while SidePassing for both static and moving obstacles. 
We were talking about how to use it on SidePass for moving obstacles, in order to see "improvements" compared to rule-based planner, since rule-based planner currently is unable to support SidePass for moving obstacles. 
Unfortunately the conclusion was, there's too much work at this point because our current code doesn't support it at all. So we decided to work on NUDGE on moving obstacle first (no change on code level, but more on data/model side).

jmtao: I'll merge for now. 
Need updating later.

PR #11563, opened 2020-06-26
planning: fixed a crash in path_reference_decider

jmtao:


modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.h
```diff
@@ -84,7 +84,7 @@ class PathReferenceDecider : public Task {
    * @return int
    */
 
-  int IsPointWithinPathBounds(const ReferenceLineInfo *reference_line_info,
+  int IsPointWithinPathBounds(const ReferenceLineInfo &reference_line_info,
```
sjiang2018, line 87: Why not pointer?

jmtao: google C++ coding style

sjiang2018: But there is no requirement to use constant reference at
https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs 

Instead google recommend to "Avoid defining functions that require a const reference parameter to outlive the call, because const reference parameters bind to temporaries. Instead, find a way to eliminate the lifetime requirement (for example, by copying the parameter), or pass it by const pointer and document the non-null requirement."

Looks like there is a BOOM for all of us!

PR #11564, opened 2020-06-26
Docker_dev: Fix eigen issue on one test.

xiaoxq:
This solution works for me :)
1. Firstly we need to keep @storypku 's initial try with EIGEN_MAKE_ALIGNED_OPERATOR_NEW macro, which creates special `new` operator.
2. According to https://stackoverflow.com/questions/48986131/eigen-runtime-assertion-due-to-fixed-size-members , we should NOT use `std::make_shared` to allocate such objects, because it doesn't really call `new`. Instead we should just explicitly call the `new` operator, like `std::shared_ptr<T>(new T)`

I still find lots of `make_shared` usage in perception as well as other modules. Hopefully we can fix most of this issue soon.

### DIFF
--- a/modules/perception/base/camera.h
+++ b/modules/perception/base/camera.h
@@ -46,6 +46,9 @@ class BaseCameraModel {
 
 class PinholeCameraModel : public BaseCameraModel {
  public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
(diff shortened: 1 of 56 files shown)


xiaoxq:
This change fixed `bazel test //modules/perception/base:object_pool_test`

PR #11587, opened 2020-06-28
Planning: seperated x_ref weighting for path_reference and reference …

sjiang2018:
…line.

### DIFF
--- a/modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
+++ b/modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
@@ -17,7 +17,6 @@
 #include "modules/planning/math/piecewise_jerk/piecewise_jerk_problem.h"
 
 #include "cyber/common/log.h"
-
 #include "modules/planning/common/planning_gflags.h"
 
 namespace apollo {
@@ -45,6 +44,8 @@ PiecewiseJerkProblem::PiecewiseJerkProblem(
 
   ddx_bounds_.resize(num_of_knots_,
                      std::make_pair(-kMaxVariableRange, kMaxVariableRange));
+
+  weight_x_ref_vec_ = std::vector<double>(num_of_knots_, 0.0);
 }
 
 OSQPData* PiecewiseJerkProblem::FormulateProblem() {
@@ -308,10 +309,21 @@ void PiecewiseJerkProblem::set_ddx_bounds(const double ddx_lower_bound,
 }
 
 void PiecewiseJerkProblem::set_x_ref(const double weight_x_ref,
-                                     std::vector<double> x_ref) {
+                                     const std::vector<double>& x_ref) {
   CHECK_EQ(x_ref.size(), num_of_knots_);
   weight_x_ref_ = weight_x_ref;
-  x_ref_ = std::move(x_ref);
+  weight_x_ref_vec_ = std::vector<double>(num_of_knots_, weight_x_ref);
+  x_ref_ = x_ref;
+  has_x_ref_ = true;
+}
+
+void PiecewiseJerkProblem::set_x_ref(
+    const std::vector<double>& weight_x_ref_vec,
+    const std::vector<double>& x_ref) {
+  CHECK_EQ(x_ref.size(), num_of_knots_);
+  CHECK_EQ(weight_x_ref_vec.size(), num_of_knots_);
+  weight_x_ref_vec_ = weight_x_ref_vec;
+  x_ref_ = x_ref;
   has_x_ref_ = true;
 }
 
--- a/modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
+++ b/modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
@@ -157,9 +157,9 @@ common::Status PiecewiseJerkPathOptimizer::Process(
 
     bool res_opt =
         OptimizePath(init_frenet_state.second, end_state, path_reference_l,
-                     path_boundary.delta_s(), is_valid_path_reference,
-                     path_boundary.boundary(), ddl_bounds, w, &opt_l, &opt_dl,
-                     &opt_ddl, max_iter);
+                     path_reference_size, path_boundary.delta_s(),
+                     is_valid_path_reference, path_boundary.boundary(),
+                     ddl_bounds, w, &opt_l, &opt_dl, &opt_ddl, max_iter);
 
     if (res_opt) {
       for (size_t i = 0; i < path_boundary_size; i += 4) {
@@ -230,14 +230,17 @@ PiecewiseJerkPathOptimizer::ConvertPathPointRefFromFrontAxeToRearAxe(
 bool PiecewiseJerkPathOptimizer::OptimizePath(
     const std::array<double, 3>& init_state,
     const std::array<double, 3>& end_state,
-    const std::vector<double>& path_reference_l_ref, const double delta_s,
+    const std::vector<double>& path_reference_l_ref,
+    const size_t path_reference_size, const double delta_s,
     const bool is_valid_path_reference,
     const std::vector<std::pair<double, double>>& lat_boundaries,
     const std::vector<std::pair<double, double>>& ddl_bounds,
     const std::array<double, 5>& w, std::vector<double>* x,
     std::vector<double>* dx, std::vector<double>* ddx, const int max_iter) {
-  PiecewiseJerkPathProblem piecewise_jerk_problem(lat_boundaries.size(),
-                                                  delta_s, init_state);
+  // num of knots
+  const int kNumKnots = lat_boundaries.size();
+  PiecewiseJerkPathProblem piecewise_jerk_problem(kNumKnots, delta_s,
+                                                  init_state);
 
   // TODO(Hongyi): update end_state settings
   piecewise_jerk_problem.set_end_state_ref({1000.0, 0.0, 0.0}, end_state);
@@ -256,13 +259,18 @@ bool PiecewiseJerkPathOptimizer::OptimizePath(
   }
   // use path reference as a optimization cost function
   if (is_valid_path_reference) {
-    std::vector<double> x_ref = std::move(path_reference_l_ref);
-    for (size_t i = 0; i < 10; ++i) {
+    // for non-path-reference part
+    // weight_x_ref is set to default value, where
+    // l weight = weight_x_ + weight_x_ref_ = (1.0 + 0.0)
+    std::vector<double> weight_x_ref_vec(kNumKnots, 0.0);
+    std::vector<double> x_ref = path_reference_l_ref;
+    // increase l weight for path reference part only
+    constexpr double weight_x_ref_path_reference = 1000.0;
+    for (size_t i = 0; i < path_reference_size; ++i) {
       ADEBUG << path_reference_l_ref.at(i);
+      weight_x_ref_vec.at(i) = weight_x_ref_path_reference;
     }
-
-    const double weight_x_ref = 10.0;
-    piecewise_jerk_problem.set_x_ref(weight_x_ref, x_ref);
+    piecewise_jerk_problem.set_x_ref(weight_x_ref_vec, x_ref);
   }
 
   piecewise_jerk_problem.set_weight_x(w[0]);
--- a/modules/planning/math/piecewise_jerk/piecewise_jerk_path_problem.cc
+++ b/modules/planning/math/piecewise_jerk/piecewise_jerk_path_problem.cc
@@ -14,10 +14,10 @@
  * limitations under the License.
  *****************************************************************************/
 
-#include "cyber/common/log.h"
-
-#include "modules/planning/common/planning_gflags.h"
 #include "modules/planning/math/piecewise_jerk/piecewise_jerk_path_problem.h"
+
+#include "cyber/common/log.h"
+#include "modules/planning/common/planning_gflags.h"
 
 namespace apollo {
 namespace planning {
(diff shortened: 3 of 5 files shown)


modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
```diff
@@ -308,9 +309,20 @@ void PiecewiseJerkProblem::set_ddx_bounds(const double ddx_lower_bound,
 }
 
 void PiecewiseJerkProblem::set_x_ref(const double weight_x_ref,
-                                     std::vector<double> x_ref) {
+                                     const std::vector<double>& x_ref) {
```
jmtao, line 312: if x_ref will be called by std::move(), better not define it as const&.

modules/planning/tasks/optimizers/piecewise_jerk_path/piecewise_jerk_path_optimizer.cc
```diff
@@ -256,13 +258,18 @@ bool PiecewiseJerkPathOptimizer::OptimizePath(
   }
   // use path reference as a optimization cost function
   if (is_valid_path_reference) {
+    // for non-path-reference part
+    // weight_x_ref is set to default value, where
+    // l weight = weight_x_ + weight_x_ref_ = (1.0 + 0.0)
+    std::vector<double> weight_x_ref_vec(kNumKnots, 0.0);
     std::vector<double> x_ref = std::move(path_reference_l_ref);
```
jmtao, line 265: don't std::move() on const& param

sjiang2018: Removed std::move. Thank you.

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
```diff
@@ -308,9 +309,20 @@ void PiecewiseJerkProblem::set_ddx_bounds(const double ddx_lower_bound,
...
   CHECK_EQ(x_ref.size(), num_of_knots_);
   weight_x_ref_ = weight_x_ref;
+  weight_x_ref_vec_ = std::vector<double>(num_of_knots_, weight_x_ref);
+  x_ref_ = std::move(x_ref);
+  has_x_ref_ = true;
+}
+
+void PiecewiseJerkProblem::set_x_ref(
+    const std::vector<double>& weight_x_ref_vec,
```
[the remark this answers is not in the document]
sjiang2018: Removed std::move. Thank you.

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
sjiang2018: Removed std::move. Thank you.

lx18233184051: I think you misunderstand the meaning , you need to use std::move(x_ref); not remove

sjiang2018: no need to use std::move for const reference.

jmtao: better continue to use std::move, and don't change the interface. i.e. not update it to const&.
(1)std::move has much better performance for large vector. 
(2) this is a very fundamental math Class, been used for a while. don't update its interface unless necessary.

sjiang2018: Here 'pass by constant reference' has better performance over 'pass-by-value and move'. Both have to copy once either at interface or at function internal. But 'pass-by-value and move' triggered variable 'x_ref_' deallocation before it's gotten reassigned. 

BTW, interface change doesn't affect client calling, should be fine.

sjiang2018:
> see comments

Replied.

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
jmtao: probably that's exactly what we want, whoever wrote this function initially, to release memory by using smart pointer. 
what do we achieve here by updating it like this? time complexity? space usage?

" interface change doesn't affect client calling, should be fine." When I did the first round code review on this, I had to manually check all the places this function gets called to make sure nothing is impacted (**for the original version, there's an assumption that, this vector will be "cleared" after this function call**. And this would be gone after your change). What if there are hundreds of places to check?  As I said "this is a fundamental math class working fine for a while =/> don't update the interface **unless necessary**. "should be fine" or even "can pass all test cases" is not a standard to evaluate this, this is a common sense of software engineering when you work on a relatively large project.

HongyiSun: please reference from all the other functions like set_dx_bounds()... They were all using std::move for better performance.

sjiang2018: @jmtao 
Please see my inline comments.


probably that's exactly what we want, whoever wrote this function initially, to release memory by using smart pointer.
what do we achieve here by updating it like this? time complexity? space usage?

-- Sure. But we are discussing the performance and I've justified why pass-by-const-reference is better. BTW, it's common sense to pass by reference for large structure with/without move op. 


" interface change doesn't affect client calling, should be fine." When I did the first round code review on this, I had to manually check all the places this function gets called to make sure nothing is impacted 

-- A short answer is `you don't have to inspect all client callings`. Both pass-by-constant-reference or pass-by-value avoid modifying values of parameters outside of function. 

(for the original version, there's an assumption on the caller side that, this vector will be "cleared" after this function call because we used std::move. And this assumption would be gone after your change without the callers being aware of). 

-- This assumption only makes sense when dealing with set_xxx functions for memory copy. But in general, it's not wise to clean memory implicitly by calling an external API from caller side, i.e, doing secret thing in one function, which should be avoided in software design. 

What if there are hundreds of places to check? As I said "this is a fundamental math class working fine for a while => don't update the interface unless NECESSARY. "should be fine" or even "can pass all test cases" is not a standard to evaluate this, this is a common sense of software engineering when you work on a relatively large project.

-- Discussion is good and it helps us to reach the best solution. Once the best known method is reached, we can make some efforts to correct error usages in current projects.

sjiang2018: > please reference from all the other functions like set_dx_bounds()... They were all using std::move for better performance.

@HongyiSun 

Please see my previous comments. 

Here are examples of similar usages: https://github.com/ApolloAuto/apollo/blob/be8acafb4aebc799ffed0467cce97de3a82fbb04/modules/planning/common/obstacle.cc#L697

https://github.com/ApolloAuto/apollo/blob/be8acafb4aebc799ffed0467cce97de3a82fbb04/modules/planning/math/discretized_points_smoothing/fem_pos_deviation_osqp_interface.h#L42

sjiang2018:
> What is the point to add a total of N weight parameters?
> Who is going to tune those N (maybe 300) weights?
> Also, what is the physical meaning for different reference weight along each position?

To make the optimization more configurable. 
The parameters can be set to a same value or certain piece-wise values if necessary.

sjiang2018:
> see comments

Replied.

HongyiSun:
> what is the physical meaning for different reference weight along each position?

HongyiSun:
> > What is the point to add a total of N weight parameters?
> > Who is going to tune those N (maybe 300) weights?
> > Also, what is the physical meaning for different reference weight along each position?
> 
> To make the optimization more configurable.
> The parameters can be set to a same value or certain piece-wise values if necessary.

What is the physical meaning for different reference weight along each position?

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
xiaoxq: Wow, really helpful discussion!

My two cents:
There are two moves here. The caller -> The function parameter -> The final holder.

Current way, passing by `std::vector<double>` and assign with `std::move`, will force the second move as `steal`, but the first move might be `copy` or `steal`, totally based on the user's expertise.
Pros: Have a chance to make 0 copy.
Cons: Ambiguous on behavior and performance. Have a chance to make 2 copies if the caller and callee are not experienced and careful.

The proposed change, passing by `const std::vector<double>&`, will force the first move as `steal`, but the second move is sure to be a copy.
Pros: It's actually the recommended coding style and best practice. Very clear on API and behavior.
Cons: Sure to make 1 copy.

I would suggest a modern `rvalue` parameter which would guarantee 0 copy:
http://www.rovrov.com/blog/2014/12/19/pass-by-rvalue-reference-or-pass-by-value/

Also CC system team for a look.
@storypku @lfcarol @lx18233184051 @changsh726

sjiang2018:
> > > What is the point to add a total of N weight parameters?
> > > Who is going to tune those N (maybe 300) weights?
> > > Also, what is the physical meaning for different reference weight along each position?
> > 
> > 
> > To make the optimization more configurable.
> > The parameters can be set to a same value or certain piece-wise values if necessary.
> 
> What is the physical meaning for different reference weight along each position?



> > > What is the point to add a total of N weight parameters?
> > > Who is going to tune those N (maybe 300) weights?
> > > Also, what is the physical meaning for different reference weight along each position?
> > 
> > 
> > To make the optimization more configurable.
> > The parameters can be set to a same value or certain piece-wise values if necessary.
> 
> What is the physical meaning for different reference weight along each position?

 The physical meaning is we trust certain points more than other points. We discussed in a group meeting, one option of learning model output is a long trajectory with sparse trajectory points. In that case, we trust the points generated from the learning model more over the points from interpolation.

HongyiSun:
> > > > What is the point to add a total of N weight parameters?
> > > > Who is going to tune those N (maybe 300) weights?
> > > > Also, what is the physical meaning for different reference weight along each position?
> > > 
> > > 
> > > To make the optimization more configurable.
> > > The parameters can be set to a same value or certain piece-wise values if necessary.
> > 
> > 
> > What is the physical meaning for different reference weight along each position?
> 
> > > > What is the point to add a total of N weight parameters?
> > > > Who is going to tune those N (maybe 300) weights?
> > > > Also, what is the physical meaning for different reference weight along each position?
> > > 
> > > 
> > > To make the optimization more configurable.
> > > The parameters can be set to a same value or certain piece-wise values if necessary.
> > 
> > 
> > What is the physical meaning for different reference weight along each position?
> 
> The physical meaning is we trust certain points more than other points. We discussed in a group meeting, one option of learning model output is a long trajectory with sparse trajectory points. In that case, we trust the points generated from the learning model more over the points from interpolation.

So how do you tune those parameters, do you trust them with same confidence? 
And it back to the problem that how do you convince about the confidence here, why does the reference in each point better than the central lane, should there be larger weight for better points and smaller weight for those not as good as lane central?

modules/planning/math/piecewise_jerk/piecewise_jerk_problem.cc
jmtao: I agreed it's recommended coding style adopted by us probably from the very beginning. 
my point is , for this particular case, PLEASE don't update the interface of a **fundamental** **MATH** class been in use for a while , unless NECESSARY. 
And if you do, please update them all for consistency.  not just a random one you happen to touch. 

I'm out of here. have fun.

HongyiSun: I agree the idea to pass by value and keep the std::move inside the function, however, there should be a change on the calling of this function, where `piecewise_jerk_problem.set_x_ref(weight_x_ref_vec, x_ref);` in `piecewise_jerk_path_optimizer.cc` should be change into something like `piecewise_jerk_problem.set_x_ref(weight_x_ref_vec, std::move(x_ref));` And that's the way with not a single copy on the whole process.

jmtao: completely agree!
that's an update I think worth doing.

xiaoxq: Yeah, if we change the API to `std::vector<double>&& x_ref`, the caller must pass parameter with `std::move`. So the 0-copy practice is well guaranteed.

jmtao: “We can do this refactor in future.” then don't make the change in this PR at this point. Keep all function in sync. 
do it well or not at all.  period

sjiang2018: Looks like we already reached the best solution, I agree with @xiaoxq to change the API to std::vector<double>&& x_ref. I don't mind following current interface style, but it surely should be refactored enforcing "move" call from caller side in future in case 0 copy is targeted.

sjiang2018:
> > > > > What is the point to add a total of N weight parameters?
> > > > > Who is going to tune those N (maybe 300) weights?
> > > > > Also, what is the physical meaning for different reference weight along each position?
> > > > 
> > > > 
> > > > To make the optimization more configurable.
> > > > The parameters can be set to a same value or certain piece-wise values if necessary.
> > > 
> > > 
> > > What is the physical meaning for different reference weight along each position?
> > 
> > 
> > > > > What is the point to add a total of N weight parameters?
> > > > > Who is going to tune those N (maybe 300) weights?
> > > > > Also, what is the physical meaning for different reference weight along each position?
> > > > 
> > > > 
> > > > To make the optimization more configurable.
> > > > The parameters can be set to a same value or certain piece-wise values if necessary.
> > > 
> > > 
> > > What is the physical meaning for different reference weight along each position?
> > 
> > 
> > The physical meaning is we trust certain points more than other points. We discussed in a group meeting, one option of learning model output is a long trajectory with sparse trajectory points. In that case, we trust the points generated from the learning model more over the points from interpolation.
> 
> So how can we tune those parameters, shall we trust them with same confidence?
> And it back to the problem that how can we convince about the confidence here. I'm wondering if the reference in each point is strictly better than the central lane? If not, should there be some larger weight for those better points and some smaller weight for those not as good as lane central?

"Do we need individual weightings?" and "How we tune those weightings?" are different questions. For the first one, see my previous comments. For the second one, there are some existing frameworks to do that and we can discuss it offline if you are interested.

Capri2014:
Let's put this on hold and move the more detailed discuss to offline.

PR #11619, opened 2020-06-30
docker_dev: apollo_repositories : add glew ffmpeg etc...

lx18233184051:


tools/workspace.bzl
```diff
@@ -29,8 +33,12 @@ def initialize_third_party():
     boost()
     eigen()
     fastrtps()
+    ffmpeg()
     gflags()
+    glew()
```
storypku, line 38: GLEW is not used across the whole Apollo project. So we should disable it in `tools/workspace.bzl`.

tools/workspace.bzl
```diff
@@ -5,8 +5,12 @@ load("//third_party/adv_plat:workspace.bzl", adv_plat = "repo")
 load("//third_party/boost:workspace.bzl", boost = "repo")
 load("//third_party/eigen3:workspace.bzl", eigen = "repo")
 load("//third_party/fastrtps:workspace.bzl", fastrtps = "repo")
+load("//third_party/ffmpeg:workspace.bzl", ffmpeg = "repo")
 load("//third_party/gflags:workspace.bzl", gflags = "repo")
+load("//third_party/glew:workspace.bzl", glew = "repo")
```
storypku, line 10: GLEW should be commented out as not used.

PR #11630, opened 2020-07-01
Planning: add cpu models and a configuration on rule-based planning to learning-based transition 

JasonZhou404:
Only one set of models is uploaded. Will reduce the updating frequency and number of models in the future commit to minimize affect to git history


lx18233184051:
thanks：）

PR #11635, opened 2020-07-01
Docker: Add utils for image release.

xiaoxq:


env.json
```diff
@@ -0,0 +1,4 @@
+{
+    "OS": "Ubuntu 18.04",
```
storypku, line 2: Better split the OS into something like `{"distro": "Ubuntu", "release":"18.04", "codename": "bionic"}`

```
$ lsb_release -si
Ubuntu
$ lsb_release -sr
18.04
$ lsb_release -sc
bionic
```

xiaoxq: We will not maintain too many versions. I would prefer to make it simple.
Currently the only user is simulation. Let's ask how they want to use it.

storypku: > We will not maintain too many versions. I would prefer to make it simple.
> Currently the only user is simulation. Let's ask how they want to use it. @vlin17 @unacao
I See. My point is,  if one day Apollo come out of docker and run on a variety of Linux distros,  canonial names as indicated by `lsb_release` can make env parsers simpler.

unacao: If it is in parallel with "Python", maybe we should keep "OS" as outer tag, while adding what Jiaming suggested as inner tags? i.e.
{
  "OS": {
    "distro": "Ubuntu",
    "release": "18.04"
  }
}

xiaoxq: Done. Thank you!

PR #11638, opened 2020-07-02
docker: add python dependencies into docker image

changsh726:


docker/build/installers/install_python_modules.sh
```diff
@@ -29,8 +29,7 @@ apt update -y && apt install -y \
     python3-pip \
     python3-psutil \
     python3-scipy \
-    software-properties-common \
-    libasound2-dev
```
storypku, line 33: So, libasound is no longer required ?

docker/build/installers/install_python_modules.sh
```diff
@@ -29,8 +29,7 @@ apt update -y && apt install -y \
     python3-pip \
```
storypku, line 29: python3-pip has been installed and upgraded in the `install_minimal_environment.sh` section.  Please remove it here. Besides, better install pthon3-psutil and python3-scipy stuff via pip3_install, or else there will be both system provided python modules (psutil, scipy, and their dependencies) and pip3 installed versions specified in `py3_requirements.txt`.

docker/build/installers/install_python_modules.sh
changsh726: libasound is installed as a dependency of pyalsaaudio however, I haven't found where the pyalsaaudio is used

docker/build/installers/install_python_modules.sh
changsh726: done

PR #11639, opened 2020-07-02
D-KIT: Update two configuration files of D-KIT

diegohu:
1. Add canbus.conf in D-KIT configuration files set
2. D-KIT only support MSF localization mode. As some dreamview hmi mode(RTK, Vehicle calibration) use RTK mode, we need to overwrite the dag with MSF mode.


modules/calibration/data/dev_kit/cancard_params/canbus.conf
```diff
@@ -0,0 +1,4 @@
+--flagfile=/apollo/modules/common/data/global_flagfile.txt
+--canbus_conf_file=/apollo/modules/canbus/conf/canbus_conf.pb.txt
+--noenable_chassis_detail_pub
```
macDure, line 3: Suggest enable_chassis_detail_pub for dkit, because we always suggest developer open the chassi_detail in our 
guide spec.

PR #11647, opened 2020-07-02
Master: Clear static class and function.

lfcarol:


modules/prediction/common/dependency_injector.h
```diff
@@ -0,0 +1,37 @@
...
+ *****************************************************************************/
+
+#pragma once
+
+#include "modules/prediction/common/junction_analyzer.h"
+
+namespace apollo {
+namespace prediction {
+
+class DependencyInjector {
```
changsh726, line 24: I don't think we need an extra DependencyInjector class to wrap the JunctionAnalyzer

lfcarol:
> I think the JunctionAnalyzer instance can be created in ObsaclesContainer rather than in PredictionComponent, which saving a lot of changes.

I'm afraid there is something relative among ObsaclesContainers.

changsh726:
> > I think the JunctionAnalyzer instance can be created in ObsaclesContainer rather than in PredictionComponent, which saving a lot of changes.
> 
> I'm afraid there is something relative among ObsaclesContainers.

According to my observation, there is only one instance of ObstaclesContainer with key of PERCEPTION_OBSTACLES in prediction module. You can have a double check.

PR #11680, opened 2020-07-06
planning: delete lattice_sampling_config.proto not in use

jmtao:


jinghaomiao:
@jmtao lint check failed.

PR #11695, opened 2020-07-07
Build: fix eigen alignment issues in localization

storypku:
Ref: http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html
Co-authored with Liu Xu

### DIFF
--- a/modules/common/util/BUILD
+++ b/modules/common/util/BUILD
@@ -180,4 +180,10 @@ cc_test(
     ],
 )
 
+cc_library(
+    name = "eigen_defs",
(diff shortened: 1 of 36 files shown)


modules/common/util/BUILD
```diff
@@ -180,4 +180,10 @@ cc_test(
     ],
 )
 
+cc_library(
+    name = "eigen_defs",
+    hdrs = ["eigen_defs.h"],
+    deps = ["@eigen"],
+)
```
xiaoxq, line 187: Nice wrap!

PR #11729, opened 2020-07-09
master: teleop bazel configuration

lfcarol:


tools/apollo.bazelrc.sample
```diff
@@ -33,4 +33,4 @@ build:cuda_clang --action_env TF_CUDA_CLANG=1
 build:tensorrt --action_env TF_NEED_TENSORRT=1
 build:nonccl --define=no_nccl_support=true
 
-build:teleop --define WITH_TELEOP=1
+build:teleop --action_env WITH_TELEOP=1
```
storypku, line 36: Better not delete the `--define` line. There is a section in `tools/platfor/BUILD` that depends on this defination.
```
def if_with_teleop(if_true, if_false = []):
    return select({
        "//tools/platform:with_teleop": if_true,
        "//conditions:default": if_false,
    })

def copts_if_with_teleop():
    return if_with_teleop(["-DTELEOP=1"])
```

PR #11737, opened 2020-07-10
Perception: down sample point cloud through filtering beams

jeroldchen:


modules/perception/base/point_cloud.h
```diff
@@ -451,6 +451,9 @@ class AttributePointCloud : public PointCloud<PointT> {
 
   const std::vector<int32_t>& points_beam_id() const { return points_beam_id_; }
   std::vector<int32_t>* mutable_points_beam_id() { return &points_beam_id_; }
+  int32_t& points_beam_id(size_t i) { return points_beam_id_[i]; }
```
storypku, line 454: What's the point of returning a refenrce?

jeroldchen: I don't understand either ... Just want to keep consistent with other methods in format.

jinghaomiao: @storypku good find. @jeroldchen I will merge this pull request. Please evaluate if there's a need to rewrite all functions where a better return format is needed together with @storypku and @changsh726 .

PR #11743, opened 2020-07-10
planning: follow bazel's suggestion on timeout=short or size=small

storypku:


scripts/apollo.bashrc
```diff
@@ -90,10 +90,12 @@ function determine_gpu_use() {
 
     if [[ "${arch}" == "aarch64" ]]; then
         if lsmod | grep -q nvgpu; then
-            use_gpu=1
+            if ldconfig -p | grep -q cudart; then
```
storypku, line 93: Check that nvgpu kmod and cudart library exist.

PR #11751, opened 2020-07-12
check "xxx:_proto" deps for python_rules.bzl

storypku:


cyber/python/cyber_py3/test/BUILD
```diff
@@ -1,5 +1,8 @@
+load("@rules_python//python:defs.bzl", "py_test")
```
xiaoxq, line 1: It seems like a built-in function of bazel, just like cc_test.
Any reason to use the one from `rules_python`?

storypku: > It seems like a built-in function of bazel, just like cc_test.
> Any reason to use the one from `rules_python`?

For higher bazel versions, `cc_test`, `py_test` were not builtin anymore,  and  the `load(...)` line was added by `buildifier`.

PR #11752, opened 2020-07-12
Bugfix: modules/canbus/proto/BUILD updated via proto_build_generator

storypku:


modules/canbus/proto/BUILD
```diff
@@ -5,6 +5,24 @@ load("//tools:python_rules.bzl", "py_proto_library")
 
 package(default_visibility = ["//visibility:public"])
 
+cc_proto_library(
```
storypku, line 8: This BUILD file  was generated by running `scripts/proto_build_generator.py modules/canbus/proto/BUILD`

lx18233184051:
pretty good, I just want to fix this bug~~

PR #11768, opened 2020-07-13
Perception:make lidar height file could be specified

SeasoulChris:
1、make lidar height file could be specificed throught fusion_camera_detection_component.pb.txt
2、add fusion_camera_detection_component.pb.txt to modules/calibration/data/dev_kit and set default value according to dev_kit params


modules/dreamview/conf/vehicle_data.pb.txt
```diff
@@ -138,3 +138,7 @@ data_files {
   source_path: "perception_launch"
   dest_path: "/apollo/modules/perception/production/launch"
 }
+data_files {
+  source_path: "fusion_camera_detection_component_conf"
```
diegohu, line 142: suggest use “perception_conf” as the source path

PR #11788, opened 2020-07-14
Perception: update build and launch for lidar perception only

jeroldchen:


modules/perception/base/hdmap_struct.h
```diff
@@ -35,10 +37,17 @@ struct alignas(16) LaneBoundary {
 };
 
 struct alignas(16) HdmapStruct {
-  std::vector<RoadBoundary> road_boundary;
-  std::vector<PointCloud<PointD>> road_polygons;
-  std::vector<PointCloud<PointD>> hole_polygons;
-  std::vector<PointCloud<PointD>> junction_polygons;
+  std::vector<RoadBoundary, Eigen::aligned_allocator<RoadBoundary> >
```
storypku, line 40: Since std::vector<A, Eigen::aligned_allocator<A>> are so wildly used in pept, maybe it's better to define some general data structure or typedefs for simplicity?

xiaoxq: Good point. Added EigenVector with https://github.com/ApolloAuto/apollo/pull/11791/files

jinghaomiao: Perception team will discuss them intrnally and submit a separate pull request to address these issues. @jeroldchen @kechxu

PR #11789, opened 2020-07-14
Cyber : fix warning when build cyber

lfcarol:


cyber/message/message_traits_test.cc
```diff
@@ -34,10 +34,10 @@ class Message {
  public:
   std::string content;
 
-  int ByteSize() const { return static_cast<int>(content.size()); }
+  int ByteSizeLong() const { return static_cast<int>(content.size()); }
```
xiaoxq, line 37: It looks confusing to me that `ByteSizeLong` returns `int` instead of `long`. Any reason to change the name?

xiaoxq:
BTW, the commit and pull request message pattern should be "<Module>: XXX", instead of "<Branch>: XXX".
Previously we use "docker_dev: XXX" just indicate others that it's some work on infra.

cyber/message/message_traits_test.cc
lfcarol: > It looks confusing to me that `ByteSizeLong` returns `int` instead of `long`. Any reason to change the name?

Because the protobuf v3.12.3 "ByteSize()" function have a compilation attribute "deprecated", it can warn us the function is out of date and recommand to use "ByteSizeLong()", which is different from the protobuf v3.3.0. So I use the "ByteSizeLong()" to match the template function "cyber/message/message_traits.h: ByteSize()".

PR #11804, opened 2020-07-15
Planning: using rule-based planner when learning model output is not ready

sjiang2018:


modules/planning/tasks/learning_model/learning_model_inference_task.cc
```diff
@@ -73,6 +73,11 @@ Status LearningModelInferenceTask::Process(Frame* frame) {
         absl::StrCat("learning_data adc_trajectory_point empty. frame_num[",
                      learning_data_frame.frame_num(), "]");
     AERROR << msg;
+    // hybrid model will use rule based planning when learning model output is
+    // not ready
+    if (FLAGS_planning_learning_mode == 3) {
```
jmtao, line 78: tasks are supposed to be independent of "context", and only takes "parameter" type of gflag(s). so that they can be easily added to any scenario/stage.
tasking context as condition may introduce unexpected behavior, and unnecessary debug effort.

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
```diff
@@ -91,6 +91,12 @@ Status PathReferenceDecider::Process(Frame *frame,
       frame->learning_based_data().learning_data_adc_future_trajectory_points();
   ADEBUG << "There are " << path_reference.size() << " path points.";
 
+  // when learning model has no output, use rule-based model instead.
+  if (path_reference.size() == 0) {
```
jmtao, line 95: shall we expose model failure(s), instead of hiding them?

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
```diff
@@ -91,6 +91,12 @@ Status PathReferenceDecider::Process(Frame *frame,
...
+    ADEBUG << "No learning model output, using rule-based planning instead";
+    return Status::OK();
+  }
+
```
jmtao, line 99: total_path_counter_ shall be moved ahead of this.
(1) it's OK not to count frames skip this decider on purpose (side pass, lane change and etc). other than that, they shall be counted to total.
(2) or to be more accurate, count all frames. so that we know the exact ratio of valid path reference. 
(3) or provide two ratio.  valid/all_frames,   valid /frames_calling_model

modules/planning/tasks/learning_model/learning_model_inference_task.cc
sjiang2018: Fixed. Thank you.

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
sjiang2018: Fixed in the next pr. Thank you.

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
sjiang2018: Fixed in the next commit. Thank you.

PR #11808, opened 2020-07-16
Support USE_SIM_TIME mode in CyberRT

rongguodong:
ROS can use simulation time from /clock topic (http://wiki.ros.org/Clock).

This PR adds the same ability to CyberRT. It is an important feature to run simulator together with Apollo.

Starting Apollo docker with "dev_start.sh -s", CyberRT will try to use value from /clock as Now(), and Apollo will use CyberRT time (which is from /clock).

### DIFF
--- a/cyber/BUILD
+++ b/cyber/BUILD
@@ -7,7 +7,7 @@ cc_library(
     name = "cyber",
     linkstatic = False,
     deps = [
-        "//cyber:cyber_core",
+        ":init",
     ],
 )
 
@@ -23,7 +23,7 @@ cc_binary(
     linkopts = ["-pthread"],
     linkstatic = False,
     deps = [
-        ":cyber_core",
+        ":init",
         "//cyber/proto:dag_conf_cc_proto",
     ],
 )
@@ -47,11 +47,13 @@ cc_library(
     srcs = ["init.cc"],
     hdrs = ["init.h"],
     deps = [
+        "//cyber:cyber_core",
         "//cyber:state",
         "//cyber/logger:async_logger",
         "//cyber/node",
         "//cyber/sysmo",
         "//cyber/timer:timing_wheel",
+        "//cyber/proto:clock_cc_proto",
     ],
 )
 
@@ -62,7 +64,6 @@ cc_library(
     linkopts = ["-lrt"],
     deps = [
         "//cyber:binary",
-        "//cyber:init",
         "//cyber:state",
         "//cyber/base",
         "//cyber/blocker:blocker_manager",
--- a/cyber/proto/BUILD
+++ b/cyber/proto/BUILD
@@ -349,3 +349,22 @@ py_proto_library(
     ],
 )
 
+cc_proto_library(
+    name = "clock_cc_proto",
+    deps = [
+        ":clock_proto",
+    ],
+)
+
+proto_library(
+    name = "clock_proto",
+    srcs = ["clock.proto"],
+)
+
+py_proto_library(
+    name = "clock_py_pb2",
+    deps = [
+        ":clock_proto",
+    ],
+)
+
--- a/cyber/time/time.cc
+++ b/cyber/time/time.cc
@@ -30,6 +30,9 @@ using std::chrono::high_resolution_clock;
 using std::chrono::steady_clock;
 using std::chrono::system_clock;
 
+static bool g_use_sim_time(false);
+static Time g_sim_time(0);
+
 const Time Time::MAX = Time(std::numeric_limits<uint64_t>::max());
 const Time Time::MIN = Time(1);
 
@@ -54,7 +57,14 @@ Time& Time::operator=(const Time& other) {
   return *this;
 }
 
+void Time::SetSimTime(const Time& sim_time) {
+  g_sim_time = sim_time;
+  g_use_sim_time = true;
+}
+
 Time Time::Now() {
+  if (g_use_sim_time) return g_sim_time;
+
   auto now = high_resolution_clock::now();
   auto nano_time_point =
       std::chrono::time_point_cast<std::chrono::nanoseconds>(now);
--- a/docker/scripts/dev_start.sh
+++ b/docker/scripts/dev_start.sh
@@ -181,6 +181,10 @@ while [ $# -gt 0 ] ; do
         ;;
     -y)
         ;;
+    -s|--use_sim_time)
+	USE_SIM_TIME=true
+	info "CyberRT uses simulation time if available"
+	;;
     stop)
         stop_containers
         exit 0
@@ -410,6 +414,7 @@ function main() {
         -e USE_GPU=$USE_GPU \
         -e NVIDIA_VISIBLE_DEVICES=all \
         -e NVIDIA_DRIVER_CAPABILITIES=compute,video,graphics,utility \
+	-e USE_SIM_TIME=$USE_SIM_TIME \
         $(local_volumes) \
         --net host \
         -w /apollo \
--- a/modules/common/time/time.h
+++ b/modules/common/time/time.h
@@ -26,14 +26,14 @@
 #include <atomic>
 #include <chrono>
 #include <stdexcept>
+#include <string>
 #include <type_traits>
 
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
+#include "cyber/common/log.h"
 #include "cyber/common/macros.h"
 #include "cyber/time/time.h"
-
-#include "cyber/common/log.h"
 #include "modules/common/configs/config_gflags.h"
 
 /**
@@ -113,7 +113,7 @@ class Clock {
    * @brief This is for mock clock mode only. It will set the timestamp
    * for the mock clock.
    */
-  static void SetNow(const absl::Time &now) {
+  static void SetNow(const absl::Time& now) {
     auto clock = Instance();
     if (clock->mode_ != ClockMode::MOCK) {
       AFATAL << "Cannot set now when clock mode is not MOCK!";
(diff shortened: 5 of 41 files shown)


modules/common/time/time.h
```diff
@@ -147,6 +147,14 @@ class Clock {
 
 inline Clock::Clock() {
   mode_ = FLAGS_use_cyber_time ? ClockMode::CYBER : ClockMode::SYSTEM;
+  // Force to use Cyber time is USE_SIM_TIME is set
+  const char* use_sim_time = ::getenv("USE_SIM_TIME");
```
storypku, line 151: Please use `apollo::cyber::common::GetEnv` from `cyber/common/environment.h` be to consistent.

modules/common/time/time.h
```diff
@@ -147,6 +147,14 @@ class Clock {
...
+  if (use_sim_time != nullptr) {
+    std::string use_sim_time_str(use_sim_time);
+    if (use_sim_time_str == "true") {
```
storypku, line 154:
```
export USE_SIM_TIME=1
export USE_SIM_TIME=true
```
The two forms listed above should both be OK.

cyber/BUILD
```diff
@@ -7,7 +7,7 @@ cc_library(
     name = "cyber",
     linkstatic = False,
     deps = [
-        "//cyber:cyber_core",
+        ":init",
```
storypku, line 10: Better find a way to keep the original deps targets while add the functionality in your PR.
`:init` seems like a local target name which should not have public visibility. ( I must admit that apollo has poor visibility control for now...)

modules/map/tools/map_datachecker/server/worker.cc
```diff
@@ -17,9 +17,9 @@
 
 #include <memory>
 
-#include "grpc++/grpc++.h"
-
 #include "cyber/cyber.h"
+#include "cyber/init.h"
+#include "grpc++/grpc++.h"
```
storypku, line 22: header orders.

modules/planning/integration_tests/planning_test_base.h
```diff
@@ -18,6 +18,7 @@
 #include <memory>
 #include <string>
 
+#include "cyber/init.h"
```
storypku, line 21: header orders.

cyber/time/time.cc
```diff
@@ -54,7 +57,14 @@ Time& Time::operator=(const Time& other) {
   return *this;
 }
 
+void Time::SetNow(const Time& new_now) {
```
storypku, line 60: Better rename this to `SetSimTimeNow` to be more specific.

docker/scripts/dev_start.sh
```diff
@@ -410,6 +414,7 @@ function main() {
         -e USE_GPU=$USE_GPU \
         -e NVIDIA_VISIBLE_DEVICES=all \
         -e NVIDIA_DRIVER_CAPABILITIES=compute,video,graphics,utility \
+	-e USE_SIM_TIME=$USE_SIM_TIME \
```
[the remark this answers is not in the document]
rongguodong: I want to have a control of using/not-using sim_time. If users do not want this feature, they can use dev_start without "-s".
If I add "export USE_SIM_TIME=true" in `cyber/setup.bash`, how can I control it via an argument?

modules/map/relative_map/tools/navigator.cc
```diff
@@ -19,17 +19,17 @@
...
 #include "cyber/common/log.h"
 #include "cyber/cyber.h"
+#include "cyber/init.h"
 #include "cyber/time/rate.h"
 #include "modules/common/adapters/adapter_gflags.h"
 #include "modules/common/util/message_util.h"
 #include "modules/map/relative_map/common/relative_map_gflags.h"
 #include "modules/map/relative_map/proto/navigation.pb.h"
 #include "modules/map/relative_map/proto/navigator_config.pb.h"
+#include "nlohmann/json.hpp"
```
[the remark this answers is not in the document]
rongguodong: When I hit Ctrl-Shift-I in VSCode, it automatically changed the header orders as this. Do you have a config file to have VSCode following Apollo rule?
Also, this code passed lint test. Can you have lint test to test the correctness of header orders?

cyber/BUILD
rongguodong: I also want to not change so many BUILD files. Do you have better ideas on how to make it simpler? I can change it accordingly.

modules/map/relative_map/tools/navigator.cc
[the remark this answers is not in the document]
storypku: > When I hit Ctrl-Shift-I in VSCode, it automatically changed the header orders as this. Do you have a config file to have VSCode following Apollo rule?
> Also, this code passed lint test. Can you have lint test to test the correctness of header orders?

It's really strang to me that your header orders  passed cpplint check. IIRC, previously when we do `master` upgrade  in `docker_dev` branch, `cpplint.py` complains a lot about this.

I will do an investigation to see why and figure out a solution.

cyber/BUILD
storypku: > I also want to not change so many BUILD files. Do you have better ideas on how to make it simpler? I can change it accordingly.

Maybe,  fine-grained header includes ? so that `:init` can depend on specific targets rather than `cyber_core`.
I will work on fixing this  with you. Hopes we can resolve it before next Monday.

cyber/proto/BUILD
```diff
@@ -349,3 +349,22 @@ py_proto_library(
...
+    ],
+)
+
+proto_library(
+    name = "clock_proto",
+    srcs = ["clock.proto"],
+)
+
+py_proto_library(
+    name = "clock_pb2",
```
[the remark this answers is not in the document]
rongguodong: Changed.

cyber/time/time.cc
rongguodong: I renamed it to SetSimTime and added comments for it.

modules/map/relative_map/tools/navigator.cc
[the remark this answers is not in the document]
xiaoxq: > When I hit Ctrl-Shift-I in VSCode, it automatically changed the header orders as this. Do you have a config file to have VSCode following Apollo rule?
> Also, this code passed lint test. Can you have lint test to test the correctness of header orders?

Third party headers should be in it's own section:
```
#include <cmath>

#include "nlohmann/json.hpp"

#include "other apollo headers"
```

Auto-format tools do have some defects.

modules/common/time/time.h
xiaoxq: I'm against ENV-based behavior which is very hard to debug. Prefer FLAG.

rongguodong: I did intend to use gflag, particularly since Apollo already has a flag named "use_cyber_time". But I did not figure out a way to set it to true (its default value is false as declared[ here](https://github.com/ApolloAuto/apollo/blob/master/modules/common/configs/config_gflags.cc#L46). Can you tell me how to change this gflag? I would like to dynamically set it via a command line argument, so that we can switch among different modes.

lx18233184051:
maybe cyber/examples/common_component_example/channel_prediction_writer.cc:28  has some problems?

storypku:
> maybe cyber/examples/common_component_example/channel_prediction_writer.cc:28 has some problems?

?

cyber/time/time.cc
```diff
@@ -54,7 +57,14 @@ Time& Time::operator=(const Time& other) {
   return *this;
 }
 
+void Time::SetSimTime(const Time& sim_time) {
+  g_sim_time = sim_time;
+  g_use_sim_time = true;
+}
+
 Time Time::Now() {
+  if (g_use_sim_time) return g_sim_time;
```
xiaoxq, line 66: Add `{}`

cyber/time/time.cc
```diff
@@ -54,7 +57,14 @@ Time& Time::operator=(const Time& other) {
   return *this;
 }
 
+void Time::SetSimTime(const Time& sim_time) {
+  g_sim_time = sim_time;
+  g_use_sim_time = true;
```
xiaoxq, line 62: Should these be protected for thread safety?

modules/common/time/time.h
xiaoxq: Simply `--use_cyber_time` will set it to true.

modules/map/relative_map/tools/navigator.cc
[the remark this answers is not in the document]
xiaoxq: gentle ping

modules/common/time/time.h
rongguodong: Using --use_cyber_time with what command? "bootstraph_lgsvl.sh --use_cyber_time", "cyber_launch start xxx/xxx.launch --use_cyber_time", or something else?

xiaoxq: We use
https://github.com/ApolloAuto/apollo/blob/master/modules/common/data/global_flagfile.txt
to config shared flags across modules, if their own flagfiles inherit from the global one correctly: https://github.com/ApolloAuto/apollo/blob/master/modules/control/conf/control.conf#L1

rongguodong: But "use_cyber_time" is already set to "true" in global_flagfile. However, I still see it is false in the code. This is why I am asking hot to set it. This global_flagfile seems not working?

lx18233184051:
this pr has been so long~ maybe we should close, you can reopen if needed? @rongguodong

rongguodong:
> this pr has been so long~ maybe we should close, you can reopen if needed? @rongguodong

With PR 11894 and PR 11841 merged, most of contents of this PR are already in. But please finish the changes to Apollo::Time as well. Then this PR is no longer needed.

storypku:
> > this pr has been so long~ maybe we should close, you can reopen if needed? @rongguodong
> 
> With [PR 11894](https://github.com/ApolloAuto/apollo/pull/11894) and [PR 11841](https://github.com/ApolloAuto/apollo/pull/11841) merged, most of contents of this PR are already in. But please finish the changes to Apollo::Time as well (as offline discussion with @storypku). Then this PR is no longer needed.

OK.

PR #11829, opened 2020-07-17
Build: build/include_order fix for cpplint

storypku:


cyber/timer/timer_test.cc
```diff
@@ -47,20 +47,22 @@ TEST(TimerTest, one_shot) {
 }
 
 TEST(TimerTest, cycle) {
+  using TimerPtr = std::shared_ptr<Timer>;
   int count = 0;
-  Timer timers[1000];
+  TimerPtr timers[1000];
```
lx18233184051, line 52: Good

PR #11831, opened 2020-07-17
Perception: enable launching perception module from dreamview

jeroldchen:


modules/perception/onboard/component/BUILD
```diff
@@ -15,12 +15,14 @@ cc_library(
     srcs = [
         "detection_component.cc",
         "fusion_component.cc",
+        "radar_detection_component.cc",
```
xiaoxq, line 18: Please split to thinner targets instead of a compound one which is very hard to maintain.

jinghaomiao: @jeroldchen split each cc file into a separate build target.

PR #11845, opened 2020-07-18
Scripts: renamed to clang_format.sh

storypku:


docker/scripts/dev_start.sh
```diff
@@ -30,7 +30,7 @@ LOCAL_IMAGE="no"
 FAST_BUILD_MODE="no"
 FAST_TEST_MODE="no"
 VERSION=""
-VERSION_X86_64="dev-x86_64-18.04-20200714_0111"
+VERSION_X86_64="dev-x86_64-18.04-20200718_0221"
```
lx18233184051, line 33: Our image update frequence could be  a little slow ?

storypku: > Our image update frequence could be a little slow ?

Just bugfix on top of previous image.

PR #11853, opened 2020-07-19
Scripts: allows proto BUILD file to be generated outside of apollo

changsh726:


scripts/proto_build_generator.py
```diff
@@ -347,8 +341,7 @@ def generate_dependency_text(workdir, protofile):
```
storypku, line 341: the `deps` section should be changed accordingly.

PR #11862, opened 2020-07-20
Docker: fix `gcc internal error` when build aarch64 on tx2

storypku:


docker/build/installers/install_vtk.sh
```diff
@@ -48,8 +50,12 @@ pushd VTK-${VERSION}
...
         -DCMAKE_INSTALL_PREFIX="${SYSROOT_DIR}" \
         -DCMAKE_BUILD_TYPE=Release
-
-    make -j${THREAD_NUM}
+    if [ "${TARGET_ARCH}" = "aarch64" ]; then
+        thread_num=$(( $(nproc) / 2 ))
+        make -j${thread_num}
+    else
+        make -j$(nproc)
+    fi
```
xiaoxq, line 58: Prefer keep as much common part as possible, like:
```
    thread_num=$(nproc)
    if [ "${TARGET_ARCH}" = "aarch64" ]; then
       thread_num=$(( ${thread_num} / 2 ))
    fi
    make -j${thread_num}    
```

storypku: Done.  And `shellcheck`  suggests that `^-- SC2004: $/${} is unnecessary on arithmetic variables.`, so removed `${}`  of L55.

PR #11868, opened 2020-07-21
Build: only build simulator when <module> was not specified

storypku:


scripts/apollo_build.sh
```diff
@@ -156,10 +156,9 @@ function main() {
         info "Running build under CPU mode on ${ARCH} platform."
     fi
     bazel_build $@
-    # Disable simulator build temporarily
-    build_simulator
     if [ -z "${SHORTHAND_TARGETS}" ]; then
         SHORTHAND_TARGETS="apollo"
+        build_simulator
```
Capri2014, line 161: This might cause confusion since previously if we only use command ‘./apollo.sh build’, it means we want to build ALL modules by default.

PR #11870, opened 2020-07-21
perf: add perf install according system kernel

lx18233184051:


lx18233184051:
when first use ./apollo.sh perf will call apollo_perf.sh to install perf and perf record  report。

docker/build/installers/install_perf.sh
```diff
@@ -0,0 +1,28 @@
...
+# Fail on first error.
+set -e
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+#install perf
+apt-get -y update && \
+    apt-get -y install \
+    linux-tools-common \
+perf 2>&1 | grep linux-tools | head -n 1 | xargs sudo apt install -y
```
storypku, line 28: ?

lx18233184051: install perf include two step ,first linux-tools-common.the second according to perf tips to isntall the first package.

PR #11872, opened 2020-07-21
Docker: make the libtorch rely on cudart library from system directory

lfcarol:


docker/build/installers/install_libtorch.sh
```diff
@@ -38,9 +42,6 @@ pushd libtorch
 popd
 
 # Cleanup
-rm -rf libtorch ${PKG_NAME}
```
storypku, line 41: Please keep this line.

PR #11880, opened 2020-07-22
perf: add perf install and function

lx18233184051:


scripts/apollo_perf.sh
```diff
@@ -0,0 +1,23 @@
...
+
+function main() {
+    if ! command -v perf >/dev/null 2>&1; then
+      bash  "${TOP_DIR}/docker/scripts/install_perf.sh"
+      
+    fi
+    cd /apollo-simulator
+    bash build.sh run >/dev/null 2>&1 & 
+    sleep 1
+    ps -ef | grep "replay_engine" | grep -v grep | awk '{print $2}' |  xargs sudo perf record  -g -p
```
changsh726, line 16: `sudo perf record -g ./build.sh run` has the same function

lx18233184051: I have tried this way ,the problem:  libc10 shared library can't load  will happen. so use this way.

docker/scripts/install_perf.sh
```diff
@@ -0,0 +1,26 @@
...
+###############################################################################
+
+set -e
+
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
+#install perf
+sudo apt-get -y update 
+sudo apt-get -y install linux-tools-common
+perf 2>&1 | grep linux-tools | head -n 1 | xargs sudo apt install -y
```
xiaoxq, line 26: Seems that `sudo apt install -y linux-tools-generic` will automatically install `linux-tools-<KERNEL>-generic` and keep upgrading along with OS update. So we can avoid this ambiguous command.

scripts/apollo_perf.sh
```diff
@@ -0,0 +1,27 @@
...
+function main() {
+    if ! command -v perf >/dev/null 2>&1; then
+      bash  "${TOP_DIR}/docker/scripts/install_perf.sh"
+      
+    fi
+    if [ -e /apollo-simulator ];then
+      cd /apollo-simulator
+      bash build.sh run >/dev/null 2>&1 & 
+      sleep 1
+      ps -ef | grep "replay_engine" | grep -v grep | awk '{print $2}' |  xargs sudo perf record  -g -p
```
xiaoxq, line 17: 1. We can do this as a start. But better to find another clear way instead of `ps | grep` which may cause miss-hit silently.
2. Fix coding style. E.g.: There are several double-spaces. And the indents are mixed with 2-spaces and 4-spaces.

lx18233184051: the best way is: sudo perf record -g ./build.sh run. however this way run crash.

PR #11882, opened 2020-07-22
neolix canbus

wangjia1990:
add neolix edu car canbus module

### DIFF
--- a/modules/canbus/vehicle/neolix_edu/protocol/BUILD
+++ b/modules/canbus/vehicle/neolix_edu/protocol/BUILD
@@ -1,0 +1,27 @@
+load("//tools:cpplint.bzl", "cpplint")
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
+    name = "canbus_neolix_edu_protocol",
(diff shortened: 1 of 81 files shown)


modules/canbus/vehicle/neolix_edu/protocol/BUILD
```diff
@@ -0,0 +1,27 @@
+load("//tools:cpplint.bzl", "cpplint")
+
+package(default_visibility = ["//visibility:public"])
+
+cc_library(
```
macDure, line 5: please separate the cc_library into every .cc file

modules/canbus/vehicle/neolix_edu/protocol/BUILD
```diff
@@ -0,0 +1,27 @@
...
+    hdrs = glob([
+        "*.h",
+    ]),
+    deps = [
+        "//modules/drivers/canbus/common:canbus_common",
+        "//modules/canbus/proto:canbus_proto",
+        "//modules/drivers/canbus/can_comm:message_manager_base",
+    ],
+)
+cc_test(
```
macDure, line 17: please separate the cc_test into every .cc file

PR #11894, opened 2020-07-23
cyber: MockTime mode support

storypku:
This PR was based on @rongguodong 's request on simulation time support for `CyberRT`.  Cf. #11808 

Env vars and gflags were not used here.  Instead, `cyber/conf/cyber.pb.conf ` was the sole config file for specifying whether system(cyber)-clock or mock-clock should be used.

A clock node was created In `cyber::init` for each process.

Also notice there are `time.h` defs under `modules/common/time`, should we combine all these to provide a unified time model ?

### DIFF
--- a/cyber/BUILD
+++ b/cyber/BUILD
@@ -48,9 +48,12 @@ cc_library(
     hdrs = ["init.h"],
     deps = [
         "//cyber:state",
+        "//cyber/common:file",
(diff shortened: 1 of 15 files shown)


rongguodong:
What is reason to introduce a new class "cyber::Clock"? Why can't we directly add a mock mode in cyber::Time as in PR #11808?

PR #11900, opened 2020-07-23
Drivers: dev audio cpp  (draft, please don't merge)

storypku:


xiaoxq:
There is a button in "Reviewers" section which you can click to mark the change as a draft :)

modules/drivers/audio/proto/speaker_config.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto2";
+
+package apollo.drivers.audio.proto;
+
```
Capri2014, line 4: Hmmm, please remove my TODO since the proto is already moved to the right place

modules/drivers/audio/speaker_component.h
```diff
@@ -0,0 +1,81 @@
+/*********************************************************************
+ *
+ * Software License Agreement (BSD License)
```
Capri2014, line 3: Apache

Crysple: Sorry lol, copy the wrong one...

third_party/BUILD
```diff
@@ -13,3 +13,15 @@ cc_library(
         ],
     }),
 )
+
+# portaudio
+cc_library(
```
storypku, line 18: Don't follow the bad practice of define `cc_library` targets with linkopts only.
See `third_party/xxx/workspace.bzl` and `third_party/xxx/xxx.BUILD` on how to configure `bazel` external dependency.

Crysple: Thanks I will. Btw the doc [how_to_add_an_external_dependency.md](https://github.com/ApolloAuto/apollo/blob/master/docs/howto/how_to_add_an_external_dependency.md) may be updated later.

modules/drivers/audio/config/respeaker.pb.txt
```diff
@@ -0,0 +1,17 @@
+speaker_model: RESPEAKER
+channel_number: 6
```
Capri2014, line 2: this should be `.../conf/repeaker.pb.txt` to be consistent with your settings in dag and launch.

third_party/BUILD
storypku: > Thanks I will. Btw the doc [how_to_add_an_external_dependency.md](https://github.com/ApolloAuto/apollo/blob/master/docs/howto/how_to_add_an_external_dependency.md) may be updated later.

OK.

PR #11919, opened 2020-07-25
Docker: updated cyber.x86_64 with doxygen pre-installed

storypku:


docker/build/installers/install_qa_tools.sh
```diff
@@ -44,6 +44,13 @@ apt-get -y update && \
 
 bash /tmp/installers/install_gperftools.sh
 
+bash /tmp/installers/install_benchmark.sh
+
+# TechDoc generation
+bash /tmp/installers/install_doxygen.sh
```
xiaoxq, line 50: How about simply "apt install doxygen"?

docker/scripts/dev_start.sh
```diff
@@ -246,7 +246,7 @@ function local_volumes() {
 
 ## customized docker cmd
 function do_docker_image_inspect() {
-    docker image inspect -f {{.Config.Image}} $1 &> /dev/null
+    docker image inspect -f "{{.Config.Image}}" $1 &> /dev/null
```
xiaoxq, line 249: How about `docker images --format "{{.Repository}}:{{.Tag}}" | grep "^$IMAGE$"`?
As the `inspect` is pretty heavy and might have bug: https://github.com/docker/for-linux/issues/397

docker/build/installers/install_qa_tools.sh
storypku: > How about simply "apt install doxygen"?
Both are OK.  Doxygen installed via apt depends on  libclang1-6.0, and we have llvm-clang 10.0 installed already, to avoid different versions of clang stuff,  I chose to install doxygen from source. However, this approach has the disadvantage of less maintainability.

docker/scripts/dev_start.sh
storypku: > docker/for-linux#397

Good. I just noticed the `-f` option should be quoted as hinted by shellcheck then.

PR #11927, opened 2020-07-27
All: use std::shared_mutex instead of boost version

changsh726:


modules/dreamview/backend/data_collection_monitor/data_collection_monitor.h
```diff
@@ -116,8 +116,7 @@ class DataCollectionMonitor {
   nlohmann::json current_progress_json_;
 
   // Mutex to protect concurrent access to current_progress_json_.
-  // NOTE: Use boost until we have std version of rwlock support.
-  boost::shared_mutex mutex_;
+  std::shared_mutex mutex_;
```
storypku, line 119: Since std::shared_mutex was added in C++17, Extra measures should be taken to remain C++-14 compatibility.

PR #11928, opened 2020-07-27
perception time: use common timer to remove perception timer

lx18233184051:


lx18233184051:
perception/utils/timer is the same as common/time/timer

PR #11935, opened 2020-07-27
ci: improve format apollo_ci.sh

lx18233184051:


scripts/apollo_ci.sh
```diff
@@ -29,8 +32,12 @@ function main() {
     if [ "$1" == "test" ]; then
         info "Running CI Test ..."
         run_ci_test
-    else
+    elif [ "$1" == "build" ]; then
```
storypku, line 35: Since local var `cmd` was defined, better use `cmd` rather than the original `$1`

PR #11937, opened 2020-07-27
Drivers: add Speak Component for audio

Crysple:
* Includes
  * Speaker Cyber RT Component -- `speaker_component.h`, `speaker_component.cc`
  * Driver for Respeaker -- `respeaker.h`, `respeaker.cc`
  * Dependency
    * `portaudio19-dev`
      * Docker: add `apt-get install` in `install_drivers_deps.sh`
      * ThirdParty & Tools: add link options

* TODO:
  * Use real sound to test
  * Add code on online inference part to extract and transform orientation of audio


docker/build/installers/install_drivers_deps.sh
```diff
@@ -28,7 +28,7 @@ bash /tmp/installers/install_proj4.sh
 
 # Required by audio-driver python version
 apt-get -y update && \
-    apt-get -y install python3-pyaudio
+    apt-get -y install python3-pyaudio portaudio19-dev
```
Capri2014, line 31: @Crysple a new docker might be needed for this pr, right? @xiaoxq @lfcarol @storypku

modules/drivers/audio/speaker_component.cc
```diff
@@ -0,0 +1,105 @@
...
+    if (chunk_ * --i < total_frames_)
+      speaker_device_->stream_->read_stream(total_frames_ - chunk_ * i, pos);
+    AINFO << "End recording";
+
+    for (int buff_i = 0, i = 0; buff_i < buffer_size_; i += 2) {
+      for (int channel_i = 0; channel_i < n_channels_; ++channel_i) {
+        (*channel_data_ptr_[channel_i])[i] = buffer_[buff_i++];
+        (*channel_data_ptr_[channel_i])[i + 1] = buffer_[buff_i++];
+      }
+    }
```
Capri2014, line 91: fulfill header before publish the data.

modules/drivers/audio/speaker_component.h
```diff
@@ -0,0 +1,65 @@
...
+  ~SpeakerComponent();
+
+ private:
+  void run();
+
+  // Configuration
+  int n_chunks_, n_channels_, chunk_, chunk_size_, buffer_size_, total_frames_;
+
+  // class data
+  std::shared_ptr<AudioData> audio_data_;
```
Capri2014, line 51: better differentiate pointer and non_pointer with name `audio_data` or `audio_data_ptr_`

Capri2014:
@Crysple give adequate namespace and correct scopes, as the other modules do.

modules/drivers/audio/speaker_component.h
Crysple: Good suggestion. Have added `ptr` suffix to all variable of pointer type

modules/drivers/audio/respeaker.h
```diff
@@ -0,0 +1,69 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+
+#pragma once
+
+#include <stdio.h>
```
storypku, line 19: Better use c++ version: <cstdio>

modules/drivers/audio/speaker_component.cc
Crysple: I use `grep -nr 'mutable_head()` under `modules/drivers` to see how other drivers do to header and mimic them, adding the following fields to header
1. `frame_id`: I guess it's the type of config for the hardware, so I add a string in speaker config as "respeaker" and set it to header
2. `sequence_num`: an int incremented by one per message
3. `timestamp_sec`: current time -- `mutable_header()->set_timestamp_sec(cyber::Time().Now().ToSecond());`

If there is any field missed in header please comment here and I will make further changes.

modules/drivers/audio/speaker_component.h
```diff
@@ -0,0 +1,67 @@
...
+#include <atomic>
+#include <future>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "modules/drivers/audio/proto/speaker_config.pb.h"
+#include "modules/drivers/audio/proto/audio.pb.h"
+#include "modules/drivers/audio/respeaker.h"
+#include "cyber/cyber.h"
```
storypku, line 28: include order: `cyber/xxx.h` before `modules/xxx/xxx.h`.  Run `scripts/clang_format.sh` to do that.

docker/build/installers/install_drivers_deps.sh
storypku: > @Crysple a new docker might be needed for this pr, right? @xiaoxq @lfcarol @storypku

Yep.

modules/drivers/audio/respeaker.h
Crysple: Yes. After examining the code again I found that stdio was not used, so I deleted this line. // It was used at the very beginning for `printf` for testing purpose.

modules/drivers/audio/speaker_component.h
Crysple: Thank you! Previously I just sorted them using my eyes... but this one is missed.
I have used `scripts/clang_format.sh` to format the four files according to your suggestion.

modules/drivers/audio/respeaker.cc
```diff
@@ -0,0 +1,156 @@
...
+}
+
+int Stream::get_chunk_size(int n_frames) {
+  return (n_frames) * (inputParameters_ptr_->channelCount) *
+         (Pa_GetSampleSize(inputParameters_ptr_->sampleFormat));
+}
+
+// Respeaker
+Respeaker::~Respeaker() { Pa_Terminate(); }
+void Respeaker::init(std::shared_ptr<SpeakerConfig> speaker_config) {
```
storypku, line 72: Better use `const std::shared_ptr<SpeakerConfig>&`  since `speaker_config` was read only in the `init` function.

modules/drivers/audio/respeaker.cc
```diff
@@ -0,0 +1,156 @@
...
+  switch (width) {
+    case 1:
+      return is_unsigned ? paUInt8 : paInt8;
+    case 2:
+      return paInt16;
+    case 3:
+      return paInt24;
+    case 4:
+      return paFloat32;
+  }
```
storypku, line 100: Add a `default: break; ` here for readability.

modules/drivers/audio/respeaker.cc
```diff
@@ -0,0 +1,156 @@
...
+  AERROR << "invalid width: " << width;
+  throw std::runtime_error("");
```
storypku, line 102: Does we prefer returning error code or throwing exception for error handling @xiaoxq @Capri2014 @jinghaomiao ?

modules/drivers/audio/respeaker.cc
```diff
@@ -0,0 +1,156 @@
...
+    if (std::string(device_info->name).find("ReSpeaker") != std::string::npos) {
+      return real_index;
+    }
+  }
+  AERROR << "respeaker device not found";
+  throw std::runtime_error("");
+}
+
+const PaDeviceInfo *Respeaker::get_device_info(PaDeviceIndex index) {
+  const PaDeviceInfo *_info =
```
storypku, line 122: local variable names started with underscore are discouraged ?

modules/drivers/audio/respeaker.h
```diff
@@ -0,0 +1,68 @@
...
+  ~Stream();
+  void init_stream(int rate, int channels, int chunk, int input_device_index,
+                   PaSampleFormat format);
+  void read_stream(int n_frames, char *buffer);
+  int get_chunk_size(int n_frames);
+};
+
+class Respeaker {
+ public:
+  std::unique_ptr<Stream> stream_ptr_;
```
storypku, line 53: Should stream_ptr_ be public ?  Have this header file passed our lint check ( by running `./apollo.sh lint`) ?

modules/drivers/audio/speaker_component.cc
```diff
@@ -0,0 +1,112 @@
...
+  // dump config, calculate size and reserve buffer for audio
+  // chunk_: number of frames per chunk; chunk_size_: number of bytes per chunk
+  n_channels_ = speaker_config_ptr_->channel_type_size();
+  chunk_ = speaker_config_ptr_->chunk();
+  total_frames_ = speaker_config_ptr_->sample_rate() *
+                  speaker_config_ptr_->record_seconds();
+  chunk_size_ = speaker_device_ptr_->stream_ptr_->get_chunk_size(chunk_);
+
+  buffer_size_ = speaker_device_ptr_->stream_ptr_->get_chunk_size(total_frames_);
+  buffer_ = reinterpret_cast<char *>(malloc(buffer_size_));
```
storypku, line 44: Better use c++ style `new[]`  instead of c-style malloc for memery allocation.

docker/build/installers/install_drivers_deps.sh
Capri2014: > > @Crysple a new docker might be needed for this pr, right? @xiaoxq @lfcarol @storypku
> 
> Yep.

Could you help with the new docker please ?

modules/drivers/audio/speaker_component.cc
Capri2014: > I use `grep -nr 'mutable_head()` under `modules/drivers` to see how other drivers do to header and mimic them, adding the following fields to header
> 
> 1. `frame_id`: I guess it's the type of config for the hardware, so I add a string in speaker config as "respeaker" and set it to header
> 2. `sequence_num`: an int incremented by one per message
> 3. `timestamp_sec`: current time -- `mutable_header()->set_timestamp_sec(cyber::Time().Now().ToSecond());`
> 
> If there is any field missed in header please comment here and I will make further changes.

Yes, those are the critical three fields.

modules/drivers/audio/respeaker.cc
Capri2014: > Does we prefer returning error code or throwing exception for error handling @xiaoxq @Capri2014 @jinghaomiao ?

No, just return of error code, we can leave this as TODO in this pr. @Crysple

docker/build/installers/install_drivers_deps.sh
storypku: > Could you help with the new docker please ?

No problem.

modules/drivers/audio/respeaker.cc
Crysple: You are right but I think it should be `const std::shared_ptr<const SpeakerConfig>&` which also prevents modification to `SpeakerConfig` itself, while `const std::shared_ptr<SpeakerConfig>&` just prevents the re-assignment of that shared_ptr variable?

modules/drivers/audio/respeaker.cc
Crysple: According to my experience, AERROR will block in anywhere except the block inside `while (!cyber::IsShutdown())`, which is in the `void run()` of the `component` class -- It will keep getting the same error again & again -- (since its return type is `void` I cannot return). That's the reason I used `throw`.  For elsewhere (init, etc.), I think I can change the code easily if returning error code is preferred.

modules/drivers/audio/respeaker.h
Crysple: Yes it did pass `./apollo.sh lint`.
I used it as public because otherwise I have to encapsulate member functions of `Stream` in `Respeaker` with the same name. (like `int Respeaker::get_chunk_size(arg){pastream_->get_chunk_size(arg)};`) Do you think it's fine? Or do you have any suggestions? Thank you.

modules/drivers/audio/speaker_component.cc
Crysple: Yes, didn't notice that. Have changed it to `new`

modules/drivers/audio/respeaker.cc
Crysple: At first I was afraid that there might be a variable called `info` globally so I added an underscore subconsciously.
Have changed it to a more reasonable name.

modules/drivers/audio/respeaker.cc
Crysple: Thank you for your suggestion. Have added it.

modules/drivers/audio/respeaker.cc
```diff
@@ -0,0 +1,156 @@
...
+      Pa_GetDeviceInfo(inputParameters_ptr_->device)->defaultLowInputLatency;
+  inputParameters_ptr_->hostApiSpecificStreamInfo = nullptr;
+
+  err = Pa_OpenStream(
+      &pastream_ptr_, inputParameters_ptr_, nullptr, rate, chunk,
+      paClipOff,  // we only use input so don't bother clipping them *
+      nullptr, nullptr);
+  if (err != paNoError) report_error(err, "Pa_OpenStream");
+  err = Pa_StartStream(pastream_ptr_);
+  if (err != paNoError) report_error(err, "Pa_StartStream");
```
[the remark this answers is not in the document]
Crysple: Fixed~

modules/drivers/audio/respeaker.cc
```diff
@@ -0,0 +1,156 @@
...
+  inputParameters_ptr_->sampleFormat = format;
+  inputParameters_ptr_->suggestedLatency =
+      Pa_GetDeviceInfo(inputParameters_ptr_->device)->defaultLowInputLatency;
+  inputParameters_ptr_->hostApiSpecificStreamInfo = nullptr;
+
+  err = Pa_OpenStream(
+      &pastream_ptr_, inputParameters_ptr_, nullptr, rate, chunk,
+      paClipOff,  // we only use input so don't bother clipping them *
+      nullptr, nullptr);
+  if (err != paNoError) report_error(err, "Pa_OpenStream");
```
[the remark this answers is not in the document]
Crysple: Fixed~

modules/drivers/audio/respeaker.cc
```diff
@@ -0,0 +1,163 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
+ *
```
Capri2014, line 3: Year 2020

modules/drivers/audio/respeaker.cc
Capri2014: Yes, please remove the throw exception logic in this file and another

modules/drivers/audio/respeaker.h
```diff
@@ -0,0 +1,67 @@
+/******************************************************************************
+ * Copyright 2017 The Apollo Authors. All Rights Reserved.
```
Capri2014, line 2: Year 2020

modules/drivers/audio/respeaker.h
Capri2014: Yes please do so and make the return value/ function as a const

modules/drivers/audio/speaker_component.cc
Crysple: Now I change to use `Fill header` in `modules/common` for code reuse. If there's any problems or concerns, please comment here.

modules/drivers/audio/respeaker.cc
Crysple: > Yes, please remove the throw exception logic in this file and another

Ok now I remove all exceptions in related to Init. I remain the one in `read_stream` but I also add a `try catch` in `run` -- so that I can catch and break from `run` returning to apollo once there's any error.

modules/drivers/audio/respeaker.h
Crysple: > Yes please do so and make the return value/ function as a const

Done. I kind of `const` everything if possible...

modules/drivers/audio/respeaker.cc
Crysple: Fix~

modules/drivers/audio/respeaker.h
Crysple: Fix~

PR #11947, opened 2020-07-28
dev_start:improve dev_start usage

lx18233184051:


lx18233184051:
> * dev_start.sh#L53 should be placed inside `else` blocks.
> * dev_start.sh#L129 should be removed. DON'T scatter your logic here and there, it's hard to maintain.

L129 can't removed now, because check_agreement is before parse the parameter.
I will refactor other part.

PR #11951, opened 2020-07-28
dev_start: refactor this file

lx18233184051:


docker/scripts/dev_start.sh
```diff
@@ -35,41 +35,41 @@ VERSION_AARCH64="dev-aarch64-20170927_1111"
...
+        return
+    fi
+
+    AGREEMENT_FILE="$APOLLO_ROOT_DIR/scripts/AGREEMENT.txt"
+    if [ ! -e "$AGREEMENT_FILE" ]; then
+        error "AGREEMENT $AGREEMENT_FILE does not exist."
+        exit 1
+    fi
+
+    cat $AGREEMENT_FILE
```
storypku, line 53: Please move L53-55 inside the `else` block at L60.

docker/scripts/dev_start.sh
```diff
@@ -321,6 +315,9 @@ function determine_gpu_use() {
 }
 
 function main() {
+    init
+    parse_arguments
```
storypku, line 319: Have you tested this script before PR?
`parse_arguments` without "$@" should not work.

PR #11953, opened 2020-07-28
Docker: installer updates to reduce image size

storypku:


xiaoxq:
BTW, could you try if we can get rid of `install_protobuf.sh`? We are actually building protobuf in workspace.
Or at least remove the python part: https://github.com/ApolloAuto/apollo/blob/master/docker/build/installers/install_protobuf.sh#L61
which is obviously unused.

storypku:
> BTW, could you try if we can get rid of `install_protobuf.sh`? We are actually building protobuf in workspace.
> Or at least remove the python part: https://github.com/ApolloAuto/apollo/blob/master/docker/build/installers/install_protobuf.sh#L61
> which is obviously unused.

IIRC, protobuf was needed to build some other libraries. On the python part,  we can do some experiments to decide whether it should be removed or not.

storypku:
> BTW, could you try if we can get rid of `install_protobuf.sh`? We are actually building protobuf in workspace.
> Or at least remove the python part: https://github.com/ApolloAuto/apollo/blob/master/docker/build/installers/install_protobuf.sh#L61
> which is obviously unused.

Please note that previously we have `pip3_install protobuf to install python protobuf support`, whose pb version may be different from the pre-built one.

xiaoxq:
> Please note that previously we have `pip3_install protobuf` to install python protobuf support, whose pb version may be different from the pre-built one.

Good point. Then let's keep current solution to guarantee consistency, otherwise we have to do things like "pip3_install protobuf==3.12.3"

PR #11960, opened 2020-07-29
location_compression: safety review to init z_stream struct

lx18233184051:


lx18233184051:
1. var_decl: Declaring variable stream_data without initializer.
6. uninit_use_in_call: Using uninitialized value stream_data.total_in when calling inflate
7. uninit_use_in_call: Using uninitialized value stream_data.total_out when calling inflate

modules/localization/msf/common/util/compression.cc
```diff
@@ -107,6 +107,14 @@ int ZlibStrategy::ZlibUncompress(BufferStr* src, BufferStr* dst) {
   stream_data.opaque = Z_NULL;
   stream_data.avail_in = 0;
   stream_data.next_in = Z_NULL;
+  stream_data.total_in = 0ULL;
```
storypku, line 110: Please don't try to init every member in struct one by one, which was hard to read/maintain.  Use `std::memset(...)` from <cstring> to zero-initialize the whole struct first.

lx18233184051: ok

modules/localization/msf/common/util/compression.cc
```diff
@@ -102,11 +102,8 @@ int ZlibStrategy::ZlibUncompress(BufferStr* src, BufferStr* dst) {
   unsigned int dst_idx = 0;
 
   /* allocate inflate state */
-  stream_data.zalloc = Z_NULL;
-  stream_data.zfree = Z_NULL;
-  stream_data.opaque = Z_NULL;
-  stream_data.avail_in = 0;
-  stream_data.next_in = Z_NULL;
+  memset(&stream_data, 0, sizeof(z_stream));
```
storypku, line 105: Better write `std::memset`

lx18233184051: ok

PR #11971, opened 2020-07-30
Docker: ffmpeg/vtk/pcl installer updates

storypku:


docker/build/installers/install_pcl.sh
```diff
@@ -31,19 +35,34 @@ if ldconfig -p | grep -q libpcl_common ; then
...
+if [ "${WORKHORSE}" = "cpu" ]; then
+    GPU_OPTIONS="-DWITH_CUDA=OFF"
+fi
+
+info "GPU Options for PCL:\"${GPU_OPTIONS}\""
+
+TARGET_ARCH="$(uname -m)"
+ARCH_OPTIONS=""
+if [ "${TARGET_ARCH}" = "x86_64" ]; then
+    ARCH_OPTIONS="-DPCL_ENABLE_SSE=ON"
```
changsh726, line 48: Does this cause compatibility problems between different machines?

storypku: > between

This setting was based on https://src.fedoraproject.org/rpms/pcl/blob/master/f/pcl.spec.
I think the main reason for doing so was, SSE settings were meaningless for non-x86 CPU architectures.

changsh726: But I remember that dreamview module might generate coredump on some machines if SSE is enabled.

changsh726: sorry, I didn't notice that you fixed it by the patch

PR #11973, opened 2020-07-30
audio: initialization

kechxu:


Capri2014:
better rename is maybe? Since similar names use under https://github.com/ApolloAuto/apollo/tree/master/modules/drivers and it is actually audio driver.

Capri2014:
Or rename that folder to `mic`

storypku:
> better rename is maybe? Since similar folder names were use under https://github.com/ApolloAuto/apollo/tree/master/modules/drivers and that one actually means audio driver.

It seems that `audio` should be part of a first-class submodule (e.g. HMI) rather than a first-class module  itself.

Capri2014:
> > better rename is maybe? Since similar folder names were use under https://github.com/ApolloAuto/apollo/tree/master/modules/drivers and that one actually means audio driver.
> 
> It seems that `audio` should be part of a first-class submodule (e.g. HMI) rather than a first-class module itself.

That’s another reason we need to rename this module to avoid confusion. - this new module is not for audio communication purpose. But let’s rename it together with the driver part once all elements ready.

PR #11974, opened 2020-07-30
add drivers to support RoboSense LiDAR

yfcube:
RS16, RS32, RS128, RSBP are now supported


Capri2014:
why directly merge into 5.5? Why not merge into master and cherry-pick to 5.5?

SeasoulChris:
> why directly merge into 5.5? Why not merge into master and cherry-pick to 5.5?
cause the the lidar driver is a litter different in master and 5.5,  proto files in master should support both cpp and python, but in 5.5, cpp supported only , so I think It's okay to merge separately

lx18233184051:
this pr has existed a long time , Is necessary to close it ? if needed you could reopen it ?

PR #11978, opened 2020-07-30
Monitor: add Module Monitor

changsh726:


modules/dreamview/proto/hmi_mode.proto
```diff
@@ -74,18 +79,23 @@ message MonitoredComponent {
...
+  // Generate SystemStatus.components[i].module_status.
+  // OK if the module is running.
+  // FATAL if the module is down.
+  optional ModuleMonitorConfig module = 2;
+
   // Generate SystemStatus.components[i].channel_status.
   // OK if delay is not notable.
   // FATAL if delay is larger than fatal_delay.
-  optional ChannelMonitorConfig channel = 2;
+  optional ChannelMonitorConfig channel = 3;
```
xiaoxq, line 90: Don't modify field ID. Just use the next available one for new field.

modules/monitor/software/module_monitor.cc
```diff
@@ -0,0 +1,90 @@
...
+#include "cyber/common/file.h"
+#include "cyber/common/log.h"
+#include "modules/common/util/map_util.h"
+#include "modules/monitor/common/monitor_manager.h"
+#include "modules/monitor/software/summary_monitor.h"
+
+DEFINE_string(module_monitor_name, "ModuleMonitor",
+              "Name of the modules monitor.");
+
+DEFINE_double(module_monitor_interval, 1.5,
```
xiaoxq, line 30: Use absl flag.
https://abseil.io/docs/cpp/guides/flags

xiaoxq:
Nice work! We can merge it if you have tested it locally.

PR #11979, opened 2020-07-30
absl_flag: convert gflags to absl_flag

lx18233184051:
left gflags will be processed tomorrow, simulation also need to update.

### DIFF
--- a/modules/v2x/v2x_proxy/app/v2x_proxy.cc
+++ b/modules/v2x/v2x_proxy/app/v2x_proxy.cc
@@ -35,14 +35,14 @@ V2xProxy::V2xProxy()
 
   auto x2v_trafficlight_timer_period = ceil(
       (1.0 / static_cast<int>(FLAGS_x2v_trafficlight_timer_frequency)) * 1000);
-  x2v_trafficlight_timer_.reset(
-      new cyber::Timer(static_cast<uint32_t>(x2v_trafficlight_timer_period),
-                       [this]() { this->OnX2vTrafficLightTimer(); }, false));
+  x2v_trafficlight_timer_.reset(new cyber::Timer(
+      static_cast<uint32_t>(x2v_trafficlight_timer_period),
+      [this]() { this->OnX2vTrafficLightTimer(); }, false));
   auto v2x_carstatus_timer_period = ceil(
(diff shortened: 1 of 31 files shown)


storypku:
> left gflags will be processed tomorrow, simulation also need to update.

Can you please make clear the reason for the switch from gflags to absl flags ?

lx18233184051:
> > left gflags will be processed tomorrow, simulation also need to update.
> 
> Can you please make clear the reason for the switch from gflags to absl flags ?

absl has some feature，such as support vector of string

ABSL_FLAG(std::vector<std::string>, languages,
          std::vector<std::string>({"english", "french", "german"}),
and support time

ABSL_FLAG(absl::Duration, timeout, absl::Seconds(30), "Default RPC deadline");

modules/v2x/v2x_proxy/app/v2x_proxy.cc
```diff
@@ -57,7 +57,8 @@ V2xProxy::V2xProxy()
   }
 
   hdmap_.reset(new apollo::hdmap::HDMap());
-  std::string map_name = FLAGS_map_dir + "/" + FLAGS_base_map_filename;
+  std::string map_name = absl::GetFlag(FLAGS_map_dir) + "/" +
+                         absl::GetFlag(FLAGS_base_map_filename);
```
xiaoxq, line 61: Use `absl::StrCat` for all such cases.
https://abseil.io/tips/3

lx18233184051: ok

modules/v2x/v2x_proxy/app/v2x_proxy.cc
```diff
@@ -57,7 +57,8 @@ V2xProxy::V2xProxy()
   }
 
   hdmap_.reset(new apollo::hdmap::HDMap());
-  std::string map_name = FLAGS_map_dir + "/" + FLAGS_base_map_filename;
+  std::string map_name = absl::StrCat(absl::GetFlag(FLAGS_map_dir), "/",
+                                      absl::GetFlag(FLAGS_base_map_filename));
```
xiaoxq, line 61: Seems like we have a lot of operations on file path.
Maybe we can add a "modules/common/util/file.h" later to provide things similar to python's
```
os.path.join()
os.path.basename()
...
```

PR #11988, opened 2020-07-31
audio: add config for button to trigger microphone

kechxu:


modules/dreamview/conf/hmi_modes/mkz_standard_debug.pb.txt
```diff
@@ -95,6 +95,12 @@ cyber_modules {
     dag_files: "/apollo/modules/third_party_perception/dag/third_party_perception.dag"
   }
 }
+cyber_modules {
+  key: "microphone"
```
xiaoxq, line 99: Title case.

PR #11989, opened 2020-07-31
Localization: raw imu frequency match with odometry frequency

lfcarol:
Localization: raw imu frequency match with odometry frequency

### DIFF
--- a/modules/localization/msf/BUILD
+++ b/modules/localization/msf/BUILD
@@ -39,6 +39,7 @@ cc_library(
         "//modules/localization/msf/local_pyramid_map/base_map:localization_pyramid_map_base_map",
         "//modules/localization/msf/local_pyramid_map/pyramid_map:localization_pyramid_map_pyramid_map",
         "//modules/localization/proto:gnss_pnt_result_cc_proto",
+        "//modules/localization/proto:gps_cc_proto",
         "//modules/localization/proto:localization_cc_proto",
         "//modules/localization/proto:localization_config_cc_proto",
         "//modules/localization/proto:localization_status_cc_proto",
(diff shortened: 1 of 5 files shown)


modules/localization/msf/BUILD
```diff
@@ -44,6 +44,7 @@ cc_library(
         "//modules/localization/proto:localization_status_cc_proto",
         "//modules/localization/proto:measure_cc_proto",
         "//modules/localization/proto:sins_pva_cc_proto",
+        "//modules/localization/proto:gps_cc_proto",
```
xiaoxq, line 47: Make it in order

modules/localization/msf/msf_localization.cc
```diff
@@ -294,6 +302,8 @@ void MSFLocalization::OnGnssHeading(
   localization_integ_.GnssHeadingProcess(*gnss_heading_msg);
 }
 
+void MSFLocalization::OnGps() { OnRawImu(raw_imu_msg_); }
```
xiaoxq, line 305: From my understanding, you are doing:
1. Cache latest IMU message.
2. OnGps() triggers processing the cached IMU message.
So that we are processing IMU at GPS's frequency.

I'm not sure if we should do the frequency control from sender side or receiver side. Let's wait for @Capri2014 's comment.

Capri2014: > From my understanding, you are doing:
> 
> 1. Cache latest IMU message.
> 2. OnGps() triggers processing the cached IMU message.
>    So that we are processing IMU at GPS's frequency.
> 
> I'm not sure if we should do the frequency control from sender side or receiver side. Let's wait for @Capri2014 's comment.

Yes, change on receiver as of now, this is by no means ideal but has least side effect theoretically. (Still need to verify via offline dataset and on vehicle performance.)

jinghaomiao: Is a lock necessary?

lfcarol: > Is a lock necessary?

OK

modules/localization/msf/BUILD
lfcarol: > Make it in order

OK

PR #11997, opened 2020-07-31
Revert "absl_flag: convert gflags to absl_flag"

xiaoxq:
This reverts commit 6581599127f5b349b314b2499695c7307a34f2e2.

### DIFF
--- a/modules/common/configs/BUILD
+++ b/modules/common/configs/BUILD
@@ -12,7 +12,6 @@ cc_library(
     ],
     deps = [
         "@com_github_gflags_gflags//:gflags",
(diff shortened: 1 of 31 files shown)


xiaoxq:
Sorry that I have to revert this change. I get it wrong that gflags and absl_flags could be used together during migration.
There might be conflicts:
1. Both of them need to parse command line arguments, i.e., `google::ParseCommandLineFlags` and `absl::ParseCommandLine()`. They'll confuse.
2. Both of them need to parse `--flagfile`.

So as to make it safe, we have to migrate the whole repo all at once I guess.

PR #12010, opened 2020-08-03
Bazel: split target //modules/map/hdmap/adapter:opendrive_adapter

changsh726:


modules/map/hdmap/adapter/xml_parser/BUILD
```diff
@@ -0,0 +1,113 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("//tools:cpplint.bzl", "cpplint")
+
+package(default_visibility = ["//visibility:public"])
+
```
storypku, line 5: What about adding a `cc_library` named `xml_parser` here to encapsulate the details of various `xxx_xml_parser` ?
cc @xiaoxq

PR #12017, opened 2020-08-03
use std::numeric_limits: update  INT_MIN/MAX, FLT_MIN/MAX, DBL_MIN/MAX etc

lx18233184051:


cyber/event/perf_event.h
```diff
@@ -157,7 +158,7 @@ class TransportEvent : public EventBase {
  private:
   std::string adder_ = "";
   uint64_t msg_seq_ = 0;
-  uint64_t channel_id_ = UINT64_MAX;
+  uint64_t channel_id_ = std::numeric_limits<u_int64_t>::max();
```
storypku, line 161: Better use the more wildly accepted `uint64_t` rather than `u_int64_t` defined in <sys/types.h>.

xiaoxq: Agree. And at least the two `T` in `T channel_id_ = std::numeric_limits<T>::max();` should be aligned.

cyber/scheduler/policy/choreography_context.cc
```diff
@@ -17,6 +17,7 @@
 #include "cyber/scheduler/policy/choreography_context.h"
 
 #include <climits>
```
storypku, line 19: The `<climits>` header can be safely deleted now as XXX_MAX/MIN was no longer used ?

cyber/record/record_viewer.h
```diff
@@ -46,9 +47,10 @@ class RecordViewer {
    * @param end_time
    * @param channels
    */
-  RecordViewer(const RecordReaderPtr& reader, uint64_t begin_time = 0,
-               uint64_t end_time = UINT64_MAX,
-               const std::set<std::string>& channels = std::set<std::string>());
+  RecordViewer(
+      const RecordReaderPtr& reader, uint64_t begin_time = 0,
+      uint64_t end_time = std::numeric_limits<uint64_t>::max(),
+      const std::set<std::string>& channels = std::set<std::string>());
```
storypku, line 53: Just write `{}` rather than `std::set<std::string>()`.

cyber/scheduler/policy/classic_context.cc
```diff
@@ -17,6 +17,7 @@
 #include "cyber/scheduler/policy/classic_context.h"
 
 #include <climits>
```
storypku, line 19: ditto.

modules/data/tools/smart_recorder/realtime_record_processor.cc
```diff
@@ -19,6 +19,7 @@
 #include <algorithm>
 #include <chrono>
 #include <csignal>
```
storypku, line 21: csignal should be put before algorithm and has its own section.

modules/perception/camera/common/twod_threed_util.h
```diff
@@ -17,6 +17,7 @@
 
 #include <algorithm>
 #include <cfloat>
```
storypku, line 19: `<cfloat>` can be removed ?

modules/perception/camera/lib/lane/common/common_functions.h
```diff
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "Eigen/Core"
+#include <limits>
```
storypku, line 21: system header `<limits>` should be put before third_party `Eigen`

modules/data/tools/smart_recorder/realtime_record_processor.cc
lx18233184051: csignal is a c++ header, it's translated from c,  so separately place first?

storypku: > csignal is a c++ header, it's translated from c, so separately place first?

yep.

PR #12018, opened 2020-08-03
Profiling: //modules/common/util:perf_util to provide PERF_XXX functions

storypku:
Please note that the `perception` module  was  shipped with time perf enabled (Since `DISABLE_PERF` was not defined by default), which may cause CPU waste for production use. Further work need to be done to toggle  this behavior via bazal config.

@xiaoxq @kechxu


modules/perception/onboard/component/radar_detection_component.cc
```diff
@@ -17,7 +17,7 @@
 
 #include "modules/common/time/time.h"
 #include "modules/perception/common/sensor_manager/sensor_manager.h"
-#include "modules/perception/lib/utils/perf.h"
+#include "modules/common/util/perf_util.h"
```
xiaoxq, line 20: Reorder.

modules/perception/onboard/component/BUILD
```diff
@@ -146,7 +146,7 @@ cc_library(
         "//modules/perception/base",
         "//modules/perception/common/sensor_manager",
         "//modules/perception/lib/registerer",
-        "//modules/perception/lib/utils",
+        "//modules/common/util:perf_util",
```
xiaoxq, line 149: ditto. If you made this change with script, maybe just call clang-format for CPP files and buildifier for BUILD files at the end of the script :)

storypku: > ditto. If you made this change with script, maybe just call clang-format for CPP files and buildifier for BUILD files at the end of the script :)

Forgot to buildify these files in a hurry. :-(

PR #12019, opened 2020-08-03
Drivers: fix hardcode of sample_width

Crysple:
Not ready
* Sample_width is the number of bytes every sample. I didn't notice that I hardcode them in `microphone_component.cc` 95 as `i` and `i+1` -- originally it was default to be 2. Now I change them to a `for` loop
* Change buffer size from **"_size for all the seconds_"** TO **"_size for only one chunk_"**, reduce startup time by 40% and reduce memory usage.
  * Previously, a `for` loop to read all required audio data into buffer (many chunks). Then another `for` loop to copy data in buffer to `bytes` in proto
  * Now, one `for` -- read a chunk, copy to buffer, then to `byte` in proto, repeatedly
* Change the way to calculate size.
  * refer to readme.
* Fix others like `malloc` to `new`.

### DIFF
--- a/modules/drivers/microphone/microphone_component.cc
+++ b/modules/drivers/microphone/microphone_component.cc
@@ -33,17 +33,17 @@ bool MicrophoneComponent::Init() {
   microphone_device_ptr_->init(microphone_config_ptr_);
 
   // dump config, calculate size and reserve buffer for audio
-  // chunk_: number of frames per chunk; chunk_size_: number of bytes per chunk
   n_channels_ = microphone_config_ptr_->channel_type_size();
+  sample_width_ = microphone_config_ptr_->sample_width();
+  // chunk_: number of frames per chunk; chunk_size_: number of bytes per chunk
   chunk_ = microphone_config_ptr_->chunk();
-  total_frames_ = microphone_config_ptr_->sample_rate() *
-                  microphone_config_ptr_->record_seconds();
-  chunk_size_ = microphone_device_ptr_->get_chunk_size(chunk_);
-
-  buffer_size_ = microphone_device_ptr_->get_chunk_size(total_frames_);
-  buffer_ = new char[buffer_size_];
+  n_chunk_ = static_cast<int>(
+      std::floor(microphone_config_ptr_->sample_rate() *
+                 microphone_config_ptr_->record_seconds() / chunk_));
+  chunk_size_ = chunk_ * n_channels_ * sample_width_;
+  buffer_ = new char[chunk_size_];
   if (buffer_ == nullptr) {
-    AERROR << "System calloc memory error, size:" << buffer_size_;
+    AERROR << "System new memory error, size:" << chunk_size_;
     return false;
   }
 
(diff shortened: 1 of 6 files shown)


modules/drivers/microphone/microphone_component.cc
```diff
@@ -72,30 +71,29 @@ bool MicrophoneComponent::Init() {
   return true;
 }
 
+void MicrophoneComponent::fill_channel_data(int chunk_i) {
+  // next index of channel data to be filled
+  int pos = chunk_i * chunk_ * sample_width_;
+  for (int buff_i = 0; buff_i < chunk_size_; pos += sample_width_){
```
xiaoxq, line 77: Run "apollo.sh lint" every time before submitting the code.

Crysple: Sorry, miss this. Will fix.

xiaoxq:
LGTM on coding. Please add other reviewers for business logic.

modules/drivers/microphone/respeaker.cc
```diff
@@ -28,15 +30,15 @@ void report_error(PaError err, const std::string &func_name) {
 }
 
 // Stream
-Stream::~Stream() { Pa_CloseStream(pastream_ptr_); }
+Stream::~Stream() {
+  Pa_CloseStream(pastream_ptr_);
+  free(inputParameters_ptr_);
```
storypku, line 35: Does naming like `inputParameters_ptr_` meet our coding style ? @xiaoxq

kechxu: will fix this naming in the next commit

storypku: > will fix this naming in the next commit

OK.

modules/drivers/microphone/microphone_component.cc
storypku, line 77: `./apollo.sh lint` fails as missing space at the end of L77-79
`cpplint.py modules/drivers/microphone/microphone_component.cc` gives:

```
modules/drivers/microphone/microphone_component.cc:77:  Missing space before {  [whitespace/braces] [5]
modules/drivers/microphone/microphone_component.cc:78:  Missing space before {  [whitespace/braces] [5]
modules/drivers/microphone/microphone_component.cc:79:  Missing space before {  [whitespace/braces] [5]
Done processing modules/drivers/microphone/microphone_component.cc
Total errors found: 3
```

modules/drivers/microphone/respeaker.cc
Crysple: In the last pr I was suggested to add _ptr suffix for pointer variable for clarification so... I am ok with both

storypku: > In the last pr I was suggested to add _ptr suffix for pointer variable for clarification so... I am ok with both

Both `input_parameters_ptr_` and `inputParametersPtr` should be ok if not constraint by our coding style,
While `inputParameter_ptr` may not look good in any coding style.

modules/drivers/microphone/microphone_component.cc
Crysple: Fixed~

modules/drivers/microphone/respeaker.cc
Crysple: Sorry I didn't get it just now lol. You are right. Let me fix it

Crysple: Fixed

PR #12020, opened 2020-08-03
Profiling: bugfix for function_signature

storypku:


modules/common/util/perf_util.cc
```diff
@@ -40,6 +40,9 @@ namespace util {
 std::string function_signature(const std::string& func_name,
                                const std::string& indicator) {
   auto simplified_name = func_name_simplified(func_name);
+  if (indicator.empty()) {
+    return simplified_name;
+  }
```
xiaoxq, line 45: BTW, we should migrate to "string_view" for immutable input parameters: https://abseil.io/tips/1
We have imported it here https://github.com/ApolloAuto/apollo/blob/master/modules/common/util/future.h#L36
to make it compatible with both C++ 14 and 17.

storypku: I see.

PR #12031, opened 2020-08-04
Build: GLOG_TIMESTAMP renamed to FORMAT_TIMESTAMP

storypku:


modules/common/time/time_util.h
```diff
@@ -48,14 +45,6 @@ class TimeUtil {
     return unix_time;
   }
 
-  static double GetCurrentTime() {
```
storypku, line 51: We don't need so many GetCurrentTime implimentations.

PR #12036, opened 2020-08-04
Localization: optimize LRUCache

lfcarol:


modules/common/util/lru_cache.h
```diff
@@ -169,12 +173,29 @@ class LRUCache {
...
+    return true;
+  }
+
  private:
   static constexpr size_t kDefaultCapacity = 10;
 
   const size_t capacity_;
   size_t size_;
-  std::unordered_map<K, Node<K, V>> map_;
+  std::map<K, Node<K, V>> map_;
```
storypku, line 198: For what considerations should we change `std::unordered_map` to `std::map` ?

lfcarol: If use unordered_map, the key which isn't built-in type must provide a hash function to make the unique key, it is not convenient for us.

modules/localization/msf/common/util/base_map_cache_test.cc
```diff
@@ -0,0 +1,154 @@
...
+class NodeData {
+ public:
+  NodeData() : is_reserved_(false), name_("") {}
+  explicit NodeData(std::string name) : is_reserved_(false), name_(name) {}
+
+ public:
+  void SetIsReserved(bool b) { is_reserved_ = b; }
+  bool GetIsReserved() { return is_reserved_; }
+
+  void SetName(std::string name) { name_ = name; }
```
storypku, line 55: `const std::string& name` ?

modules/localization/msf/common/util/BUILD
```diff
@@ -66,6 +66,16 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "localization_msf_common_util_map_cache",
```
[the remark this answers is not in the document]
lfcarol: > Just name the target `base_map_cache`.

This is the localization module naming convention, and it should commit another pr to fix it.

modules/localization/msf/common/util/base_map_cache_test.cc
lfcarol: > `const std::string& name` ?

OK

modules/localization/msf/common/util/base_map_cache.h
```diff
@@ -0,0 +1,182 @@
...
+namespace msf {
+
+template <class Key, class Element>
+using LRUCache = ::apollo::common::util::LRUCache<Key, Element*>;
+
+/**@brief The data structure of the LRUCache. */
+template <class Key, class Element, class MapLRUCache = LRUCache<Key, Element>>
+class MapNodeCache {
+ public:
+  typedef std::function<bool(Element*)> DestroyFunc;
```
changsh726, line 35: prefer `using` to `typedef`

lfcarol: > prefer `using` to `typedef`

OK

modules/localization/msf/common/util/base_map_cache_test.cc
xiaoxq: Actually "string_view" is the new recommendation. But it's OK to stick to `const string&` for a while.

modules/localization/msf/common/util/BUILD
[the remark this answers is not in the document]
xiaoxq: Yep. Let's follow the convention in this PR and maybe fix the naming pattern in another PR.

xiaoxq:
LGTM. Please fix the conflict.

lfcarol:
> LGTM. Please fix the conflict.

OK

PR #12037, opened 2020-08-04
Fix 3js object hiding

unacao:


unacao:
The main fix is in hideArrayObjects(). The rest is just cleanup.

PR #12064, opened 2020-08-07
MockTime support 

storypku:
@rongguodong


cyber/time/clock.cc
```diff
@@ -25,36 +25,68 @@ namespace cyber {
 
 using GlobalData = ::apollo::cyber::common::GlobalData;
 
+using AtomicRWLock = ::apollo::cyber::base::AtomicRWLock;
```
storypku, line 28: Use AtomicRWLock to protect Clock from multi-threaded access.

cyber/time/clock.cc
```diff
@@ -25,36 +25,68 @@ namespace cyber {
...
+      break;
+    }
+    case ClockMode::MODE_CYBER: {
+      clock->mode_ = mode;
+      break;
+    }
+    default:
+      AERROR << "Unknown ClockMode: " << mode;
+  }
+  clock->mock_now_ = Time(0);
```
storypku, line 73: Whenenver `SetMode` is called, reset `mock_now_` to time point `zero`.

cyber/time/clock.h
```diff
@@ -19,9 +19,10 @@
```
storypku, line 19: `apollo::common::time::Clock` moved and renamed to `apollo::cyber::Clock` with multi-threaded safety support.

cyber/time/time.cc
```diff
@@ -83,6 +83,10 @@ bool Time::IsZero() const { return nanoseconds_ == 0; }
 
 uint64_t Time::ToNanosecond() const { return nanoseconds_; }
 
+uint64_t Time::ToMicrosecond() const {
```
storypku, line 86: Maybe a `double Time::ToDoubleMicrosecond() const` will be needed in the future.

modules/control/controller/lon_controller.cc
```diff
@@ -405,7 +399,7 @@ void LonController::ComputeLongitudinalErrors(
       vehicle_state->linear_velocity(), matched_point, &s_matched,
       &s_dot_matched, &d_matched, &d_dot_matched);
 
-  double current_control_time = Clock::NowInSeconds();
+  double current_control_time = Time::Now().ToSecond();
```
storypku, line 402: Here `Time` was used as otherwise it would fail lon_controller_test.

PR #12065, opened 2020-08-07
Audio: add audio direction estimation code & fix bugs/lint

Crysple:
## Changes

* Add direction detection code
* `AudioInfo` Init was not invoked. Besides, we did not have a `microphone_config` to init it at first. So I remove the Init function & get `microphone_config` from `audio_data` dynamically.
* Signal is a signed `int` & due to some C++ weird problem, I change to the line that combines two raw `char` to a `int_16` to `int16_t signal = ((int16_t(data[i])) << 8) | (0x00ff & data[i + 1]);`
* Remove unnecessary symbols suggested by `buildifier`

## TODO

* Test direction detection

### DIFF
--- a/modules/audio/audio_component.cc
+++ b/modules/audio/audio_component.cc
@@ -15,7 +15,7 @@
  *****************************************************************************/
 
 #include "modules/audio/audio_component.h"
-
+#include "modules/audio/inference/direction_detection.h"
 #include "modules/audio/proto/audio_conf.pb.h"
 
 namespace apollo {
@@ -23,8 +23,7 @@ namespace audio {
 
 using apollo::drivers::microphone::config::AudioData;
 
-AudioComponent::~AudioComponent() {
-}
+AudioComponent::~AudioComponent() {}
 
 std::string AudioComponent::Name() const {
   // TODO(all) implement
@@ -48,7 +47,11 @@ bool AudioComponent::Init() {
 
 bool AudioComponent::Proc(const std::shared_ptr<AudioData>& audio_data) {
   audio_info_.Insert(audio_data);
-  // TODO(all) get data for following online processings
+  AINFO << "Current direction is: "
+        << get_direction(
+               audio_info_.GetSignals(audio_data->microphone_config().chunk()),
+               audio_data->microphone_config().sample_rate(),
+               audio_data->microphone_config().mic_distance());
   return true;
 }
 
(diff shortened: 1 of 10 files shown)


modules/audio/audio_component.cc
```diff
@@ -48,7 +47,10 @@ bool AudioComponent::Init() {
 
 bool AudioComponent::Proc(const std::shared_ptr<AudioData>& audio_data) {
   audio_info_.Insert(audio_data);
-  // TODO(all) get data for following online processings
+  AINFO << "Getting direction...";
```
Capri2014, line 50: Please change it to ADEBUG for debug information

modules/audio/inference/direction_detection.h
```diff
@@ -0,0 +1,42 @@
...
+#include "cyber/cyber.h"
+#include "torch/torch.h"
+
+namespace apollo {
+namespace audio {
+
+constexpr float SOUND_SPEED = 343.2;
+constexpr float MIC_DISTANCE_4 = 0.08127;
+constexpr float MAX_TDOA_4 = MIC_DISTANCE_4 / SOUND_SPEED;
+constexpr float PI = 3.1415926535897932384626433832795;
```
Capri2014, line 34: Please use cmath pi instead of user defined.

modules/audio/inference/direction_detection.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include "ATen/ATen.h"
+#include "cyber/cyber.h"
+#include "torch/torch.h"
+
+namespace apollo {
+namespace audio {
+
+constexpr float SOUND_SPEED = 343.2;
+constexpr float MIC_DISTANCE_4 = 0.08127;
```
Capri2014, line 32: Should we move `MIC_DISTANCE_4` to mic config?

modules/audio/inference/direction_detection.h
Crysple: Sure

modules/audio/inference/direction_detection.h
Crysple: Yes I can do that

modules/audio/audio_component.cc
Crysple: Fixed

Crysple: I change it to `AINFO << "Current direction is: << ...`. I think it's a kind of information that's needed at least for now? (Maybe delete it before release?)

PR #12066, opened 2020-08-07
Audio: add structure for siren detection

HongyiSun:


Capri2014:
@HongyiSun try to rebase and avoid commits info like "Merge branch 'master' into master"

HongyiSun:
> @HongyiSun try to rebase and avoid commits info like "Merge branch 'master' into master"

Feel free to squash it into one commit

modules/audio/inference/siren_detection.cc
```diff
@@ -0,0 +1,86 @@
...
+
+  for (const auto& channel : signals) {
+    for (const auto& i : channel) {
+      *data++ = i;
+    }
+  }
+
+  torch::Tensor torch_input = torch::from_blob(data, {6, 1, 30075});
+  std::vector<torch::jit::IValue> torch_inputs;
+  torch_inputs.push_back(std::move(torch_input.to(device_)));
```
storypku, line 57: AIUI, the temporary value returned by `torch_input.to(device_)` is  `rvalue` reference. No need to call `std::move` explicitly here.

modules/audio/common/audio_gflags.cc
```diff
@@ -17,3 +17,6 @@
 #include "modules/audio/common/audio_gflags.h"
 
 DEFINE_int32(cache_signal_time, 3, "The time to cache signal");
+DEFINE_string(torch_siren_detection_model,
+              "/apollo/modules/audio/data/siren_detection_model.pt",
+              "Siren detection model file");
```
xiaoxq, line 22: Use constexpr variable if it's only used in one place and almost never change.
https://abseil.io/tips/45

modules/audio/inference/siren_detection.cc
```diff
@@ -0,0 +1,86 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *****************************************************************************/
+
+#include "modules/audio/inference/siren_detection.h"
+
+#include <vector>
+#include <utility>
+
+#include <omp.h>
```
xiaoxq, line 22: Put C headers before C++ headers.

modules/audio/common/audio_gflags.cc
HongyiSun: will change those flags for model path together...

PR #12080, opened 2020-08-10
Localization: fix eigen aligned to avoid run time error

lfcarol:


modules/localization/msf/local_integ/localization_integ_impl.h
```diff
@@ -49,6 +49,8 @@ class LocalizationLidarProcess;
  */
 
 class LocalizationIntegImpl {
+ public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
changsh726, line 53: add an empty line

PR #12102, opened 2020-08-11
audio: add moving detection and change signal data type to double

kechxu:


modules/audio/proto/audio.proto
```diff
@@ -18,6 +18,13 @@ syntax = "proto2";
 
 package apollo.audio;
 
+enum MovingResult {
```
Capri2014, line 21: might need to rename it before the onboard integration complete, "MovingResult" is not self-explanatory at all...

Capri2014: Also you need "Stationary" as well

Capri2014: “UNKNOWN" default assigned to 0

PR #12110, opened 2020-08-12
Pept: fix hdmap init error when cyber_launch start pept_lidar.launch

storypku:


modules/common/data/global_flagfile.txt
```diff
@@ -11,7 +11,7 @@
 
 --use_navigation_mode=false
 
---map_dir=/apollo/modules/map/data/sunnyvale
+--map_dir=/apollo/modules/map/data/sunnyvale_loop
```
storypku, line 14: The directory `/apollo/modules/map/data/sunnyvale` was empty in the container, while `/apollo/modules/map/data/sunnyvale_loop` isn't. So I guess maybe the former was renamed to the latter.
`cyber_launch start modules/perception/production/launch/perception_lidar.launch` seems to run smoothly without error messages. 

Please also evaluate the influence this change made to other modules' working.

storypku:
Need for another approval before merging this PR. @kechxu @Capri2014 @jinghaomiao

modules/common/data/global_flagfile.txt
Capri2014: “Sunnyvale” suppose to be the latest map that contains all the latest updates, fix and expanded areas, since when we start to use the “Sunnyvale_loop” and what is the difference between these two?

storypku: > “Sunnyvale” suppose to be the latest map that contains all the latest updates, fix and expanded areas, since when we start to use the “Sunnyvale_loop” and what is the difference between these two?

It seems that "Sunnyvale" was provided in our internal repo with a size of ~2G, while "Sunnyvale_loop" is public available with a size of ~400M.

Wait for more details from other team members.

storypku: Had a discussion with @Capri2014 in private, and "sunnyvale_loop" was chosen as the default map.

storypku: Does the change of default map from "sunnyvale" to "sunnyvale_loop" has any influence on Dreamview? @unacao @vlin17

vlin17: Dreamview supports all maps. This change is fine.

storypku: > Dreamview supports all maps. This change is fine.

Good!

PR #12118, opened 2020-08-13
Perception: fix the issue perception can't work

lfcarol:


ntutangyun:
Hi, does this mean that the detectionComponent is a replacement or an upgrade to the previous SegmentationComponent ?  thanks.

jeroldchen:
> Hi, does this mean that the detectionComponent is a replacement or an upgrade to the previous SegmentationComponent ? thanks.

@ntutangyun Yes, segmentation component has been removed. Perception module is using detection component, which is based on PointPillars method, for 3d object detection now.

PR #12119, opened 2020-08-13
Audio: add direction transformation code, refactor & fix bugs

Crysple:
* Fix bugs in direction detection -- didn't notice LibTorch does not support complex type, etc.
* add direction transformation code
* Refactor the direction_detection to a class

### DIFF
--- a/modules/audio/inference/direction_detection.h
+++ b/modules/audio/inference/direction_detection.h
@@ -19,21 +19,60 @@
 
 #include <algorithm>
 #include <cmath>
+#include <memory>
+#include <string>
+#include <utility>
 #include <vector>
 
+#include "Eigen/Eigen"
+// Eigen 3.3.7: #define ALIVE (0)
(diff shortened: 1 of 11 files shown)


modules/audio/inference/direction_detection.h
```diff
@@ -19,21 +19,59 @@
...
+// Eigen 3.3.7: #define ALIVE (0)
+// fastrtps: enum ChangeKind_t { ALIVE, ... };
+#if defined(ALIVE)
+#undef ALIVE
+#endif
+
 #include "ATen/ATen.h"
 #include "cyber/cyber.h"
+#include "modules/common/proto/geometry.pb.h"
 #include "torch/torch.h"
```
storypku, line 37: Header orders: `torch/torch.h` should be right after `ATen/ATen.h`. And add a blank line  after it.

Crysple: Thank you I have fixed it.

PR #12120, opened 2020-08-13
run_ci: fix -t bug and other container  bug

lx18233184051:


lx18233184051:
> Please check whether `Yolo3D` and `local_third_party` volumes should be started for CI.

ok, clean cache and test.

PR #12121, opened 2020-08-13
sim_control prototype

Fla3inH0tCheet0s:
Here's a follow up from https://github.com/ApolloAuto/apollo/pull/11556.


Capri2014:
Why we want to make this change?

Fla3inH0tCheet0s:
> Why we want to make this change?

This MR addresses https://github.com/ApolloAuto/apollo/issues/9321. It is meant to separate simulation logic from visualization backend. Once separated, this module could be used in isolation. For example, it allows for integration tests around planning and routing without a frontend.

Capri2014:
> > Why we want to make this change?
> 
> This MR addresses #9321. It is meant to separate simulation logic from visualization backend. Once separated, this module could be used in isolation. For example, it allows for integration tests around planning and routing without a frontend.

ok, thanks.

Capri2014:
> Observed several issues when trying the new sim_control prototype:
> 
> 1. Dreamview shows the following message when sim_control module is on. Any idea why @xiaoxq @Capri2014 ?
>    Screen Shot 2020-08-26 at 4 19 41 PM
> 2. data/log/sim_control.INFO keeps printing the following errors:
>    E0826 16:15:31.122936 18703 sim_control_component.cc:325] Failed to calculate next point with perfect control model
>    E0826 16:15:31.132977 18701 sim_control_component.cc:361] First trajectory point is a future point!
> 3. even though ego-car can follow the route, it doesn't go forward smoothly, seems like the car is being dragged backward a little bit every sec.

For 2 and 3 I will bet on the time sequence mismatch, but can only verify if having a record. For 1, usually when it triggered the safetyMode it also recorded the detailed reasons, what is in it?

lx18233184051:
This pr has opened so long, maybe should be closed. you cloud reopen if necessary:)

PR #12125, opened 2020-08-13
apollo test: fix all fail test

lx18233184051:


modules/localization/ndt/ndt_locator/BUILD
```diff
@@ -52,7 +52,7 @@ cc_test(
 
 cc_test(
     name = "ndt_solver_test",
-    size = "small",
+    size = "enormous",
     timeout = "short",
```
storypku, line 56: So, `size="large"` is not enough ?
See https://docs.bazel.build/versions/master/be/common-definitions.html for detailed comparison of TEST_SIZE and TIMEOUT.

lx18233184051: since source is not a limit, why not make test time shorter?

storypku: > since source is not a limit, why not make test time shorter?

It's generally considered good practice to be light-weight if possible rather than heavy-weight.

PR #12147, opened 2020-08-17
mpc_osqp: malloc judge nullptr

lx18233184051:


modules/common/math/mpc_osqp.cc
```diff
@@ -220,47 +220,56 @@ OSQPSettings *MpcOsqp::Settings() {
...
   OSQPSettings *settings =
       reinterpret_cast<OSQPSettings *>(c_malloc(sizeof(OSQPSettings)));
-  osqp_set_default_settings(settings);
-  settings->polish = true;
-  settings->scaled_termination = true;
-  settings->verbose = false;
-  settings->max_iter = max_iteration_;
-  settings->eps_abs = eps_abs_;
-  return settings;
+  if (settings != nullptr) {
```
storypku, line 223: Better writen as follows:
```
if (settings == nullptr) {
   return nullptr;
}
.... // blahblahblah
```
Return early for simple cases.

modules/common/math/mpc_osqp.cc
```diff
@@ -220,47 +220,56 @@ OSQPSettings *MpcOsqp::Settings() {
...
+    data->A =
+        csc_matrix(state_dim_ * (horizon_ + 1) + state_dim_ * (horizon_ + 1) +
+                       control_dim_ * horizon_,
+                   kernel_dim, A_data.size(), CopyData(A_data),
+                   CopyData(A_indices), CopyData(A_indptr));
+    ADEBUG << "Get A matrix";
+    data->l = lowerBound_.data();
+    data->u = upperBound_.data();
+    return data;
+  } else {
```
storypku, line 270: Ditto.

PR #12149, opened 2020-08-17
hdmap_common: improve search efficiency

lx18233184051:


modules/map/hdmap/hdmap_common.cc
```diff
@@ -308,8 +308,15 @@ double LaneInfo::GetWidthFromSample(
   }
   int low = 0;
   int high = static_cast<int>(samples.size());
+  int mid;
   while (low + 1 < high) {
-    const int mid = (low + high) / 2;
+    double tmp = samples[high].first - samples[low].first;
```
changsh726, line 313: Seems like `samples[samples.size()]` will out of bounds

modules/dreamview/backend/simulation_world/BUILD
```diff
@@ -38,7 +38,7 @@ cc_library(
 
 cc_test(
     name = "simulation_world_service_test",
-    size = "small",
+    size = "large",
```
storypku, line 41: Could you please note the reason for the change here?

modules/map/hdmap/hdmap_common.cc
storypku: > Seems like `samples[samples.size()]` will out of bounds

Not "seems like",  high=N will surely make visit to the element out of bounds.

modules/planning/reference_line/BUILD
```diff
@@ -127,7 +127,7 @@ cc_test(
 
 cc_test(
     name = "discrete_points_reference_line_smoother_test",
-    size = "small",
+    size = "large",
```
storypku, line 130: Ditto. Please justify why "small" or "medium" is not enough ?

modules/dreamview/backend/simulation_world/BUILD
lx18233184051: modify hdmap_common.cc, those tests become timeout, maybe those used  cache before.

modules/planning/reference_line/BUILD
lx18233184051: I have tested, at least two minutes

modules/map/hdmap/hdmap_common.cc
```diff
@@ -308,8 +308,16 @@ double LaneInfo::GetWidthFromSample(
   }
   int low = 0;
   int high = static_cast<int>(samples.size());
+  int mid;
   while (low + 1 < high) {
-    const int mid = (low + high) / 2;
+    double tmp = samples[high - 1].first - samples[low].first;
+    if (tmp > 0) {
+      mid = low + (s - samples[low].first) /
+                      (samples[high - 1].first - samples[low].first) *
```
changsh726, line 316: why not use `tmp` instead

PR #12151, opened 2020-08-17
Scripts: add perf.sh to do performance profiling

changsh726:


scripts/perf.sh
```diff
@@ -0,0 +1,134 @@
...
+
+function main() {
+  if [ "$#" -eq 0 ]; then
+    print_usage
+    exit 1
+  fi
+
+  local opt="$1"
+  shift
+  case "${opt}" in
```
storypku, line 115: Add a "clean" option to cleanup  outputs

scripts/perf.sh
```diff
@@ -0,0 +1,134 @@
...
+    rm -f "${OUTPUT_DIR}${PERF_DATA}"
+  fi
+}
+
+# Note(All): if permission denied, run `echo 1 > /proc/sys/kernel/perf_event_paranoid` as root user out of docker
+function gen_perf_data() {
+  _check_perf
+  local module=$1
+  _check_out_dir
+  ${PERF_CMD} record -e cpu-clock -o "${OUTPUT_DIR}${PERF_DATA}" -g mainboard -d \
```
lx18233184051, line 73: what does mainboard mean?

changsh726: use mainboard to start the module

lx18233184051: mainboard is a process ?

changsh726: yep

PR #12163, opened 2020-08-18
Perception: add conf for lgsvl dag

jeroldchen:
Set `lidar_query_tf_offset` 200 in velodyne 128 detection for lgsvl, to solve the problem of matching timestamps.

### DIFF
--- a/modules/perception/production/conf/perception/lidar/velodyne128_detection_conf_lgsvl.pb.txt
+++ b/modules/perception/production/conf/perception/lidar/velodyne128_detection_conf_lgsvl.pb.txt
@@ -1,0 +1,5 @@
+sensor_name: "velodyne128"
+enable_hdmap: true
+lidar_query_tf_offset: 200
+lidar2novatel_tf2_child_frame_id: "velodyne128"
+output_channel_name: "/perception/inner/DetectionObjects"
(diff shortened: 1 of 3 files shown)


storypku:
> Set `lidar_query_tf_offset` 200 in velodyne 128 detection for lgsvl, to solve the problem of matching timestamps.

cc @rongguodong @lemketron , although it seems not working on my local machine.

Capri2014:
Why change the query offset? Correct me if I am wrong, but it looks like the simulation backend is not having proper timing sequence since this modules works in vehicle but not in LGSVL? Would this fix be a bit tricky and later let's assume, we change the Lidar output frequencies, would this offset needs to be changed again?

jinghaomiao:
> Why change the query offset? Correct me if I am wrong, but it looks like the simulation backend is not having proper timing sequence since this modules works in vehicle but not in LGSVL? Would this fix be a bit tricky and later let's assume, we change the Lidar output frequencies, would this offset needs to be changed again?

@jeroldchen Agree with @Capri2014 . Let's make sure perception works on the vehicle, and keep troubleshooting this issue to find the root cause. For LGSVL users, we can make a note of the workaround and let them know. Please check in other needed features though, but leave the query time intact until we know a reason.

PR #12170, opened 2020-08-19
cyber_recorder close file asynchronously

Fla3inH0tCheet0s:
When `RecordWriter` calls `RecordFileWriter::Close`, it has to write out its data. There's a sleep in there for 100ms, if it doesn't write fast enough. during that time new messages are coming in. If too many messages come in, cyber's callback threads will be exhausted and messages will be stuck in the queue. The result is messages with the same received timestamp.

This proposed solution is to put the destructor call on another async task. The next time the file is split, the previous async operation cleaned up and a new async operation is created via `~future()` and `async` operation respectively. Similarly, std::future will clean itself up on `~RecordWriter`.

### DIFF
--- a/cyber/record/record_writer.cc
+++ b/cyber/record/record_writer.cc
@@ -16,8 +16,11 @@
 
 #include "cyber/record/record_writer.h"
 
+#include <chrono>
+#include <future>
 #include <iomanip>
 #include <iostream>
+#include <utility>
 
 #include "cyber/common/log.h"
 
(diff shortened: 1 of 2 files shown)


lx18233184051:
two unit test failed。maybe you need to change BUILD: size small -> large.
//modules/localization/msf/local_map/ndt_map:ndt_map_test 
//modules/perception/lidar/lib/detection/lidar_point_pillars:point_pillars_test

storypku:
> two unit test failed。maybe you need to change BUILD: size small -> large.
> //modules/localization/msf/local_map/ndt_map:ndt_map_test
> //modules/perception/lidar/lib/detection/lidar_point_pillars:point_pillars_test

Correct me if I am wrong, but it seems that this PR has nothing to do with the two test failures when running CI. 
I had restarted the CI process manually.

PR #12182, opened 2020-08-20
Drivers: initial cleaning for camera and video drivers

Capri2014:


scripts/camera_and_video.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# run command_name module_name
+function run() {
+    case $1 in
+        start)
+            start
+            ;;
+        stop)
+            stop
+            ;;
```
storypku, line 47: What about adding a `restart` option here?

modules/drivers/camera/dag/camera.dag
```diff
@@ -28,74 +17,6 @@ module_config {
         config_file_path : "/apollo/modules/drivers/camera/conf/camera_front_12mm.pb.txt"
       }
     }
-    components {
```
michael4338, line 31: why remove this?

scripts/camera_and_video.sh
Capri2014: Good suggestion! We may prefer though to create a separate PR to add `restart` options for all related scrips together.

modules/drivers/camera/dag/camera.dag
Capri2014: 1. No front fisheye installed on vehicle by default. 2. The `camera_compress` are not working as of now so I remove them by default, the `video_compress` is still functional properly under `video/dag/vide.dag`.

PR #12183, opened 2020-08-20
Audio: add siren_detection

HongyiSun:


Capri2014:
@HongyiSun how large is the model as of now? You may want to make it a volume if it is too large.

HongyiSun:
> @HongyiSun how large is the model as of now? You may want to make it a volume if it is too large.

It's about 188MB as for now.

PR #12187, opened 2020-08-20
Docker: infra upgraded with cudnn8 and tensorrt-7.1

storypku:
TensorRT 7.1+ depends on cudnn-8. However, there was  some errors when compiling libtorch 1.5.1 agains cudnn8. 
So libtorch bumped to 1.6.0 (official) to bypass those bugs we fixed in libtorch 1.5 before. However, libtorch 1.6.0 made four testcases fail.

The 4 failed testcases are:

```
 //modules/perception/lidar/lib/detection/lidar_point_pillars:point_pillars_test
//modules/prediction/evaluator:evaluator_manager_test
//modules/prediction/evaluator/vehicle:cruise_mlp_evaluator_test
//modules/prediction/evaluator/vehicle:junction_mlp_evaluator_test
```

The system team will continue to investigate into this tomorrow.

@Capri2014 @jeroldchen @panfengsu @jinghaomiao @kechxu


jinghaomiao:
I wouldn't suggest to do a major upgrade for critical dependencies consider that we have limited time before the release for testing and adaptation. Let's plan it after 6.0 release

storypku:
> I wouldn't suggest to do a major upgrade for critical dependencies consider that we have limited time before the release for testing and adaptation. Let's plan it after 6.0 release

Good point.  Originally we only want to bump TensorRT to 7.1 to have ONNX 1.6.0+ supported. Didn't expect it has so many dependencies...

storypku:
Restored the `docker_dev` branch for experiment with new features. 

> > I wouldn't suggest to do a major upgrade for critical dependencies consider that we have limited time before the release for testing and adaptation. Let's plan it after 6.0 release
> 
> Good point. Originally we only want to bump TensorRT to 7.1 to have ONNX 1.6.0+ supported. Didn't expect it has so many dependencies...

PR #12190, opened 2020-08-20
scripts: add restart function

lx18233184051:


lfcarol (approved):

scripts/velodyne_16.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 ###############################################################################
-# Copyright 2018 The Apollo Authors. All Rights Reserved.
+# Copyright 2020 The Apollo Authors. All Rights Reserved.
```
storypku, line 4: Keep the original.

storypku (changes requested):
Leave the years in the copyright as it is.

scripts/velodyne_16.sh
lx18233184051: ok

storypku (approved):

PR #12193, opened 2020-08-20
Audio: fix bugs & improve audio saver

Crysple:
* Fix a bug that it is small endianness when assembling bytes.
* Since the frequency is increased, the duration of a single wave is too small. I pack them together. When finish playing the record, use ctrl+c to stop the audio saver program and it will save all frames to a wave file.

### DIFF
--- a/modules/audio/inference/direction_detection.cc
+++ b/modules/audio/inference/direction_detection.cc
@@ -60,7 +60,7 @@ double DirectionDetection::EstimateDirection(
   }
 
   double tau0, tau1;
-  int theta0, theta1;
+  double theta0, theta1;
   const double max_tau = mic_distance / kSoundSpeed;
   tau0 = GccPhat(channels_ts[0], channels_ts[2], sample_rate, max_tau, 1);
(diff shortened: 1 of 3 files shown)


modules/audio/inference/direction_detection.cc
```diff
@@ -69,12 +69,12 @@ double DirectionDetection::EstimateDirection(
 
   int best_guess = 0;
   if (fabs(theta0) < fabs(theta1)) {
-    best_guess = theta1 > 0 ? (theta0 + 360) % 360 : (180 - theta0);
+    best_guess = theta1 > 0 ? fmod(theta0 + 360, 360) : (180 - theta0);
```
storypku, line 72: Please use namespace specifier here: `std::fmod` rather than bare `fmod`.

storypku (changes requested):

modules/audio/inference/direction_detection.cc
Crysple: Got it

Crysple: Fixed

PR #12218, opened 2020-08-22
Docker:  reduced {cyber,dev}.x86_64 images with lightweight clang installer

storypku:


changsh726 (approved):
It seems that this pull request has already included changes in #12219 and #12220

storypku:
> It seems that this pull request has already included changes in #12219 and #12220

Yep. I cherry-picked them beforehand to avoid potential conflict. Luckily, It seems to introduce no extra commit as can be seen from git log history.

PR #12229, opened 2020-08-24
Perception: fix invalid symlinks

jeroldchen:


storypku (approved):
Checked out this PR and run `find modules -xtype l` again, and those broken symlinks are gone.
So approve this PR.

PR #12238, opened 2020-08-24
Audio: add audio_event and proto refactor

Capri2014:


Capri2014:
@Crysple

Capri2014:
@kechxu do you think `moving_result` is needed or no?

PR #12255, opened 2020-08-25
audio: add siren_is_on into audio event

kechxu:


kechxu:
> Can you expand this commit and add the EmergencyVehicle Type and other enums type (type definition might goes to audio_common) we just talked about?

Sure

PR #12271, opened 2020-08-26
Localization: fix the crash that gps message is ealier than imu message

lfcarol:


modules/localization/msf/msf_localization.cc
```diff
@@ -306,6 +306,9 @@ void MSFLocalization::OnGnssHeading(
 
 void MSFLocalization::OnGps() {
   std::unique_lock<std::mutex> lock(mutex_imu_msg_);
+  if (!raw_imu_msg_) {
+    return;
+  }
```
jinghaomiao, line 311: A better practice is to check the message before the lock.

lfcarol: OK

lfcarol: > A better practice is to check the message before the lock.

OK

PR #12274, opened 2020-08-27
Scripts: apollo.sh prints usage if no args are specified

changsh726:


Capri2014:
lol, how come we did not have this before....

storypku:
> lol, how come we did not have this before....

Haha, as there is a dedicated `./apollo.sh -h|--help` for that.

PR #12277, opened 2020-08-27
Add drivers to support RoboSense LiDAR

yfcube:
RS16, RS32, RS128, RSBP are now supported

### DIFF
--- a/apollo.sh
+++ b/apollo.sh
@@ -129,6 +129,7 @@ function _usage() {
 function main() {
     apollo_env_setup
     if [ "$#" -eq 0 ]; then
+        _usage
         exit 0
     fi
     local build_sh="${APOLLO_ROOT_DIR}/scripts/apollo_build.sh"


modules/drivers/robosense/conf/rs128.pb.txt
```diff
@@ -0,0 +1,15 @@
+model: "RS128"
+frame_id: "rslidar_128"
+ip: "192.168.1.200"
```
Capri2014, line 3: Why the 128, 32, 16 share the same ip, msop and difop port? If you actually recommend setting them to different ports please differentiate them in default files.

yfcube: We leave the config file as the way it is since all models of RoboSense LiDAR do have the same default settings for ip, msop and difop port?

storypku:
> ```
> RS16, RS32, RS128, RSBP are now supported
> ```

First of all, thanks for the PR!

Please see my review opinion for PR #12281 .  
We prefer one single coding style on Apollo repo. 
Make CI passed before checking in.


Regards,
The Apollo Team

PR #12282, opened 2020-08-27
Build: Improved CPU & GPU build/test/coverage support

storypku:
Build: improved CPU/GPU build/test/coverage support
    
    1) Introduce USE_GPU_TARGET (inside docker) to determine if device/tools/libraries is ready to be GPU capable
    2) Use of USE_GPU_HOST now only in cyber|dev_start.sh
    
    For build,
    
    1) Mixed used of '--config=cpu' and '--config=gpu' not allowed
    2) Make CPU build (i.e., '--config=cpu') on GPU capable machine possible
    3) Error and exit attempting to run GPU build (`--config=gpu`) on CPU-only machines
    4) Default to USE_GPU_TARGET if none of cpu|gpu build specified

    Build: --config=cpu/gpu not written into .apollo.bazelrc
    {cyber|dev}_start.sh: introduce EnvVar USE_GPU_HOST
    Bazel: bugfix for libtorch_gpu.BUILD
    Bazel: bugfix for USE_GPU to accord with bazelrc settings

### DIFF
--- a/apollo.sh
+++ b/apollo.sh
@@ -15,7 +15,7 @@ USE_ESD_CAN=false
 function check_architecture_support() {
     if [[ "${SUPPORTED_ARCHS}" != *" ${ARCH} "* ]]; then
         error "Unsupported CPU arch: ${ARCH}. Currently, Apollo only" \
-              "supports running on the following CPU archs:"
+            "supports running on the following CPU archs:"
         error "${TAB}${SUPPORTED_ARCHS}"
         exit 1
     fi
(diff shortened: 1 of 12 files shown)


unacao:
now I see 'USE_GPU', 'USE_GPU_HOST' and 'USE_GPU_TARGET', which one should we use?

storypku:
> now I see 'USE_GPU', 'USE_GPU_HOST' and 'USE_GPU_TARGET', which one should we use?

Good question. Let me try to clarify the use cases and differences of these `USE_GPU`s.


- USE_GPU_HOST was used to determine whether host environment could enable GPU devices/capabilities when {cyber,dev}_start.sh (Where `USE_GPU_HOST` was defined and got passed to docker container).
 
- USE_GPU_TARGET was used to check whether cyber/dev container is GPU capable.  For example,  A container started with USE_GPU_HOST=1 may have USE_GPU_TARGET=0 if CUDA libraries was not available inside docker. It was defined in `scripts/apollo.bashrc` and is available also in `scripts/apollo_base.sh`. (Since `apollo_base.sh` sourced `apollo.bashrc`).

- USE_GPU serves for two purposes. 
1) Used inside `apollo_build/test/coverage.sh` to determine whether should run CPU build or GPU build. Please note that It no longer exported as an Env variable.  Scripts that previously used `USE_GPU` **should** use `USE_GPU_TARGET` instead to determine whether the container is GPU capable.
2) Used as a C/C++ preprocessor macro for C++ headers/sources in Apollo repo.

So know you can understand that why `USE_GPU_HOST` and `USE_GPU_TARGET` was both exported as Env variables inside docker, while `USE_GPU` wasn't.

Hope my answer helps.

PR #12322, opened 2020-08-31
audio: add degree into audio detection output

kechxu:


Capri2014:
@kechxu Add unit as comment for audio direction

PR #12326, opened 2020-09-01
docker/build: bos download

lx18233184051:


storypku:
Please keep those links that don't start with https://apollo-platform-system.bj.bcebos.com.

lx18233184051:
> Please keep those links that don't start with https://apollo-platform-system.bj.bcebos.com.

whether change those link,both is  ok. because it's for us to make images.
image

PR #12342, opened 2020-09-01
Format: use scripts/mdfmt.sh to format md/json/yml files

storypku:


.prettierrc.json
```diff
@@ -1,3 +1,9 @@
 {
-  "proseWrap": "always"
+  "proseWrap": "always",
```
storypku, line 2: So my problem here is, is this configuration in accord with our frontend coding style? @unacao @vlin17 @lj0919 ?

vlin17: Is this for frontend? We have module/dreamview/frontend/.eslintrc to define the style.

storypku: > Is this for frontend? We have module/dreamview/frontend/.eslintrc to define the style.

Nope. We  use prettier to format markdown/json/yaml files. However, javascript files can also be formatted using prettier, and it seems that prettier can be used in combination with eslint.

changsh726 (approved):

PR #12349, opened 2020-09-02
Best practice Hesai Driver

storypku:
Demo PR to show Best Practices on Building C++ with Bazel.

1) Avoid using "using xxx;"  in header files.
2) Split targets into smaller ones, as described here: https://docs.bazel.build/versions/master/bazel-and-cpp.html#build-files
3) Shared `HESAI_COPTS` variable, See: https://docs.bazel.build/versions/master/skylark/tutorial-sharing-variables.html
4) Include necessary headers **ONLY**
5) Prefer C++ headers to C headers. E.g., using `<ctime>` rather than `<time.h>`
6) Pay attention to header orders. Run `scripts/apollo_format.sh modules/drivers/hesai` to format C/C++ files and BAZEL BUILD files.
7) Don't Over/Under estimate your dependency. No more, No less
8) Using `scripts/mdfmt.sh path/or/file` to format markdown/json/yaml files.

@xiaoxq @changsh726 @lfcarol @lx18233184051 @Capri2014 @jinghaomiao

### DIFF
--- a/modules/drivers/hesai/BUILD
+++ b/modules/drivers/hesai/BUILD
@@ -7,38 +7,122 @@ cc_binary(
     name = "libhesai_driver_component.so",
     linkshared = True,
     linkstatic = False,
-    deps = [":driver"],
+    deps = [
+        ":hesai_component",
+        ":hesai_convert_component",
+    ],
+)
+
+cc_library(
+    name = "hesai_convert_component",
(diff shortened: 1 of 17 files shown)


changsh726 (approved):

a982865151:
真心求学习v2x的代码撰写，万分感谢

PR #12360, opened 2020-09-03
Change scripts which use "pgrep" to avoid PID 1.

rongguodong:
Apollo uses `pgrep -c -f` to determine if processes are running. Unfortunately, this finds them in the inline container startup command. This is a problem when we try to run Apollo in our own container on cloud where the container startup command contains some strings matching Apollo's pgrep string.

This change uses of `pgrep -f` to ignore when PID 1 is output by filtering the output with `| grep -v '^1$'`, and use `| wc -l` to count the lines of the output.

### DIFF
--- a/scripts/apollo_base.sh
+++ b/scripts/apollo_base.sh
@@ -136,7 +136,7 @@ function decide_task_dir() {
 function is_stopped_customized_path() {
   MODULE_PATH=$1
   MODULE=$2
-  NUM_PROCESSES="$(pgrep -c -f "modules/${MODULE_PATH}/launch/${MODULE}.launch")"
+  NUM_PROCESSES="$(pgrep -f "modules/${MODULE_PATH}/launch/${MODULE}.launch" | grep -cv '^1$')"
   if [ "${NUM_PROCESSES}" -eq 0 ]; then
     return 1
   else
--- a/scripts/camera.sh
+++ b/scripts/camera.sh
@@ -25,7 +25,7 @@ source "${DIR}/apollo_base.sh"
 function start() {
   LOG="${APOLLO_ROOT_DIR}/data/log/camera.out"
   CMD="cyber_launch start /apollo/modules/drivers/camera/launch/camera.launch"
-  NUM_PROCESSES="$(pgrep -c -f "modules/drivers/camera/dag/camera.dag")"
+  NUM_PROCESSES="$(pgrep -f "modules/drivers/camera/dag/camera.dag" | grep -cv '^1$')"
   if [ "${NUM_PROCESSES}" -eq 0 ]; then
     eval "nohup ${CMD} </dev/null >${LOG} 2>&1 &"
   fi
--- a/scripts/camera_and_video.sh
+++ b/scripts/camera_and_video.sh
@@ -25,7 +25,7 @@ source "${DIR}/apollo_base.sh"
 function start() {
   LOG="${APOLLO_ROOT_DIR}/data/log/camera_and_video.out"
   CMD="cyber_launch start /apollo/modules/drivers/camera/launch/camera_and_video.launch"
-  NUM_PROCESSES=$(expr "$(pgrep -c -f "modules/drivers/camera/dag/camera.dag")" + "$(pgrep -c -f "modules/drivers/video/dag/video.dag")")
+  NUM_PROCESSES=$(expr "$(pgrep -f "modules/drivers/camera/dag/camera.dag" | grep -cv '^1$')" + "$(pgrep -f "modules/drivers/video/dag/video.dag" | grep -cv '^1$')")
   if [ "${NUM_PROCESSES}" -eq 0 ]; then
     eval "nohup ${CMD} </dev/null >${LOG} 2>&1 &"
   fi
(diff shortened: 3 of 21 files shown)


storypku:
> Apollo uses **pgrep -c -f** to determine if processes are running. Unfortunately, this finds them in the inline container startup command. This is a problem when we try to run Apollo in our own container on cloud where the container startup command contains some strings matching Apollo's pgrep string.
> 
> This change uses of **pgrep -f** to ignore when PID 1 is output by filtering the output with **| grep -v '^1$'**, and use **| wc -l** to count the lines of the output.

Can you further explain the situation in your case? I didn't quite get what you means by saying "this finds them in the inline container startup command". It seems that you want to exclude PID 1 ( On a normal Linux host, pid 1 is for the init process). Maybe it's related to docker ? Can you give us a minimal environment to reproduce the case?

BTW,  we can use `grep -cv '^1$' ` to count all pids that are not 1.

storypku (changes requested):
Please provide more details and change `grep -v | wc -l` operation to `grep -cv`.

rongguodong:
We use an inline startup script when launching our Apollo containers under Kubernetes. (A container's startup script is always PID 1 under Kubernetes.) What is meant by "inline" is that script is passed as one long string to "bash -c", and therefore appears in its entirety, including any comments, in the command line that "pgrep -f" searches. While developing the script, things inexplicably stopped working. Imagine our surprise when, after a many hours of debugging, it turned out that the cause was a comment mentioning "cyber_recorder" had been added to the script and the Apollo code now considered it to be running.

Because it matches substrings anywhere in the command line, using "pgrep -f" makes it easy for there to be unexpected side-effects. For example, "pgrep -f modules/dreamview/launch/dreamview.launch" outputs the PID for "vi modules/dreamview/launch/dreamview.launch". And if the container is run with --pid=host, it would find such a vi running on the host outside of the container! It would be much safer to use "pgrep -x", but it was more expedient for us to just filter out PID 1.

rongguodong:
@storypku If you want to reproduce it locally, you may change dev_start.sh as following:
* Remove [this line](https://github.com/ApolloAuto/apollo/blob/master/docker/scripts/dev_start.sh#L530); and
* Change [this line](https://github.com/ApolloAuto/apollo/blob/master/docker/scripts/dev_start.sh#L533) to the following line:
/bin/bash -c ": cyber_recorder; sleep inf"

After you do "dev_start.sh" and "dev_into.sh", "ps aux" will show something like:
[guodong@in-dev-docker:/apollo]$ ps aux
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root         1  0.1  0.0  18376  3004 pts/1    Ss+  14:45   0:00 /bin/bash -c : cyber_recorder; sleep 
root         6  0.0  0.0   4532   880 pts/1    S+   14:45   0:00 sleep inf
guodong     60  1.0  0.0  19444  4448 pts/2    Ss   14:45   0:00 /bin/bash
guodong     88  0.0  0.0  34404  2856 pts/2    R+   14:45   0:00 ps aux

And if you use pgrep -f "cyber_recorder", you will get "1".

rongguodong:
> Please provide more details and change `grep -v | wc -l` operation to `grep -cv`.

Done.

storypku:
> We use an inline startup script when launching our Apollo containers under Kubernetes. (A container's startup script is always PID 1 under Kubernetes.) What is meant by "inline" is that script is passed as one long string to "bash -c", and therefore appears in its entirety, including any comments, in the command line that "pgrep -f" searches. While developing the script, things inexplicably stopped working. Imagine our surprise when, after a many hours of debugging, it turned out that the cause was a comment mentioning "cyber_recorder" had been added to the script and the Apollo code now considered it to be running.
> 
> Because it matches substrings anywhere in the command line, using "pgrep -f" makes it easy for there to be unexpected side-effects. For example, "pgrep -f modules/dreamview/launch/dreamview.launch" outputs the PID for "vi modules/dreamview/launch/dreamview.launch". And if the container is run with --pid=host, it would find such a vi running on the host outside of the container! It would be much safer to use "pgrep -x", but it was more expedient for us to just filter out PID 1.

Thanks for your details explanation. Maybe the team should modify these scripts to avoid the `vim modules/dreamview/launch/dreamview.launch` case you mentioned in next PRs.

Regards,
Jiaming Liu

storypku (approved):

rongguodong:
Is `--pid=host` only needed for debugging? If you remove this line, you cannot see processes running outside of the docker.
For our use case, we just need to run Apollo on the cloud and do not need any debugging. Is it safe to simply remove this line?

storypku:
> Is `--pid=host` only needed for debugging? If you remove this line, you cannot see processes running outside of the docker.
> For our use case, we just need to run Apollo on the cloud and do not need any debugging. Is it safe to simply remove this line?

I will do some investigation first. 
Would you please file an issue for the `--pid=host` thing ?
Thank you.

HongyiSun:
There is a problem cause by escape characters in pythons, where `'` in the python string was misinterpreted. I have made a PR #12443 on fixing this.

PR #12363, opened 2020-09-03
Canbus: modify some protocol name and add some units in vehicle devkit

macDure:
devkit
1. modify some protocol name；
2. add some units in vehicle devkit.


Capri2014 (changes requested):
Do not change proto name and field, if the interface changes, mark old ones as deprecated and add new ones.

jinghaomiao (approved):

PR #12400, opened 2020-09-04
Planning: reuse path when path reference is not valid

sjiang2018:


modules/planning/conf/planning.conf
```diff
@@ -44,7 +44,7 @@
 --use_st_drivable_boundary=true
 
 #--enable_skip_path_tasks=true
-#--enable_reuse_path_in_lane_follow=true
```
JasonZhou404, line 47: Does the conf change necessarily affect regular lane following scenario?

JasonZhou404 (approved):

PR #12407, opened 2020-09-07
Scripts: print apollo evironment settings only when apollo.sh config

changsh726:


storypku (commented):
I would suggest keeping Env settings for `bazel build/test/coverage` etc. It may help us track issues.

Suppose when someone opened an issue on `bazel build`, we don't need to ask him/her to provide `./apollo.sh config` info if env settings were reserved in `bazel build`.

apollo.sh
```diff
@@ -72,6 +72,12 @@ function apollo_env_setup() {
     APOLLO_ENV="${APOLLO_ENV} USE_ESD_CAN=${USE_ESD_CAN}"
     # Add more here ...
 
+    if [ ! -f "${APOLLO_ROOT_DIR}/.apollo.bazelrc" ]; then
+        env ${APOLLO_ENV} bash "${APOLLO_ROOT_DIR}/scripts/apollo_config.sh" --noninteractive
+    fi
+}
+
+function print_env() {
     info "Apollo Environment Settings:"
```
storypku, line 81: Although I would suggest keep the `printing env summary` section, moving  the "apollo_config.sh" stuff before printing environ summary is meaningful.

PR #12389, opened 2020-09-04
Doc: DKIT add control profiling online guide

macDure:

PR #12408, opened 2020-09-07
Doc: DKIT: add control profiling online guide (#12389)

macDure:
DKIT add control profiling online guide


macDure:
which was cherry-picked from master.

PR #12442, opened 2020-09-08
Planning: added stitching to path reference

sjiang2018:


JasonZhou404 (dismissed):

modules/planning/tasks/deciders/path_reference_decider/path_reference_decider.cc
```diff
@@ -162,7 +163,12 @@ Status PathReferenceDecider::Process(Frame *frame,
     return Status::OK();
   }
   std::vector<PathPoint> path_reference;
-  ConvertTrajectoryToPath(learning_model_trajectory, &path_reference);
+  // stiching
```
JasonZhou404, line 166: misspelling

JasonZhou404: maybe using  the expression like "adjusting trajectory init point" ? stitching is not an accurate word here

sjiang2018: Fixed. Thanks!

JasonZhou404 (approved):

PR #12445, opened 2020-09-09
Dreamview|Build: split main.cc from dreamview's backend library

storypku:


modules/dreamview/main.cc
```diff
@@ -20,7 +20,7 @@
 
 int main(int argc, char *argv[]) {
   google::ParseCommandLineFlags(&argc, &argv, true);
-  // add by caros for dv performance improve
+  // add by caros to improve dv performance
```
jinghaomiao, line 23: Added

jinghaomiao (changes requested):

modules/dreamview/main.cc
storypku: > Added

Done.

jinghaomiao (approved):

PR #12451, opened 2020-09-09
cyber_logger: fix a off-by-one problem

lx18233184051:


lx18233184051:
if the link size is sizeof(path), then it could be OOW

cyber/common/global_data.cc
```diff
@@ -41,7 +41,7 @@ namespace {
 const std::string& kEmptyString = "";
 std::string program_path() {
   char path[PATH_MAX];
-  auto len = readlink("/proc/self/exe", path, sizeof(path));
+  auto len = readlink("/proc/self/exe", path, sizeof(path) - 1);
```
storypku, line 44: Good catch.

storypku (approved):

storypku:
> if the link size is sizeof(path), then it could be OOW
BTW, what does oow mean ?

cyber/common/global_data.cc
xiaoxq: How about `boost::filesystem::canonical()`? But it will bind us to boost.

xiaoxq:
> > if the link size is sizeof(path), then it could be OOW
> 
> BTW, what does oow mean ?

OOM maybe? I think it's more like stack overflow.

PR #12461, opened 2020-09-10
Cyber: binary threading fixes

Fla3inH0tCheet0s:
There's some potential threading problems when using the header only binary library. I separated the header and the source file to hide the mutex header and added a mutex. The problems are reported by running a threading sanitizer.

### DIFF
--- a/cyber/binary.cc
+++ b/cyber/binary.cc
@@ -14,18 +14,28 @@
  * limitations under the License.
  *****************************************************************************/
 
-#ifndef CYBER_BINARY_H_
-#define CYBER_BINARY_H_
+#include "cyber/binary.h"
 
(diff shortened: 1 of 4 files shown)


cyber/binary.cc
```diff
@@ -0,0 +1,41 @@
...
+
+#include <mutex>
+#include <string>
+
+namespace apollo {
+namespace cyber {
+namespace binary {
+
+static std::mutex m;
+static std::string binary_name;  // NOLINT
```
storypku, line 27: Variables in anonymous namespace should be preferred over C style `static`.

storypku (changes requested):

storypku (approved):

PR #12496, opened 2020-09-11
planning: format

lx18233184051:


modules/planning/learning_based/model_inference/model_inference_test.cc
```diff
@@ -14,16 +14,19 @@
  * limitations under the License.
  *****************************************************************************/
 
-#include "cyber/common/file.h"
-#include "gflags/gflags.h"
 #include "gtest/gtest.h"
+
```
JasonZhou404, line 18: Why blank line here?

JasonZhou404 (changes requested):

JasonZhou404 (approved):

modules/planning/learning_based/model_inference/model_inference_test.cc
lx18233184051: I'm not sure. vscode format this automatically :)

JasonZhou404: I got it now, "gtest/gtest.h" is the header file and the other one is third-party file, so there should be a blank line in between

PR #12512, opened 2020-09-13
Bazel: simplified apollo_config.sh with support for both interactive and non-interactive mode

storypku:


storypku:
Finally, we have `./apollo.sh config --interactive` to build with native CUDA compute capability.

xiaoxq (approved):

PR #12520, opened 2020-09-14
Perception: add traffic light detection libtorch inference module

xiaoMrzhang:


jeroldchen:
`perception/inference` directory should only contain tool functions. Functions that have data flow logic need to be defined in `camera`, `lidar` or other directories that belong to some specific scenarios. Thus, `torch_det_net` and previous script `torch_net` should be divided into tool functions and work flow functions. Then put all tool functions into `torch_net` as well as rename those scripts containing work flow functions and move them into `camera/lib/traffic_light`.

modules/perception/inference/libtorch/torch_det_net.cc
```diff
@@ -0,0 +1,108 @@
...
+  return iter->second;
+}
+
+void TorchDetNet::Infer() {
+  torch::Device device(device_type_, device_id_);
+  auto blob = blobs_[input_names_[0]];
+  auto input_param = blobs_[input_names_[1]];
+
+  torch::Tensor tensor_image = torch::from_blob(
+                              blob->data()->mutable_cpu_data(),
```
jeroldchen: If you just need to read cpu data, you don't need `mutable_cpu_data` which is usually used to write data. There should be a function just named `cpu_data()`.

modules/perception/inference/libtorch/torch_det_net.cc
```diff
@@ -0,0 +1,108 @@
...
+                              {blob->shape(0), blob->shape(1), blob->shape(2),
+                              blob->shape(3)}, torch::kFloat32);
+  torch::Tensor tensor_param = torch::from_blob(
+                              input_param->data()->mutable_cpu_data(),
```
jeroldchen: Same here.

modules/perception/inference/libtorch/torch_det_net.cc
xiaoMrzhang: I tried cpu_data( ) function but torch::from_blob need void* as initializing argument.

modules/perception/inference/libtorch/torch_det.cc
```diff
@@ -0,0 +1,108 @@
...
+      blobs_.emplace(name, blob);
+    }
+  }
+  return true;
+}
+
+TorchDet::TorchDet(const std::string &net_file,
+                         const std::string &model_file,
+                         const std::vector<std::string> &outputs,
+                         const std::vector<std::string> &inputs)
```
changsh726, line 61: The format looks odd here.

xiaoMrzhang: I will fix them soon, thanks.

modules/perception/inference/libtorch/torch_det_net.cc
jeroldchen: Got it.

modules/perception/inference/libtorch/torch_det.cc
```diff
@@ -0,0 +1,108 @@
...
+    device_id_ = gpu_id_;
+  } else {
+    device_type_ = torch::kCPU;
+  }
+
+  // Init net
+  torch::Device device(device_type_, device_id_);
+  net_ = torch::jit::load(model_file_, device);
+
+  for (auto name : output_names_) {
```
lx18233184051, line 43: const auto& name : output_names_

jinghaomiao (approved):

PR #12532, opened 2020-09-14
V2X: rename and split targets

changsh726:


storypku (approved):

modules/v2x/fusion/libs/fusion/fusion_test.cc
```diff
@@ -20,6 +20,7 @@
 
 #include "gtest/gtest.h"
 
+#include "modules/v2x/fusion/libs/fusion/fusion.h"
```
lx18233184051, line 23: include twice. line 16 with line 23.

lx18233184051 (changes requested):

lx18233184051 (approved):

PR #12591, opened 2020-09-18
Calibration: add a new vehicle configuration files

cuishiwei:


modules/calibration/data/dev_kit_standard/gnss_params/ant_imu_leverarm.yaml
```diff
@@ -0,0 +1,21 @@
+#carid:mkz058
```
SeasoulChris, line 1: remove mkd comment

SeasoulChris (approved):

PR #12603, opened 2020-09-18
Dreamview: dependabot alerts fix

lj0919:
修复了四个警告 升级了四个yarn依赖  跑起来测试了一下dreamview没有问题


vlin17:
Don't forget to commit the latest bundle files and test with localhost:8888

modules/dreamview/frontend/package.json
```diff
@@ -18,15 +18,19 @@
         "lint-fix": "./node_modules/eslint/bin/eslint.js --fix src"
     },
     "dependencies": {
+        "bl": "^4.0.3",
```
vlin17, line 21: Don't add indirectly dependencies.

The Ideal way of upgrading packages is to find out which direct dependencies use those problematic ones and upgrade direct dependencies instead.

vlin17 (changes requested):

modules/dreamview/frontend/package.json
lj0919: 好的收到

PR #12622, opened 2020-09-21
Calibration: add dev kit advanced ne-b

zyc0222:


modules/calibration/data/dev_kit_advanced_ne-b/localization_conf/localization.conf
```diff
@@ -0,0 +1,213 @@
...
+
+# point cloud step
+# type: int
+# default: 2
+--point_cloud_step=2
+
+# The height from the center of lidar to ground plane
+# type: double
+# default: 1.80
+--lidar_height_default=1.80
```
SeasoulChris, line 40: the value of lidar_height_default is different to the value of velodyne16_height.yaml

SeasoulChris (approved):

PR #12648, opened 2020-09-23
Scripts: add spell check for input command of apollo.sh

changsh726:


scripts/command_checker.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python3
+
+###############################################################################
+# Copyright 2018 The Apollo Authors. All Rights Reserved.
+#
```
storypku, line 5: 2020

storypku (changes requested):

storypku (approved):

PR #12694, opened 2020-09-28
apollo.sh: create directory for APOLLO_BAZEL_DIST_DIR if non-exist

storypku:


changsh726 (dismissed):

apollo.sh
```diff
@@ -3,6 +3,7 @@ set -e
 
 TOP_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd -P)"
 source "${TOP_DIR}/scripts/apollo.bashrc"
+source "${TOP_DIR}/cyber/setup.bash"
```
changsh726, line 6: A question, I remember that this script has been sourced in `usr.bash_aliases`, why cannot CI find `${APOLLO_BAZEL_DIST_DIR}`

storypku: > A question, I remember that this script has been sourced in `usr.bash_aliases`, why cannot CI find `${APOLLO_BAZEL_DIST_DIR}`？

Good question. @lx18233184051 Do you have any ideas, Big Xu?

storypku: > > A question, I remember that this script has been sourced in `usr.bash_aliases`, why cannot CI find `${APOLLO_BAZEL_DIST_DIR}`？
> 
> Good question. @lx18233184051 Do you have any ideas, Big Xu?

I think the main reason is that CI runs apollo.sh directly without user.bash_aliases sourced.

changsh726 (approved):

PR #12709, opened 2020-09-30
cyber: unified preprocessing format

daohu527:
cyber: unified preprocessing format

### DIFF
--- a/cyber/class_loader/test/base.h
+++ b/cyber/class_loader/test/base.h
@@ -14,10 +14,13 @@
  * limitations under the License.
  *****************************************************************************/
 
-#pragma once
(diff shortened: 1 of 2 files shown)


storypku:
> cyber: unified preprocessing format

Hi, In Apollo, we prefer `#pragma once` guard, which is a bit different from the original Google C++ style guide. 

See [Drake's C++ Style Guide on `#pragma once` guard](https://drake.mit.edu/styleguide/cppguide.html#The__pragma_once_Guard)

daohu527:
There are two ways to deal with it. 
* cyber use #ifndef 
* apollo modules use #pragma

There is a lot of discussion about which way is better, I think the current style is unified, at least in the same module.

changsh726 (approved):

PR #12802, opened 2020-10-17
Monitor: Optimized GetReaderAndLatestMessage function

cflaviu:
Replaced channel identification based on serial string matching with hash map matching.
Ran format, lint and tests on monitor module.

### DIFF
--- a/modules/monitor/software/channel_monitor.cc
+++ b/modules/monitor/software/channel_monitor.cc
@@ -27,21 +27,22 @@
 #include "google/protobuf/descriptor.h"
 #include "google/protobuf/dynamic_message.h"
 
-#include "cyber/common/log.h"
-#include "cyber/cyber.h"
-#include "modules/common/adapters/adapter_gflags.h"
 #include "modules/common/latency_recorder/proto/latency_record.pb.h"
-#include "modules/common/util/map_util.h"
 #include "modules/control/proto/control_cmd.pb.h"
 #include "modules/drivers/proto/conti_radar.pb.h"
 #include "modules/drivers/proto/pointcloud.pb.h"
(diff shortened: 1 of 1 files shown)


modules/monitor/software/channel_monitor.cc
```diff
@@ -53,79 +54,54 @@ namespace apollo {
 namespace monitor {
 namespace {
 
+using ReaderAndMessagePair =
+    std::pair<std::shared_ptr<cyber::ReaderBase>,
+              std::shared_ptr<google::protobuf::Message>>;
+
+template <typename T>
+ReaderAndMessagePair CreateReaderAndLatestsMessage(const std::string& channel) {
+  const auto reader = MonitorManager::Instance()->CreateReader<T>(channel);
```
storypku, line 63: Type of reader should be `auto`  rather than `const auto`. 
See
https://github.com/ApolloAuto/apollo/blob/master/cyber/node/reader_base.h#L73

storypku (changes requested):
Thanks for your PR! Please fix the issue I mentioned before merging into master.

modules/monitor/software/channel_monitor.cc
cflaviu: The reader is a constant shared pointer object. Getting the internal pointer from shared pointer object is a "const" operation. Otherwise the compiler would have complained. The previous version of code use the same const auto reader.

storypku (approved):

modules/monitor/software/channel_monitor.cc
storypku: OK, I see. `const` was for shared_ptr rather than ReaderBase class instance.

PR #12861, opened 2020-10-26
map: refactored HDMapImpl

cflaviu:
map:
- refactored HDMapImpl class by creating a hierarchy of classes for lanes, junctions, signals, crosswalks, stop-signs, yield-signs, speed-bumps, overlaps, roads, PNC junctions and RSUs removing duplicate code;
- faster map initialization by preallocating memory for vectors, unordered_maps and unordered sets;
- faster lookup using unordered sets instead of ordered sets;
- uniform access to proto-buffer object in newly created classes by using inner_object() method following DRY principle;
- optimized serial if-else based look-up by using switch instruction;
- smaller refactorings removing duplicate code;

dreamview, planning, prediction, v2x:
- updated code according to map related changes;

- ran code format, code linting, map unit tests and dreamview simulation.


modules/map/tools/map_xysl.cc
```diff
@@ -91,6 +92,8 @@ class MapUtil {
   GET_ELEMENT_BY_ID(SpeedBump);
   GET_ELEMENT_BY_ID(StopSign);
   GET_ELEMENT_BY_ID(YieldSign);
+  GET_ELEMENT_BY_ID(PNCJunction);
+  GET_ELEMENT_BY_ID(RSU);
```
cflaviu, line 96: Added new functions for PNCJunction and RSU.

modules/map/tools/map_xysl.cc
```diff
@@ -140,51 +143,85 @@ class MapUtil {
...
+      case ObjectOverlapInfo::kPncJunctionOverlapInfo: {
+        std::cout << "PncJunction: " << raw_id << std::endl;
+        Print(GetPNCJunction(raw_id)->inner_object());
+        break;
+      }
+      // case ObjectOverlapInfo::kRsuOverlapInfo: {
+      //   std::cout << "RSU: " << raw_id << std::endl;
+      //   Print(GetRSU(raw_id)->inner_object());
+      //   break;
+      // }
```
cflaviu, line 202: Definition of GetRSU was not found leading to linking error. So currently is commented.

modules/map/hdmap/hdmap_objects.cc
```diff
@@ -0,0 +1,755 @@
+/* Copyright 2020 The Apollo Authors. All Rights Reserved.
```
cflaviu, line 1: The code of below methods/functions was taken from HDMapImpl implementation.

Issue #12840, opened 2020-10-23
A recent commit causes record IO error


xiaoxq:
https://github.com/ApolloAuto/apollo/commit/c9e58846902d3fbdefee1e1326abe4e45dcf479f
Reason is still unknown.

PR #12863, opened 2020-10-27
Cyber: cyber_recorder file writer threading bug fix

lfcarol:
bug fix [link](https://github.com/ApolloAuto/apollo/pull/12377)
reason:
  The reason is data race of two functions:  1) RecordFileWriter::Close() ; 2) RecordFileWriter::Flush()
  When the [record_viewer_test.cc:59](https://github.com/ApolloAuto/apollo/blob/de70315b459fd53f527e342eb884099406ce7481/cyber/record/record_viewer_test.cc#L63) call the 'Close()' function, and there are some message flushing to file at that moment(call the 'Flush()' function). The two functions run in two threads, so it can produce the two data race points : 1) RecordFileWriter::chunk_flush_;  2)RecordFileWriter::is_writing_ 

@Fla3inH0tCheet0s

Fixed issue #12840

### DIFF
--- a/cyber/record/file/record_file_base.h
+++ b/cyber/record/file/record_file_base.h
@@ -45,7 +45,7 @@ class RecordFileBase {
   std::string path_;
   proto::Header header_;
   proto::Index index_;
-  int fd_;
+  int fd_ = -1;
 };
 
 }  // namespace record
--- a/cyber/record/file/record_file_base.h
+++ b/cyber/record/file/record_file_base.h
@@ -45,7 +45,7 @@ class RecordFileBase {
   std::string path_;
   proto::Header header_;
   proto::Index index_;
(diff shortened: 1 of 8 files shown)


knzeus (approved):

storypku (approved):

cyber/record/file/record_file_base.h
```diff
@@ -45,7 +45,7 @@ class RecordFileBase {
   std::string path_;
   proto::Header header_;
   proto::Index index_;
-  int fd_ = -1;
+  int fd_;
```
storypku, line 48: Please keep the original as otherwise there is no default value for `fd`.

storypku (changes requested):

storypku (approved):

PR #12872, opened 2020-10-28
Build: BUILD files for sub-directories in cyber/transport

storypku:


cyber/transport/common/identity.cc
```diff
@@ -24,50 +24,47 @@ namespace apollo {
 namespace cyber {
 namespace transport {
 
-Identity::Identity(bool need_generate) : hash_value_(0), hash_value_str_("") {
```
storypku, line 27: No need to store `hash_value_str_`

changsh726 (approved):

PR #12909, opened 2020-11-04
Canbus: add the battery info in vehicle devkit and ch for chassis info

macDure:
add the new battery soc protocol in vehicle dev_kit
add the battery info in vehicle ch


macDure:
> add the new battery soc protocol in vehicle dev_kit
> add the battery info in vehicle ch

which is cherry-picked from master

PR #12915, opened 2020-11-04
Build: remove use of "#define private/protected public"

storypku:


modules/perception/fusion/common/kalman_filter_test.cc
```diff
@@ -13,10 +13,11 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  *****************************************************************************/
+#include "modules/perception/fusion/common/kalman_filter.h"
```
storypku, line 16: Done by `apollo.sh format`

jinghaomiao (approved):

PR #13008, opened 2020-11-16
Drivers: rearrange driver files and configurations

lfcarol:


storypku (approved):
Try to split the big PR into smaller ones.

PR #13020, opened 2020-11-17
dreamview:add the battery and gear ui

walle-zhao:


modules/dreamview/frontend/src/components/StatusBar/Electricity.js
```diff
@@ -0,0 +1,35 @@
...
+    const {
+      electricityPercentage,
+    } = this.props;
+
+    const percentageString = `${electricityPercentage}%`;
+    const rectWidth = 24 * (electricityPercentage / 100);
+    return (
+    <div className="show-electricity">
+        <div className="battery-icon">
+	    <svg version="1.1">
```
vlin17, line 17: indentation is off. Same for all other js files. Please use formatter to fix all lint issues.

vlin17 (commented):
please also provide a screenshot that includes the changes. Easier to review by knowing what the GUI looks like. thanks

modules/dreamview/frontend/src/components/StatusBar/Electricity.js
```diff
@@ -0,0 +1,35 @@
...
+              <rect rx="2" ry="2" x="50" y="10" width="30" height="15" fill="rgb(15,127,18)">
```
vlin17, line 18: i think you don't need the close tag "\</rect>". Just add "/>" to the end of the open tag.

modules/dreamview/frontend/src/components/StatusBar/Gears.js
```diff
@@ -0,0 +1,26 @@
...
+
+@observer
+export default class Gears extends React.Component {
+
+  render() {
+    const {
+      currentGear,
+    } = this.props;
+
+    const gearAlphabet = currentGear.charAt(5);
```
vlin17, line 12: Have we handle cases if Dreamview doesn't receive batterySocPercentage and currentGear values?

walle-zhao:
> please also provide a screenshot that includes the changes. Easier to review by knowing what the GUI looks like. thanks

Normal case
581605770954_ pic_hd

This is the case when it doesn't receive batterySocPercentage and currentGear values.
image

modules/dreamview/frontend/src/components/StatusBar/Electricity.js
walle-zhao: got it, thank u.

modules/dreamview/frontend/src/components/StatusBar/Electricity.js
walle-zhao: got it, thank u.

modules/dreamview/frontend/src/components/StatusBar/Gears.js
walle-zhao: It has been modified.Thank u.

modules/dreamview/frontend/src/components/StatusBar/index.js
```diff
@@ -43,6 +45,12 @@ export default class StatusBar extends React.Component {
                         isAutoMode={meters.isAutoMode}
                     />
                 </div>
+                <Electricity
+                    electricityPercentage={meters.batterySocPercentage}
+                />
+                <Gears
+                    currentGear={meters.gearLocaltion}
```
vlin17, line 52: typo: gearLocation

friendly note: VS code has spelling check plugin

modules/dreamview/frontend/src/store/meters.js
```diff
@@ -48,6 +48,8 @@ export default class Meters {
 
     @observable batterySocPercentage = 0;
```
vlin17, line 49: what does "Soc" mean?

vlin17:
> > please also provide a screenshot that includes the changes. Easier to review by knowing what the GUI looks like. thanks
> 
> Normal case
> 
> WechatIMG55
> 
> This is the case when it doesn't receive batterySocPercentage and currentGear values.
> image

1. On Dreamview, each grid indicates one status information. It's fine to have widths of Gear and Battery to be twice as traffic light's, but for consistency, i think it would be better not splitting its label and value. 

2. Battery isn't a status that can be used for all kinds of cars. It might be better to hide it if a car doesn't support it. One possible approach is to initialize batterySocPercentage = null in store/meters.js and render it only if a number is assigned.

modules/dreamview/frontend/src/components/StatusBar/Electricity.js
```diff
@@ -0,0 +1,31 @@
...
+    const rectWidth = 24 * (electricityPercentage / 100);
+    return (
+            <div className="show-electricity">
+                <div className="battery-icon">
+                    <svg version="1.1">
+                        <rect rx="2" ry="2" x="44" y="10" width="30" height="15" fill="rgb(15,127,18)"/>
+                        <rect x="46" y="12" width="26" height="11" fill="rgb(0,0,0,0.5)"/>
+                        <rect x="47" y="13" width={ rectWidth } height="9" fill="rgb(15,127,18)"/>
+                        <rect rx="2" ry="2" x="74" y="13" height="8" width="2" fill="rgb(15,127,18)"/>
+                    </svg>
```
vlin17, line 22: as we are using svg to show battery label, do we want to change it to red when percentage is low?

modules/dreamview/frontend/src/styles/main.scss
```diff
@@ -661,6 +661,94 @@ body {
         }
     }
 
+    .show-electricity{
```
vlin17, line 664: 1. extra space between the name and "{".
2. please choose a noun to name the class (e.g. electricity-status)

spot format issues elsewhere as well, please also use formatter to fix css format.

modules/dreamview/frontend/src/styles/main.scss
```diff
@@ -661,6 +661,94 @@ body {
...
+
+        .electricity-percentage {
+            position: absolute;
+            top: 0px;
+            right: 0px;
+            width: 105px;
+            height: 35px;
+            background: rgba(0, 0, 0, 0.5);
+
+            .text {
```
vlin17, line 688: This .text style is repeated in .gears-label.label and current-gears.text. Please try reuse the same code.  One possible approach is to use SCSS's extend feature, or simply define a common one for React component to refer to.

modules/dreamview/frontend/src/styles/main.scss
```diff
@@ -661,6 +661,68 @@ body {
...
+        left: 50%;
+
+        float: right;
+        transform: translate(-50%, -50%);
+        text-align: center;
+        font-size: 14px;
+        color: #FFFFFF;
+     }
+
+    .electricity-status {
```
vlin17, line 676: 1. now the electricity-status class and gear class are almost identical, except the top. Can we try consolidating the two styles? Would be great to reduce repeated code and follow reusability guideline

2. might want to move gear into the same row as battery's.

modules/dreamview/frontend/src/components/StatusBar/index.js
```diff
@@ -16,6 +18,11 @@ export default class StatusBar extends React.Component {
       showPlanningRSSInfo, monitor,
     } = this.props;
 
+    let electricityPercentage = meters.batteryPercentage;
```
vlin17, line 21: the batteryPercentage is set to null as default. Is there any reason why we need to do this change here? Also, if this really is needed, maybe set it in meters.js as it is the place for data manipulation.

vlin17 (approved):

PR #13040, opened 2020-11-19
task_manager: add a new module for task management

JackFu123:


modules/task_manager/proto/.pb.h
```diff
@@ -0,0 +1,80 @@
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
+// source: modules/task_manager/proto/
```
storypku, line 2: Please remove the ".pb.h" and ".pb.cc" files first.

And, I personally don't think it's appropriate to create a top-level module here.
Correct me if I am wrong.

storypku (changes requested):

modules/task_manager/proto/.pb.h
JackFu123: 1).pb.h & .pb.cc removed.
2)This module is for task management and relatively independent from the functional modules of autonomous driving. So, it should be on the same level as other modules. Otherwise, it would be inconvenient to extend if there are other tasks(such as different routing tasks when demenstrating by D-kit) in the future.

guoweiwan (approved):

PR #13076, opened 2020-11-29
Use the nvidia-docker2 package for the installation/setting of the NVIDIA container toolkit.

hongyi-zhao:
The current command for the installation of NVIDIA container toolkit has been changed into the following:

`
$ sudo apt-get install -y nvidia-docker2`

See [the official document](https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/install-guide.html#setting-up-nvidia-container-toolkit) and this [discussion](https://github.com/NVIDIA/nvidia-docker/issues/1407) filed on nvidia-docker issues list for more info.

### DIFF
--- a/docs/specs/prerequisite_software_installation_guide.md
+++ b/docs/specs/prerequisite_software_installation_guide.md
@@ -122,7 +122,7 @@ distribution=$(. /etc/os-release;echo $ID$VERSION_ID)
 curl -s -L https://nvidia.github.io/nvidia-docker/gpgkey | sudo apt-key add -
 curl -s -L https://nvidia.github.io/nvidia-docker/$distribution/nvidia-docker.list | sudo tee /etc/apt/sources.list.d/nvidia-docker.list
 sudo apt-get -y update
-sudo apt-get install -y nvidia-container-toolkit
+sudo apt-get install -y nvidia-docker2
 ```
 
 ![tip_icon](images/tip_icon.png) Don't forget to restart the Docker daemon for


ntutangyun:
When using docker 19.03+ there is no need to install Nvidia-rocker any more. Docker + Nvidia-container-toolkit is sufficient

hongyi-zhao:
> When using docker 19.03+ there is no need to install Nvidia-rocker any more. Docker + Nvidia-container-toolkit is sufficient

If we manage the docker with k8s, then nvidia-docker2 is needed. Anyway, the installation of nvidia-docker2 will cause nvidia-container-toolkit to be installed automatically.

storypku:
> > When using docker 19.03+ there is no need to install Nvidia-rocker any more. Docker + Nvidia-container-toolkit is sufficient
> 
> If we manage the docker containers in a cluster environment with k8s, where there is no way to pass **--gpus** down to docker from within Kubernetes, then nvidia-docker2 is needed. Anyway, the installation of nvidia-docker2 will cause nvidia-container-toolkit to be installed automatically.

As Docker 19.03 + nvidia-container-toolkit is sufficient to run CUDA-enabled Docker containers in most cases, it seems no change is needed here. For k8s environment, maybe a *Note* section would be appropriate.

xiaoxq (approved):

xiaoxq:
> As Docker 19.03 + nvidia-container-toolkit is sufficient to run CUDA-enabled Docker containers in most cases, it seems no change is needed here. For k8s environment, maybe a _Note_ section would be appropriate.

The upstream does change their installation guide. Let's just align with it.

PR #13081, opened 2020-11-30
task_manager: modify logic for cycle counting

JackFu123:


modules/task_manager/cycle_routing_manager.cc
```diff
@@ -45,18 +45,26 @@ common::Status CycleRoutingManager::Init(
...
+    << ", end point " << end_point_.x() << " " << end_point_.y()
+    << ", threshold " << FLAGS_threshold_for_destination_check
+    << ", allowed_to_send_routing_request " << is_allowed_to_route_;
+
+    if (!CheckPointDistanceInThreshold(
           begin_point_,
-          FLAGS_threshold_for_destination_check)) {
-        last_point_ = pose.position();
+          pose.position(),
+          FLAGS_threshold_for_destination_check * 2)) {
```
JasonZhou404, line 58: Why double it here?

modules/task_manager/cycle_routing_manager.h
```diff
@@ -64,9 +64,9 @@ class CycleRoutingManager {
  private:
   int cycle_;
   int waypoint_num_;
+  bool is_allowed_to_route_;
```
JasonZhou404, line 67: Just In case you forget to to use init() function, please manually initialize the private data field to default values here as you use a default constructor

JasonZhou404 (changes requested):

modules/task_manager/cycle_routing_manager.cc
JackFu123: In case of localization pose drift, routing is only allowed when the pose is 2*Threshold away from the beginning position. I will change 2 to a FLAG.

modules/task_manager/cycle_routing_manager.h
JackFu123: OK

JasonZhou404 (approved):

JasonZhou404 (approved):

PR #13132, opened 2020-12-04
D-Kit : Open the building of the visualizer

lishuanghua:
D-Kit : Open the building of the visualizer


storypku:
So you have checked that cyber_visualizer was able to run on the `r5.5.0` branch @lishuanghua ?

lishuanghua:
> So you have checked that cyber_visualizer was able to run on the `r5.5.0` branch @lishuanghua ?

yeah,and you can check it again.

PR #13144, opened 2020-12-08
Docker and Build: independent BUILD file for mainboard and bump Bazel to 3.7.1

storypku:


storypku:
It seems that this only works with higher Bazel versions. e.g. Bazel 3.7.1

changsh726 (approved):

PR #13161, opened 2020-12-10
Bazel&Build: use @platforms//cpu:xxx to refer to cpu architecture

storypku:


storypku:
It seems this requires a minumu working Bazel version > 3.5.0, exact version not checked yet.

changsh726 (approved):

PR #13172, opened 2020-12-11
canbus: add aeb control command in devkit

macDure:


modules/calibration/data/dev_kit_standard/cancard_params/canbus.conf
```diff
@@ -2,3 +2,4 @@
 --canbus_conf_file=/apollo/modules/canbus/conf/canbus_conf.pb.txt
 --enable_chassis_detail_pub
 --noreceive_guardian
+--enable_devkit_standard_aeb
```
diegohu, line 5: change to enable_aeb

diegohu (changes requested):

modules/calibration/data/dev_kit_standard/cancard_params/canbus.conf
macDure: has fixed

diegohu (approved):

PR #13193, opened 2020-12-15
Scripts: bugfix for rtk_recorder.sh

storypku:


modules/tools/record_play/rtk_recorder.py
```diff
@@ -30,6 +30,7 @@
 from gflags import FLAGS
 
 from modules.tools.common.logger import Logger
+import modules.tools.common.proto_utils as proto_utils
```
storypku, line 33: Otherwise  rtk_recorder.py#L78 will fail.

lishuanghua (approved):

PR #13256, opened 2020-12-28
cyber_py3: bugfix for cyber_py3 service & client examples

storypku:


cyber/python/cyber_py3/cyber.py
```diff
@@ -290,7 +290,9 @@ def create_client(self, name, request_data_type, response_data_type):
         return Client(c, response_data_type)
 
     def service_callback(self, name):
-        v = self.services[name]
+        # Temporary workaround for cyber_py3 examples: service & client
+        v = self.services[name.decode("utf-8")]
```
storypku, line 294: It's really weird that argument `name` was `b'server_01'` in cyber_py3/examples/{service,client}.py

@xiaoxq Do you have any ideas?

xiaoxq (approved):

cyber/python/cyber_py3/cyber.py
xiaoxq: It's common in py3 that many APIs use bytes instead of string. I don't like it also :(

PR #13257, opened 2020-12-29
Shared library: remove poco dependency authored by daohu527

storypku:
With original notes/license from the Poco project.
And test cases for shared_library specifically.

@daohu527

PR #13261, opened 2020-12-30
Build: add missing boost deps for several BUILD files

storypku:


storypku:
The missing dependency of Boost appears after removing Poco dependency in PR #13257 , which is really weird.
Since CyberRT doesn't depend on Boost...

zhoujie3 (approved):

PR #13263, opened 2020-12-31
fix a bug in postprocess_cuda.cu that affects nms of pointpillars

zhanggefan:
fix a bug that affects nms


zhanggefan:
Looks like the code (and the bug as well) is migrated from Autoware. ;-)

jeroldchen (approved):
Thank you for your contribution!

PR #13271, opened 2021-01-04
Optimized IsNearJunction function.

johnnyleeRH:
use is_near_junction feature attribute to reduce computing.

### DIFF
--- a/modules/prediction/container/obstacles/obstacle.cc
+++ b/modules/prediction/container/obstacles/obstacle.cc
@@ -130,6 +130,9 @@ bool Obstacle::IsNearJunction() {
   if (feature_history_.empty()) {
     return false;
   }
+  if (latest_feature().has_is_near_junction()) {
+    return latest_feature().is_near_junction();
+  }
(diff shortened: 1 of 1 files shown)


modules/prediction/container/obstacles/obstacle.cc
```diff
@@ -130,6 +130,9 @@ bool Obstacle::IsNearJunction() {
   if (feature_history_.empty()) {
     return false;
   }
+  if (latest_feature().has_is_near_junction()) {
+    return latest_feature().is_near_junction();
+  }
   double pos_x = latest_feature().position().x();
   double pos_y = latest_feature().position().y();
```
cflaviu, line 137: I would replace duplicate function calls with const references to "latest_feature" and "position". Locals pos_x and pos_y are not necessary. IsNearJunction can be marked as const method.

johnnyleeRH: ok,  overall handing these duplicated function is better, I just close this pull request now

PR #13274, opened 2021-01-04
Dreamview:Loop routing optimization on master

lj0919:
1. add no loop routing situation
2. add threshold at dv be


modules/dreamview/frontend/src/store/route_editing_manager.js
```diff
@@ -20,7 +22,7 @@ export default class RouteEditingManager {
 
     currentDefaultRouting = 'none';
 
-    @action updateDefaultRoutingEndPoint(data) {
+  @action updateDefaultRoutingEndPoint(data) {
```
changsh726, line 25: At least make it aligned to `@observable...`

lj0919: thx

changsh726 (approved):

PR #13305, opened 2021-01-12
Docker: added smoke model volume

changsh726:


.teamcity/run_ci.sh
```diff
@@ -229,6 +229,12 @@ function mount_other_volumes() {
     reuse_or_start_volume "${faster_rcnn_volume}" "${faster_rcnn_image}"
     volume_conf="${volume_conf} --volumes-from ${faster_rcnn_volume}"
 
+    # SMOKE
+    local smoke_volume="apollo_smoke_volume_${USER}"
+    local smoke_image="${DOCKER_REPO}:smoke_volume-yolo_obstacle_detection_model-${TARGET_ARCH}-latest"
+    reuse_or_start_volume "${smoke_volume}" "${smoke_image}"
+    volume_conf="${volume_conf} --volumes-from ${smoke_volume}"
+
```
storypku, line 237: Since only x86_64 volume was ready, we'd better add code like
```
if [[ "$(uname -m)" == "x86_64" ]]; then
    your logic here
fi

storypku (changes requested):

.teamcity/run_ci.sh
changsh726: May you help build an arm volume?

storypku (approved):

PR #13409, opened 2021-02-04
Cyber: bugfix for dlopen operations in SharedLibrary impl

storypku:
@daohu527 
@JackFu123 

Without this PR, running `mainboard -d modules/planning/dag/planning.dag` will complains:
    
```text
    INTEL MKL ERROR: /usr/local/lib/libmkl_avx2.so: undefined symbol: mkl_sparse_optimize_bsr_trsm_i8.
    Intel MKL FATAL ERROR: Cannot load libmkl_avx2.so or libmkl_def.so.
```

### DIFF
--- a/cyber/class_loader/shared_library/shared_library.cc
+++ b/cyber/class_loader/shared_library/shared_library.cc
@@ -29,6 +29,8 @@ namespace apollo {
 namespace cyber {
 namespace class_loader {
 
+std::mutex SharedLibrary::mutex_;
+
 SharedLibrary::SharedLibrary(const std::string& path) { Load(path, 0); }
 
 SharedLibrary::SharedLibrary(const std::string& path, int flags) {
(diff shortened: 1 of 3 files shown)


cyber/class_loader/shared_library/shared_library.cc
```diff
@@ -43,9 +45,9 @@ void SharedLibrary::Load(const std::string& path, int flags) {
 
   int real_flag = RTLD_LAZY;
   if (flags & SHLIB_LOCAL) {
-    real_flag |= SHLIB_LOCAL;
```
daohu527, line 46: Seems like this is the problem. I check the source code and found
```
/* If the following bit is set in the MODE argument to `dlopen',
the symbols of the loaded object and its dependencies are made
visible as if the object were linked directly into the program.  */
#define RTLD_GLOBAL        0x00100
 
/* Unix98 demands the following flag which is the inverse to RTLD_GLOBAL.
The implementation does this by default and so we can define the
value to zero.  */
#define RTLD_LOCAL        0
```
The poco use `RTLD_LOCAL` and `RTLD_GLOBAL`, I may have made a mistake before!

daohu527 (approved):

JackFu123 (approved):
It works for me in sim_control.

PR #13434, opened 2021-02-13
Docker: Update deviceQuery install script

gsa88:
I encountered a trouble (download time out) while trying to install the prebuillt deviceQuery binary and made the alternative install script.

### DIFF
--- a/docker/build/installers/install_deviceQuery.sh
+++ b/docker/build/installers/install_deviceQuery.sh
@@ -24,6 +24,15 @@ cd "$(dirname "${BASH_SOURCE[0]}")"
 
 TARGET_ARCH="$(uname -m)"
 
+if [[ -z "${CUDA_VERSION}" ]]
+then
+    if [[ "${TARGET_ARCH}" == "x86_64" ]]; then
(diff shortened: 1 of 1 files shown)


storypku:
> I encountered a trouble (download time out) while trying to install the prebuillt deviceQuery binary and made the alternative install script.

Hi @gsa88 , there is a Docker build-arg `INSTALL_MODE` to decide whether to use prebuilt package(s) or to build and install package(s) from source.  This applies to `deviceQuery` installer as well.

Besides, we prefer official package releases to the ones cloned from VCS, since the former is more stable and well-tested.  Moreover, git clone operations to GitHub from China (where the maintenance team members live) is relatively slow, so the use of direct GIT operations is discouraged. Sorry for the inconvenience.

We will look into the download timeout issue of the deviceQuery binary program. Please stay tuned.

storypku (changes requested):
Please adjust the deviceQuery installer be conform to the following struct:
```
if [[ "${INSTALL_MODE}" == "download" ]]; then
     # steps to download
else
    # steps to build from source
fi
```

storypku (approved):

PR #13436, opened 2021-02-13
Docker: Fix typo into build_docker.sh

gsa88:
There is no `-b `argument, but the example specifies it:
```bash
...
echo "${TAB}${prog} -f dev.aarch64.dockerfile -m download -b testing"
...
```
Must be fixed to` -d`


storypku (approved):

storypku:
Thanks for pointing it out !

PR #13458, opened 2021-02-22
socket can: CAN_RAW_FILTER_MAX is 512 from linux kernel>=4.4.39

panxuechao:


modules/drivers/canbus/can_client/socket/socket_can_client_raw.cc
```diff
@@ -79,11 +81,9 @@ ErrorCode SocketCanClientRaw::Start() {
...
   // 1. for non virtual busses, set receive message_id filter, ie white list
   if (interface_ != CANCardParameter::VIRTUAL) {
-    struct can_filter filter[2048];
-    for (int i = 0; i < 2048; ++i) {
-      filter[i].can_id = 0x000 + i;
-      filter[i].can_mask = CAN_SFF_MASK;
-    }
+    struct can_filter filter[1];
+    filter[0].can_id = 0x000;
+    filter[0].can_mask = CAN_EFF_MASK & CAN_SID_FILTER;
```
storypku, line 86: Please describe the reasons for the code change. Thank you.

panxuechao: Set a scope for each Eid instead of a single filter rule for each Eid

macDure: > Please describe the reasons for the code change. Thank you.

2 resons:
From the linux kernel 4.4.39 and above, 1) in linux CAN_RAW_FILTER_MAX has been limited below 512(0x100), this change cause set CAN ID filter return -1; 2) And this change cause our CAN ID filter only filter 0x000~0x100, but we need  CAN ID filter for 0x000 ~ 0x7FF, so we need to set CAN ID filter for new method to solve this problem.

storypku (approved):

PR #13483, opened 2021-02-26
CI: add coverage python interface

lx18233184051:
Note: python need pygithub.


docker/build/installers/py3_requirements.txt
```diff
@@ -42,3 +42,5 @@ h5py
 pyyaml
 utm
 
+#Github
+pygithub
```
storypku, line 46: Since pygithub is only used for CI, we'd better not include it in Docker images.

storypku (dismissed):

docker/build/installers/py3_requirements.txt
lx18233184051: If not, CI will install it every time when container restart, which is very inefficient.

JackFu123 (approved):

PR #13539, opened 2021-03-10
Add check for SocketCan interface presence when setup device for aarch64

gsa88:


scripts/apollo_base.sh
```diff
@@ -46,13 +46,21 @@ function determine_bin_prefix() {
...
     warning "No CAN device named ${can_dev}. "
-    return
   fi
 
-  sudo ip link set can0 type can bitrate 500000
-  sudo ip link set can0 up
+  if [[ -x "$(command -v ip)" ]]; then
+    if ! ip link show type can | grep "${socket_can_dev}" &>/dev/null; then
+      warning "No SocketCAN device named ${socket_can_dev}."
+    else
```
storypku, line 57: Hi @gsa88 , can you please describe the purpose for checking can0 device existence using `ip link show type can` rather than check the existence of `/dev/can0` directly ?

gsa88: Hi @storypku! 
In Linux, there are different types of CAN bus communication using device nodes `/dev/can*` as a character device or SocketCAN `can*` as network device. The logic of the code is now as follows: 
- If there is no character CAN device `/dev/can0` check for the existence of SocketCAN network device `can0` .
- If the SocketCAN `can0` device exists set it up

Without checking the existence of  SocketCAN network device `can0` the commands `sudo ip link set can0 type can bitrate 500000` and `sudo ip link set can0 up` throws the error `Cannot find device "can0"`

storypku (approved):

PR #13541, opened 2021-03-11
I think here should be an add instead of minus

lycismail:
I think here should use add. Because if we suppose that the vehicle should driving in clockwise，means the ref traj like a circle，and the real point of our car is outside the circle .Then if we use cross product of the (cos_ref_theta, sin_ref_theta) and (dx, dy) , the result is a positive num. But the fuction here want it to be a negative one, because that make the one_minus_kappa_r_d = R_real / R_ref， the R_real is bigger than R_ref

### DIFF
--- a/modules/control/common/trajectory_analyzer.cc
+++ b/modules/control/common/trajectory_analyzer.cc
@@ -130,7 +130,7 @@ void TrajectoryAnalyzer::ToTrajectoryFrame(const double x, const double y,
 
   *ptr_d_dot = v * sin_delta_theta;
 
-  double one_minus_kappa_r_d = 1 - ref_point.kappa() * (*ptr_d);
+  double one_minus_kappa_r_d = 1 + ref_point.kappa() * (*ptr_d);
   if (one_minus_kappa_r_d <= 0.0) {
     AERROR << "TrajectoryAnalyzer::ToTrajectoryFrame "
               "found fatal reference and actual difference. "


HongyiSun:
It's quite hard to denote the math formula here.
Please refer to the APPENDIX I of Werling's Optimal Trajectory Generation for Dynamic Street Scenarios in a Frenet Frame paper here[https://www.researchgate.net/publication/224156269_Optimal_Trajectory_Generation_for_Dynamic_Street_Scenarios_in_a_Frenet_Frame]

HongyiSun (changes requested):
It's quite hard to denote the math formula here.
Please refer to the APPENDIX I of Werling's Optimal Trajectory Generation for Dynamic Street Scenarios in a Frenet Frame paper [here](https://www.researchgate.net/publication/224156269_Optimal_Trajectory_Generation_for_Dynamic_Street_Scenarios_in_a_Frenet_Frame)

lycismail:
Thx a lot. But i dont know why should keep 1 - Kr*d > 0 , whitch means Rr - d > 0. Is it a stability criteria ？

lycismail:
> It's quite hard to denote the math formula here.
> Please refer to the APPENDIX I of Werling's _Optimal Trajectory Generation for Dynamic Street Scenarios in a Frenet Frame_ paper [here](https://www.researchgate.net/publication/224156269_Optimal_Trajectory_Generation_for_Dynamic_Street_Scenarios_in_a_Frenet_Frame)

Thx a lot. But i dont know why should keep 1 - Kr*d > 0 , whitch means Rr - d > 0. Is it a stability criteria ？

HongyiSun:
> > It's quite hard to denote the math formula here.
> > Please refer to the APPENDIX I of Werling's _Optimal Trajectory Generation for Dynamic Street Scenarios in a Frenet Frame_ paper [here](https://www.researchgate.net/publication/224156269_Optimal_Trajectory_Generation_for_Dynamic_Street_Scenarios_in_a_Frenet_Frame)
> 
> Thx a lot. But i dont know why should keep 1 - Kr*d > 0 , whitch means Rr - d > 0. Is it a stability criteria ？

Yes, It's kind of a stability criteria.
As for we transform our path between Cartesian Space and Frenet Frame, there has to be a criteria that our points should stay close to the reference line, and the angle difference should remain small, or the transformation is not valid any more.

lycismail:
> > > It's quite hard to denote the math formula here.
> > > Please refer to the APPENDIX I of Werling's _Optimal Trajectory Generation for Dynamic Street Scenarios in a Frenet Frame_ paper [here](https://www.researchgate.net/publication/224156269_Optimal_Trajectory_Generation_for_Dynamic_Street_Scenarios_in_a_Frenet_Frame)
> > 
> > 
> > Thx a lot. But i dont know why should keep 1 - Kr*d > 0 , whitch means Rr - d > 0. Is it a stability criteria ？
> 
> Yes, It's kind of a stability criteria.
> As for we transform our path between Cartesian Space and Frenet Frame, there has to be a criteria that our points should stay close to the reference line, and the angle difference should remain small, or the transformation is not valid any more.

Thank u so much. I will think it carefully. Would you mind I ask another question here? I wonder why use osqp method to solve the MPC matrix without build QP problem like before which used QPOASES and made an H & f with reference to matrix A and B. Now we just use matrix Q & P to solve. I can not understand the QP progress. Could u give me a reference, like a paper or something? Very appreciate！

PR #13549, opened 2021-03-12
Docker: updated dev.x86_64.stable image w/ adv_plat rebuilt

storypku:


changsh726 (approved):

storypku:
This PR aims to fix the issue that wecam can't boot. We tracked down the problem to the compilation of adv_plat.

Need to figure out the core reason why build output of adv_plat on ubuntu 20.04 didn't work.

PR #13601, opened 2021-03-24
fix low risk code

tthhee:
fix low-level risk code.


storypku (approved):

storypku (changes requested):
Please change the commit msg to something like `Localization: limit level to pass security check`

PR #13611, opened 2021-03-25
Scripts: re-enable buggy core pattern

storypku:
The tricky part with setting core dump path under `/apollo` is as follows:

Changing core pattern inside Docker will change also that on the host. However, there may not exist a top level directory named `/apollo`.

Of course we can change core pattern to something like `/tmp/core/core_%e.%p`, which will be lost when Apollo container restarts.

What's your opinion? @zhoujie5517 @xiaoxq

### DIFF
--- a/scripts/docker_start_user.sh
+++ b/scripts/docker_start_user.sh
@@ -87,6 +87,14 @@ function setup_apollo_directories() {
   # chown -R "${uid}:${gid}" "${apollo_dir}"
 }
 
+# FIXME(infra): This will change core pattern on the host also,
+# where the `/apollo` directory may not exist.
+function setup_core_pattern() {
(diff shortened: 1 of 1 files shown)


zhoujie5517 (approved):

xiaoxq:
In the past we do create a `/apollo` link on host, pointing to the apollo dir. This works well for single-user and single-apollo-dir.

storypku:
> In the past we did create a `/apollo` link on host, pointing to the active apollo dir. This works well for single-user and single-apollo-dir.

Yeap. However, creating `/apollo` is a bit tricky and I wonder if we can move coredump directory to `/tmp/` directory. Or just don't set core pattern and leave it to the user.

PR #13631, opened 2021-03-29
Add Bazel install support to image_decomress module and unify bootstrap scripts

shr-project:


BUILD
```diff
@@ -25,6 +25,7 @@ install(
         "//modules/drivers/camera:install",
         "//modules/drivers/gnss:install",
         "//modules/drivers/microphone:install",
+        "//modules/drivers/tools/image_decompress:install",
         "//modules/guardian:install",
```
storypku, line 29: Better add a `//modules/drivers/tools:install` target here which depends on  `image_decompress:install` as the top level BUILD file should deal with too much details.

shr-project: Then shoudln't there be another target for all drivers? There is only image_decompress in //modules/drivers/tools and there is no BUILD file now, while //modules/drivers has one which can encapsulate all 4 currently listed here.

storypku (approved):

BUILD
storypku: > Then shoudln't there be another target for all drivers? There is only image_decompress in //modules/drivers/tools and there is no BUILD file now, while //modules/drivers has one which can encapsulate all 4 currently listed here.

Thanks for your improvement. Approved.

PR #13622, opened 2021-03-28
Scripts: re-enable buggy core pattern

SeasoulChris:
re-enable buggy core pattern

PR #13632, opened 2021-03-29
scripts: write core_pattern only when it's writable

shr-project:
Fix https://github.com/ApolloAuto/apollo/pull/13622

Without this, it seems to fail on host
```
 ---> Running in 427e7861043d
Adding group `apollo' (GID 1001) ...
Done.
Adding user `apollo' ...
Adding new user `apollo' (1001) with group `apollo' ...
Creating home directory `/home/apollo' ...
Copying files from `/etc/skel' ...
/apollo/scripts/docker_start_user.sh: line 96: /proc/sys/kernel/core_pattern: Read-only file system
The command '/bin/sh -c /apollo/scripts/docker_start_user.sh' returned a non-zero code: 1
```

### DIFF
--- a/scripts/docker_start_user.sh
+++ b/scripts/docker_start_user.sh
@@ -90,7 +90,7 @@ function setup_apollo_directories() {
 # FIXME(infra): This will change core pattern on the host also,
 # where the `/apollo` directory may not exist.
 function setup_core_pattern() {
-  if [ -e /proc/sys/kernel ]; then
+  if [[ -w /proc/sys/kernel/core_pattern ]]; then
     echo "/apollo/data/core/core_%e.%p" > /proc/sys/kernel/core_pattern
   fi
 }


scripts/docker_start_user.sh
```diff
@@ -90,7 +90,7 @@ function setup_apollo_directories() {
 # FIXME(infra): This will change core pattern on the host also,
 # where the `/apollo` directory may not exist.
 function setup_core_pattern() {
-  if [ -e /proc/sys/kernel ]; then
+  if [ -e /proc/sys/kernel -a -w /proc/sys/kernel/core_pattern ]; then
     echo "/apollo/data/core/core_%e.%p" > /proc/sys/kernel/core_pattern
   fi
 }
```
storypku, line 96: Shellcheck suggests:

```
In hello.sh line 3:
if [ -e /proc/sys/kernel -a -w /proc/sys/kernel/core_pattern ]; then
                         ^-- SC2166: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
```

shr-project: Interesting, I'm using -a all the time inside single braces as it's documented in test https://linux.die.net/man/1/test.

Will update it to:
```
if [[ -w /proc/sys/kernel/core_pattern ]]; then
```

to make it even simpler.

storypku (approved):

PR #13648, opened 2021-03-31
docs: modified format

spicyog:


JackFu123:
When a new PR is opened, the title should be "module_name: main purpose". So, "modified format" should be changed this way.

docs/technical_documents/open_space_decider.md
```diff
@@ -0,0 +1,175 @@
+# OPEN SPACE DECIDER
+
+# Introduction
+
+Apollo planning is a scenario based planning method, each driving use case is treated as a different driving scenario.
```
JackFu123, line 5: Maybe: Apollo planning is scenario-based, where each driving case is treated as a different driving scenario.

docs/technical_documents/open_space_decider.md
```diff
@@ -0,0 +1,175 @@
...
+
+open space decider is used to process related infomation and provide information for subsequent optimizers
```
JackFu123, line 7: open->Open, add '.' in the end.

docs/technical_documents/open_space_decider.md
```diff
@@ -0,0 +1,175 @@
...
+    5. Get parking boundary: convert parking spot points coordinates to vehicle coorinates and get points' projections on reference line;
+    
+    ```cpp
+        bool GetParkingBoundary(Frame *const frame,
+                                const std::array<common::math::Vec2d, 4> &vertices,
+                                const hdmap::Path &nearby_path,
+                                std::vector<std::vector<common::math::Vec2d>>
+                                *const roi_parking_boundary);
+    ```
+    6. Get road boundary: it starts from parking space center line(center_line_s) and extends a distance(roi longitudinal range) to both side along the s direction to get start_s and end_s; search key points(the point on the left/right lane boundary is close to a curb corner) and anchor points(a start/end point or the point on path with large curvatures) in this roi range; those key points, anchor points is called boundary points and the line segements between those points is called roi parking boundary;
```
JackFu123, line 54: both side->maybe "both sides" or "either side"?
anchor points is -> "are"
those points is -> "are"

docs/technical_documents/open_space_decider.md
```diff
@@ -0,0 +1,175 @@
...
+    ```cpp
+          void SetParkAndGoEndPose(Frame *const frame);
+    ```
+
+    ```cpp
+          bool GetParkAndGoBoundary(Frame *const frame, const hdmap::Path &nearby_path,
+                                std::vector<std::vector<common::math::Vec2d>>
+                                *const roi_parking_boundary);
+    ```
+2. By calling formulateboundaryconstraints function to gather vertice needed by warm start and distance approach, then transform vertices into the form of Ax > b;
```
JackFu123, line 107: Code style `formulateboundaryconstraints()` could be used if it's a function.

docs/technical_documents/open_space_decider.md
```diff
@@ -0,0 +1,175 @@
...
+                                ReferenceLineInfo* const reference_line_info);
+    ```
+2. Return process status
+
+3. Output: pre stop fence for open space planner
+
+## Open space fallback decider 
+1. Input: obstacles info \ vehicle info \ road info \ parking space info
+
+2. Based on the prdicted trajectory of obstacles, the bounding box info of obstacles in each time interval is obtained,then add into predicted_bounding_rectangles
```
JackFu123, line 156: add -> added.

docs/technical_documents/open_space_decider.md
```diff
@@ -0,0 +1,175 @@
...
+```cpp
+    void BuildPredictedEnvironment(const std::vector<const Obstacle*>& obstacles,
+                                   std::vector<std::vector<common::math::Box2d>>&
+                                   predicted_bounding_rectangles);
+```
+3. Calling IsCollisonFreeTrajectory function to determine whether it will intersect with obstacles;
```
JackFu123, line 162: Personally, it would be nicer if IsCollisonFreeTrajectory is formatted as `IsCollisonFreeTrajectory()`.

JackFu123: Since the original form is frequently used in this doc, why use Present Participle "Call-ing" here?

JackFu123: Since the original form is frequently used in this doc, why use Present Participle "Call-ing" here?

docs/technical_documents/open_space_decider.md
```diff
@@ -0,0 +1,175 @@
...
+```cpp
+    bool IsCollisionFreeTrajectory(
+        const TrajGearPair& trajectory_pb,
+        const std::vector<std::vector<common::math::Box2d>>&
+        predicted_bounding_rectangles,
+        size_t* current_idx, size_t* first_collision_idx);
+```
+4. If ADC trajectroy is collision free, the chosen partitioned trajectory can be used directly, else a fallback trajectroy base on current partition trajectroy will be gererated;
```
JackFu123, line 170: else? Maybe "otherwise"? Try not to use so many semicolons when it's a complete sentence.

docs/technical_documents/open_space_trajectory_partition.md
```diff
@@ -0,0 +1,104 @@
+# OPEN SPACE TRAJECTORY PARTITION
+
+# Introduction
+
+Apollo planning is a scenario based planning method, each driving use case is treated as a different driving scenario.
```
JackFu123, line 5: Same here. It's inappropriate to call planning a method, I would say. :)

docs/technical_documents/open_space_trajectory_partition.md
```diff
@@ -0,0 +1,104 @@
...
+
+# Where is the code
+
+Please refer [open space trajectory parition](https://github.com/ApolloAuto/apollo/modules/planning/tasks/optimizers/open_space_trajectory_partition/open_space_trajectory_partition.cc)
+
+# Code Reading
+
+1. Input : stitched trajectory(without optimization) / vehicle position info
+
+2. The interpolated trajectory is obtained by calling InterpolateTrajectory function to increase stitched trajectory points
```
JackFu123, line 17: `InterpolateTrajectory()`

docs/technical_documents/open_space_trajectory_partition.md
```diff
@@ -0,0 +1,104 @@
...
+        const DiscretizedTrajectory& stitched_trajectory_result,
+        DiscretizedTrajectory* interpolated_trajectory);
+    ```
+3. According to the heading angle and traking angle, the gear shift point can be determined;
+   Use std::vector<TrajGearPair> to store gear infomation and trajectory points, then partition trajectory into a group of trajectories from the gear shift point;
+    ```cpp
+    void PartitionTrajectory(const DiscretizedTrajectory& trajectory,
+                             std::vector<TrajGearPair>* partitioned_trajectories);       
+    ```
+4. If replan due to fallback stop, open space staus -> position_init = false;
```
JackFu123, line 29: Maybe rewrite this sentence?

docs/technical_documents/open_space_trajectory_partition.md
```diff
@@ -0,0 +1,104 @@
...
+    void AdjustRelativeTimeAndS(
+        const std::vector<TrajGearPair>& partitioned_trajectories,
+        const size_t current_trajectory_index,
+        const size_t closest_trajectory_point_index,
+        DiscretizedTrajectory* stitched_trajectory_result,
+        TrajGearPair* current_partitioned_trajectory);
+    ```
+5. When fallback is not required, choose the one to follow based on the closest partitioned trajectory;
+   
+   1. Based on ADC position, heading, body size and velocity info to get vehicle center point info and  moving direction;
```
JackFu123, line 41: two blanks before 'moving'

docs/technical_documents/open_space_trajectory_partition.md
```diff
@@ -0,0 +1,104 @@
...
+    ```cpp
+        void UpdateVehicleInfo();
+    ```
+   2. Encode partitioned trajectories;
+    ```cpp
+        bool EncodeTrajectory(const DiscretizedTrajectory& trajectory,
+                            std::string* const encoding);     
+    ```
+   3. To find the closest point on trajectory, a search range needed to be determined, it required the distance between path end point and ADC enter point, the heading search difference and the head track difference all within the threshold;
```
JackFu123, line 50: Why use past tense 'required' here?

docs/technical_documents/open_space_trajectory_partition.md
```diff
@@ -0,0 +1,104 @@
...
+   Based on ADC box and path point box, the IOU(intersection over union) is computed for each path point in the search range; 
+   If the IOU of the trajectory end point is bigger than threshold and partitioned trajectories group has other trajectory can be used, it means ADC reach the end of a trajectory, need to switch to next one;
```
JackFu123, line 52: ... another trajectory to be used.

docs/technical_documents/open_space_trajectory_partition.md
```diff
@@ -0,0 +1,104 @@
...
+    ```
+    ```cpp
+        bool CheckTrajTraversed(const std::string& trajectory_encoding_to_check);
+    ```
+    ```cpp
+        void UpdateTrajHistory(const std::string& chosen_trajectory_encoding);
+    ```
+   4. When there is no need for ADC to switch to next trajectory, use IOU info mentioned above to find the closest trajectory point(the biggest IOU point) to follow;
+
+   5. If the closest trajectory point doesn't belong to current trajectory or couldn't find closest trajectory point due to some unnormal cases, calling UseFailSafeSearch function to get a safe trajectory to follow;
```
JackFu123, line 70: `UseFailSafeSearch()`

JackFu123 (changes requested):

Issue #13353, opened 2021-01-21
Running Multiple Apollo Instances for Multiple users


xiafawu:
### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 18.04)**: Linux Ubuntu 18.04
- **Apollo installed from (source or binary)**: Source
- **Apollo version (3.5, 5.0, 5.5, 6.0)**: 6.0

### Question
Is it possible to run multiple Apollo instances on a single machine? When trying to run Apollo modules while other users running Apollo already on a different container, I get messages like module <module name> is already running, skipping, when I try to start modules including planning, routing, and dreamview. I could only start my modules when other apollo_dev containers are stopped. Thank you!

storypku:
Hi @EricWu416, I think the main reason for this was that Apollo Docker containers was started  using  the `--pid=host --net=host` options for `docker run`, see: https://github.com/ApolloAuto/apollo/blob/master/docker/scripts/dev_start.sh#L540 . 

You can try to tweak this to see if it works.

storypku:
And, please refer to #13255 for more.

EricWu416:
@storypku Thanks for the suggestion. I tried to remove the `--pid=host` option for `docker run`, which appeared to be working as I won't observe outputs similar to "module is already running". However, when I try to run `bash scripts/planning.sh start_fe`, I noticed it complains about duplicated node of planning, and the module will stop execution afterwards automatically while other users' `apollo_dev` containers are running. Of course, when theses containers are stopped, I could successfully start my modules. 

In addition, you mentioned hardware access in the issue that you wanted me to refer to. I am not entirely sure of what can be counted as hardware access. Could you further explain what you mean by that? Thanks!

gsa88:
Hi @EricWu416! 
In addition to removing `--pid=host`, you should also remove `--net=host`, because without this, the module's channels will be shared across all instances. But when you remove `--net=host` your network inside the docker will be isolated and you will need to expose all necessary ports (eg, 8888 for dreamview `-p 8888:8888` ).

A hardware issue can be caused by the fact that when you only have one CAN device connected to your PC and multiple instances running with one device can cause a collision.

EricWu416:
@gsa88 Sorry for the late reply and thanks for the suggestion. I ended up running several VMs as an alternative, but I will give this solution a try! Thanks again!

Issue #13255, opened 2020-12-28
How to run two Apollos on one machine？


ForresterLu:
Hello team, I am working on Apollo recently, and i need to run 2 apollos parallely, but when i run the second apollo then the first will stop, is there any method to run two apollos parallely?

daohu527:
It does not seem to be supported at present, which will cause some problems, such as hardware sharing.

ForresterLu:
> It does not seem to be supported at present, which will cause some problems, such as hardware sharing.

Can i modify the host of Apollo docker and then run them parallely?

storypku:
> It does not seem to be supported at present, which will cause some problems, such as hardware sharing.

As @daohu527 said, it **may** cause problems if you need to access hardware of the host machine. If hardware access if not needed at all, you can twist `dev_start.sh` and `dev_into.sh` to run two copies of Apollo Dev containers by 1) specifying each a unique `APOLLO_DEV` value. 2) You **may** also need to replace all [`docker_restart_volume`](https://github.com/ApolloAuto/apollo/blob/master/docker/scripts/dev_start.sh#L387) with  [`reuse_or_start_volume`]( https://github.com/ApolloAuto/apollo/blob/master/.teamcity/run_ci.sh#L167). 

To better separate the two containers,  and allow Bazel to build simultaneously, you'd better have two copies of Apollo source code on your host. 

A CyberRT (`cyber_start.sh` & `cyber_into.sh`)  container and a Dev container can run in parallel on the same HOST directory.

Please note that all the stuffs I mentioned are  **experimental** and **undocumented**, use it at your own risk.

daohu527:
Closed due to inactivity. If the problem persists, pls feel free to reopen it or create a new one and refer to it.

PR #13661, opened 2021-04-04
Dev start multiuser mode

gsa88:
There are some issues (#13353 #13255) where users want to be able to run multiple development containers for multiple users on the same host. The script (`dev_start_multiuser.sh`) was prepared for this purpose. 

Using this script, several users work with the apollo project on the same server in my environment and no problems arise.

### DIFF
--- a/docker/scripts/README.md
+++ b/docker/scripts/README.md
@@ -1,0 +1,58 @@
+# Working with Docker containers
+
+## Start and enter Apollo development container
+
+The following commands are assumed to run from Apollo root directory.
+
+```bash
+bash docker/scripts/dev_start.sh
+bash docker/scripts/dev_into.sh
+```
+
+## Start and enter multiple Apollo development containers for multiple users
+
+Each Apollo development container must be run by each user separately. Switch to the desired user and run:
+
+```bash
+bash docker/scripts/dev_start_multiuser.sh
+bash docker/scripts/dev_into.sh
+```
+
+Note:
+
+> Multiple Apollo development containers are isolated now and can interact with host environment through exposed ports for Dreamview and Bridge.
+
+### Start Dreamview in multiple Apollo development containers
+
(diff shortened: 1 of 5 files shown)


MaisJamal:
I want to run multiple instances of Apollo on the same user. Can you please suggest a modification to your code so I can do that?

gsa88:
> I want to run multiple instances of Apollo on the same user. Can you please suggest a modification to your code so I can do that?

Hi MaisJamal. For this, run a script [dev_start_multi_same_user.sh](https://github.com/gsa88/apollo/blob/multiple-instances-on-the-same-user/docker/scripts/dev_start_multi_same_user.sh) as many times as you want to create instances..

After multiple runs of `dev_start_multi_same_user.sh` for the same user, the new apollo dev containers will have different names (e.g. apollo_dev_user_23, `apollo_dev_user_99`, etc.). In this case, the `dev_into.sh` script will not work. To get to a specific container (for example, `apollo_dev_user_23`), you need to run the following command:

```
$ xhost +local:root 1>/dev/null 2>&1

$ docker exec \
    -u user \
    -e HISTFILE=/apollo/.dev_bash_hist \
    -it apollo_dev_user_23 \
    /bin/bash
```
**note**: set the host username with the `-u` option of the `docker exec` command.

MaisJamal:
I run the script dev_start_multi_same_user.sh many times, But all the containers have the same name: apollo_dev_user, so when the second container start the first one closes.

gsa88:
> I run the script dev_start_multi_same_user.sh many times, But all the containers have the same name: apollo_dev_user, so when the second container start the first one closes.

Could you provide output of `dev_start_multi_same_user.sh`?

MaisJamal:
My bad. I was using dev_start_multiuser.sh instead of dev_start_multi_same_user.sh. Now everything works fine.
Thanks a lot for sharing your code and for answering my questions.

PR #13685, opened 2021-04-12
docs: add planning tasks docs: path_bounds_decider,piecewise_jerk…

liouyx:
…_path_optimizer,path_assessment_decider,path_decider,rule_based_stop_decider


docs/technical_documents/tasks/path_assessment_decider_en.md
```diff
@@ -0,0 +1,297 @@
...
+  return Status::OK();
+```
+
+Update necessary information:
+
+1.Update static obstacles info in front of adc.
+2.Update self lane usage information.
+3.Update the direction of the side lane
+(1) According to `PathDeciderStatus` is RIGHT_BORROW or LEFT_BORROW,decide whether to borrow from the left or from the right.
+
```
storypku, line 251: We don't name English documents with an `_en`  suffix, only documents in other languages (like Chinese) should be suffixed with `_cn`.

PR #13727, opened 2021-04-25
dreamview: fix bug at sim control

qwetqwe:
InitStartPoint() can handle  using localization point and using dummy point.if localization is not empty, program will jump here.

### DIFF
--- a/modules/dreamview/backend/sim_control/sim_control.cc
+++ b/modules/dreamview/backend/sim_control/sim_control.cc
@@ -107,7 +107,7 @@ void SimControl::InitTimerAndIO() {
       false));
 }
 
-void SimControl::Init(bool set_start_point, double start_velocity,
+void SimControl::Init(double start_velocity,
                       double start_acceleration) {
   if (!FLAGS_use_navigation_mode) {
     InitStartPoint(start_velocity, start_acceleration);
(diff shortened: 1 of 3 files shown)


storypku:
@lj0919 Please help review this PR. Thank you.

lj0919:
@qwetqwe https://github.com/ApolloAuto/apollo/blob/f02eda58ca1abf939c567fc8db3a461fcbcbc95a/modules/dreamview/backend/sim_control/sim_control.cc#L255 when localization reader has no data, set_start_point is passed in true, we will init start point

qwetqwe:
Hi, @lj0919 
In SimControl::InitStartPoint ,the function seems that if localization not empty, sim control will use lastest localization message.
https://github.com/ApolloAuto/apollo/blob/f02eda58ca1abf939c567fc8db3a461fcbcbc95a/modules/dreamview/backend/sim_control/sim_control.cc#L140-L164
The function explaination :
https://github.com/ApolloAuto/apollo/blob/f02eda58ca1abf939c567fc8db3a461fcbcbc95a/modules/dreamview/backend/sim_control/sim_control.cc#L120-L121
But program can not get into this branch, and set start point  when localization not empty, and will output localization 0 0 0 as default.

lj0919:
@qwetqwe  Hi, Sorry for the late reply, I misunderstood what you meant at first, the else branch looks really redundant.

modules/dreamview/backend/sim_control/sim_control.cc
```diff
@@ -109,7 +109,7 @@ void SimControl::InitTimerAndIO() {
 
 void SimControl::Init(bool set_start_point, double start_velocity,
```
lj0919, line 110: The first argument set_start_point is no longer needed, remove it to make the function cleaner,  then test to make sure there is no problem with sim_control

lj0919 (changes requested):
Remove unused argument to make the function cleaner

lj0919 (approved):

PR #13853, opened 2021-06-15
Routing: Fix bug of vertical parking failure.

chenxiaoying21:
Fix $[IDG-Apollo-4346].
When the vehicle is in front of the parking spot, and on the lane next
to the nearest lane to the parking spot, the parking action will fail
because of the following error:
Failed to extract segments from routing.
and the relative error of pnc map like:
Failed to find nearest point: x: 437868.09950172278 y: 4433040.1114324955 z: 0

Cause:
The lane segments of the routing response don't contain the lane nearest
to the parking spot because the vehicle is in front of the parking spot.
When the vehicle approach to the parking spot, no project point can be
found on the routing segments and thus CreateReferenceLine failed.

Solution:
1. Add SupplementParkingRequest function. When the routing segements
doesn't contain the lane nearest to the parking spot, add it into the
sgement list.
2. Modify the way of searching the parking spot according to the vehicle
state as well as the way of searching the nearest lane to the
vehicle(GetNearestPointFromRouting).

### DIFF
--- a/modules/map/hdmap/hdmap_common.cc
+++ b/modules/map/hdmap/hdmap_common.cc
@@ -192,14 +192,17 @@ void LaneInfo::GetWidth(const double s, double *left_width,
 }
 
 double LaneInfo::Heading(const double s) const {
-  const double kEpsilon = 0.001;
-  if (s + kEpsilon < accumulated_s_.front()) {
-    AERROR << "s:" << s << " should be >= " << accumulated_s_.front();
+  if (accumulated_s_.empty()) {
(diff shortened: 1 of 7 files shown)


JasonZhou404 (approved):

twbabyduck:
@JasonZhou404 
Hi Jason,
Do you have instant message or email that I can contact you?
twbabyduck@gmail.com
I would like to send an email to ask some details regarding to parking algorithm.
hopefully can make contribution :)
Thanks

PR #13871, opened 2021-06-21
Control:add set station PID and reset integral

panxueliang0:
Control: Reset integral when GEAR_NEUTRAL or full_stop.

Fix [IDG-Apollo-4329].
Move too fast after shifting or when entering self-driving.

Cause:
After shifting or entering autonomous driving, no integral zeroing
is carried out, resulting in excessive initial control when re-entering control

Solution:
Reset integral when GEAR_NEUTRAL or full_stop.

Control: Reset the Station PID after switching from reverse to forward.

Fix [IDG-Apollo-4330].
The Station PID parameter is reset in reverse gear, but the Station PID
is not reset after switching from reverse to forward, only the Speed PID
parameter is reset

Solution:
Reset the Station PID after switching from reverse to forward

### DIFF
--- a/modules/control/common/pid_controller.cc
+++ b/modules/control/common/pid_controller.cc
@@ -58,6 +58,11 @@ double PIDController::Control(const double error, const double dt) {
   return output;
 }
 
+void PIDController::Reset_integral() {
+  integral_ = 0.0;
+  integrator_saturation_status_ = 0;
+}
+
 void PIDController::Reset() {
   previous_error_ = 0.0;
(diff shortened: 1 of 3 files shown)


chenxiaoying21 (changes requested):
Commit的第一行不要写Fix [IDG-Apollo-4330]，这句话没法看出来改了什么。
要写修改的摘要，比如：
Control: Reset the Station PID after switching from reverse to forward.

chenxiaoying21 (approved):

Issue #13882, opened 2021-06-23
apollo 5.5 & 6.0 module/planning/tasks/deciders: LaneChangeDecider::IsPerceptionBlocked() question


zynlemon:
We appreciate you go through Apollo documentations and search previous issues before creating an new one. If neither of the sources helped you with your issues, please report the issue using the following form. Please note missing info can delay the response time.

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 18.04)**: Linux Ubuntu 20.04
- **Apollo installed from (source or binary)**: official git reporsitory
- **Apollo version (3.5, 5.0, 5.5, 6.0)**: version 5.5 & 6.0
- **Output of `apollo.sh config` if on `master` branch**: don't know what this mean

### Steps to reproduce the issue:
- **Please use bullet points and include as much details as possible**:
hello dear developers, I have a question when reading planning module code:
in LaneChangeDecider::IsPerceptionBlocked():
```
right_most_angle = beam_heading
...
left_most_angle = beam_heading - search_angle
```
then:
```
std::fabs(common::math::NormalizeAngle(left_most_angle - right_most_angle))
```
is this angle the search_angle when beam first blocked by obstacle poligon point?
what is the meaning of this angle?
should left_most_angle = beam_heading and then the angle mentioned before becomes the angle that beam blocked by the obstacle?

### Supporting materials (screenshots, command lines, code/script snippets):
![IsPerceptionBlocked](https://user-images.githubusercontent.com/76237917/123089565-afb53b80-d459-11eb-8e07-923ce0f70722.png)

daohu527:
The beam_heading is rotate from right to left, and when it found obstacle then will search for the obstacle left boundary. If the (left-right) > 0.5(radian) then means it is blocked. 

## sorted
`reference_line_info.path_decision().obstacles()` should be sort from right to left, Otherwise `right_most_angle` will be wrong

#### right_most_angle 
if obstacles are sorted, then `right_most_angle` will found only once be **the right boundary of the rightmost obstacle**.

#### left_most_angle
after found `right_most_angle`, `left_most_angle` will continue found be `0`

## In summary
So, this is an issue to be fixed, thank you for your submission.

The real intention is shown in the figure below
![Agile Board](https://user-images.githubusercontent.com/10419854/124408122-20bcf300-dd78-11eb-8c52-16e653595fd9.jpg)

tuanngokien:
Hi @daohu527,
[...]

Issue #13228, opened 2020-12-22
Unexpected decision of side pass stop 


ZiwenWan:
Hi Apollo team, 

I met an unexpected result when I ran the planning module. I manually added a static obstacle in single-lane road map under a lane borrowing scenario.  The function IsPerceptionedBlocked in lane change decider considers the small static obstacle is blocking the perception since the blocking angle is larger than the threshold (0.5). 

### System information

- **OS Platform and Distribution (e.g., Linux Ubuntu 18.04)**: Ubuntu 14.04
- **Apollo installed from (source or binary)**: from LGSVL repo: https://github.com/lgsvl/apollo-5.0.git
- **Apollo version (3.5, 5.0, 5.5, 6.0)**: 5.0
- **Output of `apollo.sh config` if on `master` branch**:  default branch (simulator)

### Steps to reproduce the issue:
The steps of reproducing:
1. Put a static vehicle in front of the ego car in the same lane (as shown in picture)
2. Put a static obstacle out of the lane (as shown in the picture)
Without adding the obstacle in step 2, the ego car will generate a lane borrow trajectory. However, if I put the static obstacle out of the lane, ego car will make a side pass stop decision because the lane change decider thinks the perception is blocked by the small static obstacle. 

### Supporting materials (screenshots, command lines, code/script snippets):
In the following code(planning/tasks/deciders/lane_change_decider/lane_change_decider.cc), the difference between left_most_angle and right_most_angle is larger than 2. In my understanding, the difference should be the blocking angle caused by the obstacle, which is much larger than the actual value here. 
`if (std::fabs(common::math::NormalizeAngle(
            left_most_angle - right_most_angle)) > is_block_angle_threshold) {
      return true;
    }`
<img width="238" alt="Issue_pic" src="https://user-images.githubusercontent.com/38177263/102850407-467f3d80-43ce-11eb-8e74-07b0c7c57177.png">

ZiwenWan:
@liouyx @JackFu123 @JasonZhou404 Hi, just want to check out whether there is any progress on this issue.

PR #13928, opened 2021-07-07
Fix #13921, #13882, #13228, missing initialization in IsPerceptionBlocked

### LINKED ISSUE #13921: Lack of initialization of variables.
Dear Apollo team, thanks for creating this wonderful tool!   
**Describe the bug**
We recently find a bug via our automated tool. This bug is located in lane_change_decider.cc IsPerceptionBlocked (https://github.com/ApolloAuto/apollo/blob/266afbf68d83fa6fac7a812ff8a950223f5ab2c0/modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc#L368). We found that the necessary initialization are missing in IsPerceptionBlocked function. To be more specific, right_most_found, right_most_angle, and left_most_angle are variables maintained for each obstacle and they are not initialized when processing a new obstacle. The following figure is an example we discovered that can trigger the vulnerability. All the lanes are clear except the front blocking obstacle we put to create a lane borrow scenario. The ego vehicle makes a stop decision for this side-pass. After more investigation, we found that this could be caused by uninitialized variables. When the beam scanning first scans the left side obstacle, it will get updated left_most_angle and right_most_angle. Also, the right_most_found is set to be true. When processing the 2nd obstacle, as long as the obstacle does not have intersection with the first beam, the statement “left_most_angle = beam_heading - search_angle;” will be executed. As a result, left_most_angle will be set with “adv_heading - 0.5 search_range” due to the right_most_found being uninitialized. As long as the difference between right_most_angle of the first obstacle and the current left_most_angle is larger than the predefined threshold (0.5), the ego vehicle will make a stop decision on side-pass.

**To Reproduce**
Here is a setup of a driving scenario that trigger this bug. 
We first set up a lane borrow scenario with a vehicle blocking the lane. Then we put two small boxes off the road. The requirements for the boxes are very easy to be satisfied. We can put the first box on the left side of the road to make sure the right_most_angle of this box is large enough. The second box just needs to be visible to the ego vehicle. As a result, the ego vehicle will stop there permanently. 

**Expected behavior**
The Apollo should finish lane borrow successfully since no obstacle is on the road.
**Screenshots**

image

randomacco:
Hi I have fixed the initialization problem in function IsPerceptionBlocked in lane change decider raised by 3 issues. Please review and consider fixing it. If you have any questions, please let me know. 
Thanks a lot!

### DIFF
--- a/modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
+++ b/modules/planning/tasks/deciders/lane_change_decider/lane_change_decider.cc
@@ -366,18 +366,12 @@ bool LaneChangeDecider::IsClearToChangeLane(
 }
 
 bool LaneChangeDecider::IsPerceptionBlocked(
-    const ReferenceLineInfo& reference_line_info,
-    const double search_beam_length, const double search_beam_radius_intensity,
-    const double search_range, const double is_block_angle_threshold) {
+  const ReferenceLineInfo& reference_line_info,
+  const double search_beam_length, const double search_beam_radius_intensity,
+  const double search_range, const double is_block_angle_threshold) {
   const auto& vehicle_state = reference_line_info.vehicle_state();
   const common::math::Vec2d adv_pos(vehicle_state.x(), vehicle_state.y());
   const double adv_heading = vehicle_state.heading();
-
-  double left_most_angle =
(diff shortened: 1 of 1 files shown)


randomacco:
@daohu527 Actually, only the front obstacles will be considered in the original code. Note that the search range is PI.

PR #13955, opened 2021-07-20
Bazel: unified --repository_cache for bazel sync

storypku:
If we specify `--repository_cache` for the `build` command rather than
a `common` option, commands like `bazel sync` will not respect the
settings there.

Ref:
https://docs.bazel.build/versions/main/guide.html#option-defaults


storypku:
@ChangSonghong @diegohu

changsh726 (approved):

PR #13988, opened 2021-07-29
Planning: add readme and pictures

xiaoluauto:


JasonZhou404 (changes requested):
Don't delete the old readme, but write after that

PR #14021, opened 2021-08-12
optimize the filter rule of channel for cyber_monitor

Allenhe123:
If using cyber_monitor tool to monitor channels, the Arm V8 will be slow when there were a lot of channels in the LAN.
The old filter can only filter one channel.People often need to filter channels by key words.


qwetqwe (commented):
There is a spelling mistake about channelname.
It can only filter channels with same substring. Is it general?

PR #14157, opened 2021-11-01
add new vehicle hunter and add new controller speed controller for hunter

panxueliang0:
Add new vehicle hunter2.
Use the throttle to send speed instructions.

A new controller is registered，select a controller based on the control_conf.pb.txt


qwetqwe (commented):
1. why changed localization module?
2.Is it necessary to enable chassis detail

panxueliang0:
> 1. why changed localization module?

> 2.Is it necessary to enable chassis detail

PR #14177, opened 2021-11-15
Planning: fix bug in lane borrow

qwetqwe:
at the middle of lane borrowing, adc will get into self_lane


qwetqwe:
IDG-Apollo-4513

lx18233184051 (approved):

PR #14219, opened 2021-12-06
Add the case in fem_pos_deviation_osqp_interface when the number of optimized path points is 3 

nkuwenjian:
The minimum number of optimized path points is 3 in fem_pos_deviation_osqp_interface. Currently, the QP modeling in the CalculateKernel function only works when the number of path points is greater than or equal to 4. If the number of path points is 3, the code will crash. So I add a separate case for the QP modeling when the number of optimized path points is 3.


daohu527:
@qwetqwe pls help to review this!

PR #14426, opened 2022-04-24
Prediction: cruise mlp extrapolation code bug

Xinchengzelin:
Generate 4 features for each lane point 
`
relative_s_l.second, relative_s_l.first, relative_ang, kappa
`


Xinchengzelin:
Sorry, Only the `modules/prediction/evaluator/vehicle/cruise_mlp_evaluator.cc` need to be changed

PR #14442, opened 2022-05-05
correct start timestamp of planning component

lihaoxiang1989:
It makes no sense to have dt=0. And cause the optimizing algorithm is really time-consuming, a time shift is necessary.
I have tested it in several cases and got a better control output speed curve.

### DIFF
--- a/modules/planning/planning_component.cc
+++ b/modules/planning/planning_component.cc
@@ -16,6 +16,7 @@
 #include "modules/planning/planning_component.h"
 
 #include "cyber/common/file.h"
+#include "cyber/time/clock.h"
 #include "modules/common/adapters/adapter_gflags.h"
 #include "modules/common/configs/config_gflags.h"
 #include "modules/common/util/message_util.h"
@@ -31,6 +32,7 @@ namespace apollo {
 namespace planning {
 
 using apollo::cyber::ComponentBase;
+using apollo::cyber::Clock;
 using apollo::hdmap::HDMapUtil;
 using apollo::perception::TrafficLightDetection;
 using apollo::relative_map::MapMsg;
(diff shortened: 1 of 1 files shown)


lihaoxiang1989:
@qwetqwe @changsh726, please review this PR, thx

qwetqwe:
Thanks for you PR. This may be a history bug. on_lane_planning has recorded start_timestamp:

> https://github.com/ApolloAuto/apollo/blob/e076feecb46232bacdf08490c5ff213832e6bdd8/modules/planning/on_lane_planning.cc#L131

 It will be better to change like this:
```cpp
  planning_base_->RunOnce(local_view_, &adc_trajectory_pb);
  auto start_time = adc_trajectory_pb.header().timestamp_sec();
  common::util::FillHeader(node_->Name(), &adc_trajectory_pb);

  // modify trajectory relative time due to the timestamp change in header
  const double dt = start_time - adc_trajectory_pb.header().timestamp_sec();
```

lihaoxiang1989:
@qwetqwe I agree with this. Can I modify it from here? or make another PR? :)

lihaoxiang1989:
> Thanks for you PR. This may be a history bug. on_lane_planning has recorded start_timestamp:
> 
> > https://github.com/ApolloAuto/apollo/blob/e076feecb46232bacdf08490c5ff213832e6bdd8/modules/planning/on_lane_planning.cc#L131
> 
> It will be better to change like this:
> 
> ```c++
>   planning_base_->RunOnce(local_view_, &adc_trajectory_pb);
>   auto start_time = adc_trajectory_pb.header().timestamp_sec();
>   common::util::FillHeader(node_->Name(), &adc_trajectory_pb);
> 
>   // modify trajectory relative time due to the timestamp change in header
>   const double dt = start_time - adc_trajectory_pb.header().timestamp_sec();
> ```

Based on your advice, I made another PR for this issue. Pls check

PR #14519, opened 2022-07-14
Update hdmap_util.cc

Bzhnja:


daohu527 (commented):
pls format the commit like below:
map: fix path error FLAGS_map_dir to dir in FindFirstExist

PR #14542, opened 2022-07-28
Cyber: fixed a bug for TimerComponent.

davidhopper2003:
The old message callback handler for TimerComponent didn't take the termination condition into account. It is fixed now.

### DIFF
--- a/cyber/component/timer_component.cc
+++ b/cyber/component/timer_component.cc
@@ -45,7 +45,7 @@ bool TimerComponent::Initialize(const TimerComponentConfig& config) {
 
   std::shared_ptr<TimerComponent> self =
       std::dynamic_pointer_cast<TimerComponent>(shared_from_this());
-  auto func = [self]() { self->Proc(); };
+  auto func = [self]() { self->Process(); };
   timer_.reset(new Timer(config.interval(), func, false));
   timer_->Start();
   return true;


davidhopper2003:
There is  a typo in PiecewiseJerkProblem::CalculateAffineConstraint(), which has been corrected. Please refer to the original author Yajia Zhang's paper "Optimal Vehicle Path Planning Using Quadratic Optimization for Baidu Apollo Open Platform".

daohu527 (commented):
The timer will destroy in `Shutdown`, and `Proc` is a virtual function overridden by the derived class. seems to be ok

davidhopper2003:
> The timer will destroy in `Shutdown`, and `Proc` is a virtual function overridden by the derived class. seems to be ok

First, if there is a very time-consuming task in the `Proc` function and if it is called directly, the `Proc` function needs to be waited for a long time before exit when the shutdown signal (i.e. `is_shutdown_ = true`) is received. If the `Process` function is called , the routine can be finished immediately:Secondly, from the design point of view, Proc should be called by Process, see the similar lambda expression in ``:
````cpp
  auto func = [self, blocker1](const std::shared_ptr<M0>& msg0) {
       auto ptr = self.lock();
       if (ptr) {
         if (!blocker1->IsPublishedEmpty()) {
           auto msg1 = blocker1->GetLatestPublishedPtr();
           ptr->Process(msg0, msg1);
         }
       } else {
         AERROR << "Component object has been destroyed.";
       }
     };
````

daohu527 (approved):
looks like you are right : )

dreamholy (approved):

qilinhu (approved):

PR #14584, opened 2022-09-09
Update node.h

Charlie-Li-Code:
fix a typo


daohu527:
@Charlie-Li-Code Thanks for your interest in Apollo, We remind you that we should ensure that the format of the commit is consistent.

You can follow the docs in https://github.com/ApolloAuto/apollo/blob/master/CONTRIBUTING.md#commit-message

PR #14601, opened 2022-09-20
Bug fixed

xxmiprai:
`use_object_builder_`使用逻辑或来计算，得到的结果永远是true，而根据算法特性，PointPillars、MaskPillars和CenterPoint是无需使用object_builder的，因此应该使用逻辑与。

### DIFF
--- a/modules/perception/lidar/app/lidar_obstacle_detection.cc
+++ b/modules/perception/lidar/app/lidar_obstacle_detection.cc
@@ -46,8 +46,8 @@ bool LidarObstacleDetection::Init(
   ACHECK(cyber::common::GetProtoFromFile(config_file, &config));
   use_map_manager_ = config.use_map_manager();
   use_object_filter_bank_ = config.use_object_filter_bank();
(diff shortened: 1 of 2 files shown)


daohu527:
This condition is indeed problematic, but `PointPillarsDetection, MaskPillarsDetection` also use `use_object_builder_`

xxmiprai:
> @xxmiprai Thank you for your feedback.The condition is indeed problematic, but `PointPillarsDetection, MaskPillarsDetection` also use `use_object_builder_`

I don't think `PointPillarsDetection`， `MaskPillarsDetection` and `CenterPointDetection` need to use `use_object_builder_`. And I have verified it， please check it.

daohu527:
@xxmiprai In principle it is really not needed, but this may be a historical issue.

In object_builder we set:
- object->polygon
- object->anchor_point
- object->latest_tracked_time
- object->theta   // set in PointPillarsDetection
- object->size
- object->center

So we still need `object_builder` to help set the properties above. 

It's not reasonable after all, and we will optimize it later!!!

PR #14693, opened 2022-11-25
ROCm support added.

GowthamKudupudi:
It's tested on MI100 and MI200 cards but should support any  card that supports ROCm.


GowthamKudupudi:
reopened the https://github.com/ApolloAuto/apollo/pull/14681 which has all the commits.

PR #14699, opened 2022-11-28
Planning: fix bug in Hybrid a star

wildbluepeacelove:
for case when node in open set already visited with lower trajectory cost


xianglunkai:
I don't think there is a big mistake here, because we just need a rough solution to open up a convex space, and then get the real trajectory through optimization。

PR #14732, opened 2022-12-31
add feature：dump record to excel

InterKnight:
add a feature：dump record to excel

### DIFF
--- a/modules/tools/dump_record_to_excel/BUILD
+++ b/modules/tools/dump_record_to_excel/BUILD
@@ -1,0 +1,13 @@
+load("@rules_python//python:defs.bzl", "py_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+py_binary(
+    name = "dump_record",
+    srcs = ["dump_record.py"],
+    deps = [
+        "//cyber/python/cyber_py3:cyber",
+        "//cyber/python/cyber_py3:record",
+        "//modules/tools/common:message_manager",
+    ],
+)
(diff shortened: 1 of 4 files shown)


daohu527:
@InterKnight Pls make a english version of readme, and we'd better change the name to readme.md.

In addition, please explain the application scenario of the function, can we not directly output to csv format

InterKnight:
> 
_Pls make a english version of readme, and we'd better change the name to readme.md._

no problem


_please explain the application scenario of the function, can we not directly output to csv format_

As far as I know，if we want to parse some topic to file，we need write code for each topic because any topic’s proto is different，so I write this script to parse any topic. 
I haven't support csv format now, but it can be added to TODO list

And I have two quesetions:
1, Do I need to make a new pull_requset for the changes above?
2, I see the **apollo_iscan** is faild, what can I do to make it pass?

daohu527:
@InterKnight Thank you for your contribution, we need to evaluate whether this tool need to merge into the main branch, in fact we may want to create a separate repository for tools.

Actually Apollo now supports packages, like python packages. The merging of projects will also be graded, and it will be decided whether to merge into apollo according to the usage of the tools. During this period, you can promote your tools.

A bit like an incubation mechanism for new projects.
1. Create a project, and it is really useful, and has a certain number of users
2. Submit the package and merge it into the official repository or recommended list

I think we are at the first stage, I think you can keep your repo and promote it in the community, we are always here to help!

PR #14933, opened 2023-05-11
fix(perception): remove redundant semicolons

MambaWong:


daohu527:
Thank you for your reminder, this bug has been fixed in the next version. We will release it later

Issue #15161, opened 2023-10-17
timer_test Fails if changing cyber configuration


CHOcho-quan:
**Describe the bug**
I change default process number of the default cyber conf file from 16 to 4 and rerun tests. Timer test will have a segmentation fault.

**To Reproduce**
1. Change default_proc_num from 16 to 4 in cyber/conf/cyber.pb.conf.
2. Run tests by: ./apollo.sh test cyber

Timer test will fail and use gdb to check we'll find segmentation fault.

**Screenshots**
<img width="1160" alt="截屏2023-10-17 上午9 46 59" src="https://github.com/ApolloAuto/apollo/assets/37149574/17bbd3a0-f961-45e1-a96d-6bb58c8a2132">

**Environment**
 - OS: Ubuntu 22.04
 - GCC-11.04

PR #15198, opened 2023-11-18
Fix Timing Wheel Concurrency bug

CHOcho-quan:
As #15161 stated, there's a concurrency issue under timing wheel.

Considering 3 threads, running the following content
* Thread I: Starts the timer, and spinning for sometime. Then Stops the timer.
* Thread II: The timing wheel thread, ticking new tasks
* Thread III: A task added to TaskManager and running asynchronously

Now Let's say Thread I stops the timer and reset task_ when Thread II has already run line 52 (timing_wheel.cc) & got an instance of the shared pointer. The only shared pointer task is only maintained in Thread II. Let's say somehow the scheduler haven't got enough resource to run Thread III and Thread II finished first. In this case, task_ is released.

If now, Thread III is scheduled by TaskManager and running line 60 (timing_wheel.cc) , since this callback is caught by task->callback, this is a heap-use-after-release problem. This could also be detected by ASAN.

### DIFF
--- a/cyber/timer/timing_wheel.cc
+++ b/cyber/timer/timing_wheel.cc
@@ -49,18 +49,18 @@ void TimingWheel::Tick() {
     std::lock_guard<std::mutex> lock(bucket.mutex());
     auto ite = bucket.task_list().begin();
     while (ite != bucket.task_list().end()) {
-      auto task = ite->lock();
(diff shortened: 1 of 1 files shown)


daohu527:
Will check and feedback soon!

PR #15407, opened 2024-05-23
modify Enbroad gnss driver  code and improve compatibility

dengwei19840721:
具体详细说明，参考《亿波达组合惯导接入Apollol设计方案》


modules/drivers/gnss/conf/gnss_conf.pb.txt
daohu527, line 1: Roll back changes to this file

PR #15446, opened 2024-06-26
fix : fix README_cn.md ;The “planning” in the document was misspelled as “lanning”.

is-whale:
The “planning” in the document was misspelled as “lanning”.


daohu527 (approved):
LGTM

PR #15706, opened 2025-06-27
fix: timer error accumulation not being reduced after compensation

lazyze:
The `accumulated_error_ns` was being used to compensate timing jitter, 
but its value was never reduced, causing long-term overcompensation.


MambaWong:
never reduced?
You can try to shorten the execution time or increase the interval time. 
Then see how it goes.

PR #15762, opened 2026-02-09
Recover seyond lidar driver

guyuechaocc:
1. recover seyond lidar driver code
2. support seyond hummingbird lidar
3. update readme


yuxin-2:
Seyond lidar

PR #15814, opened 2026-08-30
Cyber: prevent queued timer callback use-after-free

### LINKED ISSUE #15810: TimingWheel::Tick() use-after-free crash when Timer::Stop() called concurrently
**Describe the bug**
  When stopping a timer while TimingWheel::Tick() is executing, the application crashes with SIGSEGV due to use-after-free. The crash typically occurs in the coroutine scheduler thread (.CRT thread).

  Stack trace:
  #0 0x... in std::function<void()>::operator()()
  #1 in TimingWheel::Tick()
  #2 in std::call_once

**questions**
  In timing_wheel.cc, the Tick() function captures a raw pointer to the callback function instead of capturing the shared_ptr<TimerTask>:

  // Bug in timing_wheel.cc:54-75
  void TimingWheel::Tick() {
    auto task = ite->lock();
    if (task) {
      auto* callback = reinterpret_cast<std::function<void()>*>(&(task->callback));
      gaea::Async([this, callback] {  // BUG: captures raw pointer!
        if (this->running_) {
          (*callback)();  // crashes if TimerTask already freed
        }
      });
    }
  }

  Race condition timeline:
  1. Thread A: Tick() gets callback* pointer
  2. Thread B: Timer::Stop() calls task_.reset(), freeing TimerTask
  3. Thread A: Async task executes (*callback)() on freed memory → CRASH

**To Reproduce**
Long time start and stop application
1. Create a timer with 20ms period
2. In a separate thread, repeatedly call Stop() + reset() on the timer
3. When Tick() and Stop() race, crash occurs

**Expected behavior**
Not crash 

**Additional context**
  This bug is similar to the classic use-after-free pattern in concurrent systems where a raw pointer is used after the underlying object has been freed. The fix ensures proper lifetime management
  through shared_ptr reference counting.

timothyanderson096-ocdealcheck:
## Summary

* replace the queued raw callback pointer with a weak `TimerTask` reference
* discard queued work when `Timer::Stop()` has already destroyed the task
* preserve the canonical callback object across periodic firings
* add regressions for stop-before-dequeue lifetime and mutable callback state

## Rationale

Capturing a strong task reference would keep a stopped timer alive, while copying the callback would change stateful periodic-callback semantics. Locking a weak task reference at execution time avoids the reported use-after-free without either behaviour change.

## Testing

* added `TimerConcurrencyTest.stop_before_queued_callback_runs`
* added `TimerTest.periodic_callback_preserves_state`
* standalone ASan reproducer: the original raw-pointer path reports heap-use-after-free; the weak-reference path completes without invoking the stopped callback
* `git diff --check`

The Apollo Bazel target could not be run locally because the Apollo development container/toolchain was unavailable. Repository CI will provide the project-level build and test result.

Fixes #15810

### DIFF
--- a/cyber/timer/timer.h
+++ b/cyber/timer/timer.h
@@ -25,6 +25,10 @@
 namespace apollo {
 namespace cyber {
 
+namespace timer {
+class TimerConcurrencyTest;
+}  // namespace timer
+
 /**
  * @brief The options of timer
  *
@@ -110,6 +114,8 @@ class Timer {
   void Stop();
 
  private:
+  friend class timer::TimerConcurrencyTest;
+
   bool InitTimerTask();
   uint64_t timer_id_;
   TimerOption timer_opt_;
--- a/cyber/timer/timer_test.cc
+++ b/cyber/timer/timer_test.cc
@@ -18,14 +18,21 @@
 
 #include "cyber/timer/timer.h"
 
+#include <atomic>
+#include <chrono>
+#include <cstdint>
+#include <future>
 #include <memory>
+#include <thread>
 #include <utility>
+#include <vector>
 
 #include "gtest/gtest.h"
 
 #include "cyber/common/util.h"
 #include "cyber/cyber.h"
 #include "cyber/init.h"
+#include "cyber/scheduler/scheduler_factory.h"
 
 namespace apollo {
 namespace cyber {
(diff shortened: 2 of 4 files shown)


ZeusJupiter:
这是来自QQ邮箱的自动回复邮件。您好，您的邮件我已经收到。我会尽快给您回复。
谢谢。

